PR #3, opened 2025-03-12
ci: fix the cpu test workflow to main branch

SaboniAmine:
## Description
This PR proposes a trigger configuration to fix the workflow version to the main branch.

## Type of Change

## Checklist

### DIFF
--- a/.github/workflows/external_contributions_tests.yaml
+++ b/.github/workflows/external_contributions_tests.yaml
@@ -1,21 +1,21 @@
-name: Fast CPU Tests
+name: Fast CPU Tests for external contributions
 
 on:
-  push:
-    branches:
-      - main
   pull_request_target:
     types: [ review_requested ]
-    branches:
-      - main
+    branches: [ main ]
 
 jobs:
   test:
+    if: >
+      github.event.pull_request.author_association != 'OWNER' &&
+      github.event.pull_request.author_association != 'MEMBER' &&
+      github.event.pull_request.author_association != 'COLLABORATOR'
     runs-on: pruna-cpu
 
     strategy:
       matrix:
-        python-version: ["3.10"]
+        python-version: [ "3.10" ]
 
     steps:
       - name: Checkout code
@@ -39,9 +39,6 @@ jobs:
 
       - name: Install dependencies
         run: pip install -e .[tests] --extra-index-url https://prunaai.pythonanywhere.com/
-
-      - name: Install intel dependencies
-        run: pip install -e .[intel] --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/cpu/cn/
 
       - name: Run tests
         run: |
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -2,7 +2,8 @@ name: Algorithm Docs Generation
 
 on:
   pull_request:
-    branches: [ main, development ]
+    branches:
+      - main
 
 jobs:
   generate-docs:
--- a/.github/workflows/internal_contributions_tests.yaml
+++ b/.github/workflows/internal_contributions_tests.yaml
@@ -1,0 +1,53 @@
+name: Fast CPU Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    if: >
+      github.event.pull_request.author_association == 'OWNER' ||
+      github.event.pull_request.author_association == 'MEMBER' ||
+      github.event.pull_request.author_association == 'COLLABORATOR'
+    runs-on: pruna-cpu
+
+    strategy:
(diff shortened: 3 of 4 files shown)


.github/workflows/cpu_tests.yaml
```diff
@@ -4,14 +4,16 @@ on:
   push:
     branches:
       - main
-      - development
```
johannaSommer, line 7: thanks a lot - forgot about this. Could we please remove this also for the other workflows? --> linting and documentation

.github/workflows/external_contributions_tests.yaml
```diff
@@ -0,0 +1,53 @@
+name: Fast CPU Tests for external contributions
+
+on:
+  pull_request_target:
```
johannaSommer, line 4: thanks a lot, `on: pull_request_target` was exactly the setting I was referring to! Do you think it would make sense to also add this to linting and documentation (not necessary for `cpu_tests.yaml` I think, this is anyway only for members).

.github/workflows/external_contributions_tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+    types: [review_requested]
```
johannaSommer: Hm not sure I can 100% follow - so now we run the the tests as soon as the review is requested? I remember discussing that this tests need approval and I can't quite parse yet where this setting is in this yaml

.github/workflows/external_contributions_tests.yaml
```diff
@@ -0,0 +1,53 @@
...
+          echo "${HOME}/.local/bin" >> $GITHUB_PATH
+
+      - name: Configure Poetry to not create virtual environments
+        run: poetry config virtualenvs.create false
+
+      - name: Install dependencies
+        run: pip install -e .[tests] --extra-index-url https://prunaai.pythonanywhere.com/
+
+      - name: Install intel dependencies
+        run: pip install -e .[intel] --extra-index-url https://pytorch-extension.intel.com/release-whl/stable/cpu/cn/
```
johannaSommer: I think "intel" is outdated for Pruna and will no longer work

johannaSommer: okay this is also still present in the CPU tests, @johnrachwan123 can you confirm this is outdated?

.github/workflows/cpu_tests.yaml
```diff
@@ -4,14 +4,16 @@ on:
   push:
```
sharpenb, line 4: There is now one workflom for internal cpu tests andone for external cpu tests. Coudl we name the workflow files accordingly i.. internal/external_contribution_cpu_tests

.github/workflows/external_contributions_tests.yaml
sharpenb: Since the other workflows are cheaper, it is fine if we do not change but I agree that it could also make sense for the other workflows.

johannaSommer: @sharpenb just for context - in my understanding currently the `on: pull_request_target` setting is not so much about when someone is allowed to run the models - it specifies that the PR is run i the context of the main branch. This means that even if someone opens a PR modifying the workflow to always run, this will not happen and only once this would be merged into the main branch (which we catch in a review).

.github/workflows/external_contributions_tests.yaml
SaboniAmine: Yes, for first time contributors, there still is a need for an approval before being able to access the workflow.
Second time contributors won't have this gate, so they will be able to run tests directly with a review request. Is it the expected behavior, or should I change ?

.github/workflows/external_contributions_tests.yaml
SaboniAmine: Let me know if I should update this :)

.github/workflows/cpu_tests.yaml
SaboniAmine: Sure!

.github/workflows/external_contributions_tests.yaml
SaboniAmine: FYI: this prevents from updating / testing workflows from a PR, which could be a bit annoying when iterating on a new pipeline.

.github/workflows/external_contributions_tests.yaml
johannaSommer: no that is perfect then, thanks!

.github/workflows/external_contributions_tests.yaml
johannaSommer: yes I double checked - intel can be removed

.github/workflows/external_contributions_tests.yaml
sharpenb: After tests, this is validated :)

PR #5, opened 2025-03-13
feat: loading/removing hyperparameter utility functions

johannaSommer:
## Description
This PR adds two small utility functions to the `SmashConfig` that allows for more concise handling of the hyperparameters.

## Related Issue
None.

## Type of Change

- [x] This change requires a documentation update --> will be autodocumented, no need to update manually

## How Has This Been Tested?
None.

## Checklist

## Additional Notes
Example:
```
smash_config = SmashConfig()
smash_config.load_dict({'cacher': 'deepcache', 'deepcache_interval': 4})
smash_config

> SmashConfig(
>  'cacher': 'deepcache',
>  'deepcache_interval': 4,
> )

smash_config.flush_configuration()
smash_config
> SmashConfig(
> )

```

### DIFF
--- a/tests/algorithms/test_combinations.py
+++ b/tests/algorithms/test_combinations.py
@@ -28,9 +28,7 @@ class CombinationsTester(AlgorithmTesterBase):
     def prepare_smash_config(self, smash_config: SmashConfig, device: str) -> None:
         """Prepare the smash config for the test."""
         smash_config["device"] = device
-
-        for key, value in self.config_dict.items():
-            smash_config[key] = value
+        smash_config.load_dict(self.config_dict)
 
 
 @pytest.mark.cuda


src/pruna/config/smash_config.py
```diff
@@ -178,6 +178,25 @@ def save_to_json(self, path: str) -> None:
         if self._data is not None:
             pruna_logger.info("Data detected in smash config, this will be detached and not reloaded...")
 
+    def load_dict(self, config_dict: dict) -> None:
+        """
+        Load a dictionary of hyperparameters into the SmashConfig.
```
gsprochette, line 183: It would be nice to have an example to know how to use it, like the one you gave in the PR description.

src/pruna/config/smash_config.py
```diff
@@ -178,6 +178,25 @@ def save_to_json(self, path: str) -> None:
...
+        Parameters
+        ----------
+        config_dict : dict
+            The dictionary to load into the SmashConfig.
+        """
+        # first load the algorithm settings
+        # otherwise fine-grained hyperparameters will not be active yet and we can not set them
+        # lambda returns False for keys in ALGORITHM_GROUPS (and False sorts before True)
+        for k, v in sorted(config_dict.items(), key=lambda item: item[0] not in ALGORITHM_GROUPS):
+            self[k] = v
```
gsprochette, line 210: maybe we can check if some keys are in ADDITIONAL_ARGS? Currently I think I could do load_dict({'tokenizer': tokenizer}), so two options:
- we keep this functionality but those have to be set before the ALGORITHM_GROUPS and hyperparameters, and update the method's name
- or we log a error or warning and ignore ADDTIONAL_ARGS because I think this is not intended

johannaSommer: thanks for the comment - since this function is specifically to load algorithm configurations I would suggest to discard them and log it, added that 👍🏻

src/pruna/config/smash_config.py
johannaSommer: done! 😊

PR #6, opened 2025-03-13
fix: `EvaluationAgent` metric computation bugs

johannaSommer:
## Description
This PR includes a few fixes for the full smashing & metrics computation pipeline:
- the `gpu_memory` metric is now accessible as a string (as input to `Task)` by setting a default for `mode`
- return formats of transformers logits is now float, fixing `inf` results for e.g. Perplexity computation
- ensuring that all metrics are returned as float, not e.g. torch Tensors or even Tensors on CUDA
- adding the functionality to raise an error for `move_to_device` when device movement fails
- adding exceptions to the inference handler registration, as some algorithms that reload the model into a different class can not be recognized as compatible with the transformers handler even though they are

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Verified locally with algorithms AWQ and GPTQ by computing Perplexity and GPUMemory metric.

## Checklist

## Additional Notes
None.

### DIFF
--- a/src/pruna/engine/handler/handler_utils.py
+++ b/src/pruna/engine/handler/handler_utils.py
@@ -19,6 +19,11 @@ from pruna.engine.handler.handler_inference import InferenceHandler
 from pruna.engine.handler.handler_standard import StandardHandler
 from pruna.engine.handler.handler_transformer import TransformerHandler
 
+HANDLER_EXCEPTIONS: dict[type[InferenceHandler], list[str]] = {
+    TransformerHandler: ["OptAWQForCausalLM", "AutoHQQHFModel", "TranslatorWrapper", "GeneratorWrapper"],
+    DiffuserHandler: ["OnediffWrapper", "AutoHQQHFDiffusersModel"],
+}
+
 
 def register_inference_handler(model: Any) -> InferenceHandler:
     """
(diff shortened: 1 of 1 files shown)


src/pruna/evaluation/evaluation_agent.py
```diff
@@ -88,6 +88,10 @@ def evaluate(self, model: Any) -> Dict[str, Any]:
                 )
         else:
             self.subsequent_model_results = results
+
+        for key, value in results.items():
+            if isinstance(value, torch.Tensor):
+                results[key] = value.item()
```
guennemann, line 94: I assume the goal is to convert to float if the value is currently of type torch.Tensor? What if the value is of a completely different type? It might be helpful to check the types more generally and give a warning of it is neither float nor toch.Tensor.

src/pruna/engine/handler/handler_utils.py
```diff
@@ -14,29 +14,71 @@
...
 
+HANDLER_EXCEPTIONS: dict[InferenceHandler, dict[str, set[str]]] = {
+    TransformerHandler: {
+        "quantizer": {"awq", "hqq"},
+        "compiler": {"ctranslate", "cgenerate"},
+    },
+    DiffuserHandler: {
+        "compiler": {"onediff"},
+        "quantizer": {"hqq_diffusers"},
+    },
```
guennemann: Probably better to base the exceptions not on the smash config entries, but based on the class.

begumcig: If we are sure that awq and hqq are explicitly used with I think this is also fine. But I agree that exception cases based on the class would be more explicit.

johannaSommer: thanks a lot, is adressed!

src/pruna/engine/handler/handler_utils.py
```diff
@@ -14,29 +14,67 @@
 
 from typing import Any
 
+from pruna import SmashConfig
 from pruna.engine.handler.handler_diffuser import DiffuserHandler
 from pruna.engine.handler.handler_inference import InferenceHandler
 from pruna.engine.handler.handler_standard import StandardHandler
 from pruna.engine.handler.handler_transformer import TransformerHandler
 
+HANDLER_EXCEPTIONS: dict[InferenceHandler, dict[str, set[str]]] = {
```
guennemann: I believe you forgot to update the types because of the change. (dict[str, set[str]] is outdated.)

src/pruna/engine/handler/handler_utils.py
```diff
@@ -14,29 +14,67 @@
...
+
     if "diffusers" in model.__module__:
         return DiffuserHandler()
     elif "transformers" in model.__module__:
         return TransformerHandler()
     else:
         return StandardHandler()
+
+
+def scan_for_exceptions(model: Any, smash_config: SmashConfig) -> InferenceHandler | None:
```
guennemann: It seems that the smash_config is no longer required. Please remove.

PR #23, opened 2025-03-19
feat: enable combination of torch_compile and hqq_diffusers

johnrachwan123:
## Description
This PR enables the combination of `torch_compile` and `hqq_diffusers`. This allows the speedup of hqq quantized models in all precisions. This means that we can accelerate any diffusion model after quantization!

You can use it by simply defining the following config:

```
from pruna import SmashConfig

smash_config = SmashConfig()
smash_config['compiler'] = 'torch_compile'
smash_config['quantizer'] = 'hqq_diffusers'
# smash_config['torch_compile_mode'] = 'max-autotune' # Uncomment to enable extra speedups
```

This PR also includes many fixes to hqq_diffusers save/load functions. They were only working in 1/3 usecase (transformer-based models).

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
2 unit tests added that pass (1 unet-based and 1 transformer-based). Tutorial notebook added as well that is also passing.

## Checklist

## Additional Notes
This can also be extended to LLMs through the `hqq` method but is out of this scope of this PR.

### Small Benchmark on H100, SD1.4, 50 steps

| Model                          | E2E Time  |
|-------------------------------|-----------|
| base                          | 1.4s      |
| hqq (4 bits)                  | 2s        |
| hqq (8 bits)                  | 1.57s         |
| hqq (4 bits) + torch.compile  | 872ms     |
| hqq (8 bits) + torch.compile  | 830ms     |

### Visual Comparison

**Base image:**  
Base image

**8 bits image:**  
8 bits image

**4 bits image:**  
4 bits image

### DIFF
--- a/docs/tutorials/diffusion_quantization_acceleration.ipynb
+++ b/docs/tutorials/diffusion_quantization_acceleration.ipynb
@@ -1,0 +1,194 @@
+{
+    "cells": [
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "#  Speedup and Quantize any Diffusion Model"
+            ]
+        },
+        {
+            "cell_type": "raw",
+            "metadata": {
+                "vscode": {
+                    "languageId": "raw"
+                }
+            },
+            "source": [
+                "<a target=\"_blank\" href=\"https://colab.research.google.com/github/PrunaAI/pruna/blob/v|version|/docs/tutorials/sana_diffusers_int8.ipynb\">\n",
+                "    <img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/>\n",
+                "</a>"
+            ]
+        },
+        {
+            "cell_type": "raw",
+            "metadata": {
+                "raw_mimetype": "text/restructuredtext",
+                "vscode": {
+                    "languageId": "raw"
+                }
+            },
+            "source": [
+                "This tutorial demonstrates how to use the ``pruna`` package to optimize both the latency and the memory footprint of any diffusion model from the diffusers package.\n",
+                "We will use the ``Flux Dev`` model as an example, but this tutorial is working on any stable diffusion or flux model."
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "# if you are not running the latest version of this tutorial, make sure to install the matching version of pruna\n",
+                "# the following command will install the latest version of pruna\n",
+                "!pip install pruna"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "### 1. Loading the Diffusion Model\n",
+                "\n",
+                "First, load your diffusion model."
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "import torch\n",
+                "from diffusers import FluxPipeline\n",
+                "\n",
+                "pipe = FluxPipeline.from_pretrained(\"black-forest-labs/FLUX.1-schnell\", torch_dtype=torch.bfloat16, cache_dir=\"/efs/hf_cache\").to(\"cuda\")"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "### 2. Initializing the Smash Config"
+            ]
+        },
+        {
+            "cell_type": "raw",
+            "metadata": {
+                "raw_mimetype": "text/restructuredtext",
+                "vscode": {
+                    "languageId": "raw"
+                }
+            },
+            "source": [
+                "Next, initialize the smash_config (we make use, here, of the :doc:`hqq-diffusers </compression>` and :doc:`torch-compile </compression>` algorithms)."
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": 2,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "from pruna_pro import SmashConfig\n",
+                "\n",
+                "smash_config = SmashConfig()\n",
+                "smash_config['compiler'] = 'torch_compile'\n",
+                "smash_config['quantizer'] = 'hqq_diffusers'"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "### 3. Smashing the Model\n",
+                "\n",
+                "Now, smash the model. This can take up to 30 seconds."
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "from pruna import smash\n",
+                "\n",
+                "# Smash the model\n",
+                "pipe = smash(\n",
+                "    model=pipe,\n",
+                "    smash_config=smash_config,\n",
+                ")"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "### 4. Running the Model\n"
+            ]
+        },
+        {
+            "cell_type": "markdown",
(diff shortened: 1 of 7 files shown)


docs/tutorials/diffusion_quantization_acceleration.ipynb
```diff
@@ -0,0 +1,195 @@
...
+        },
+        {
+            "cell_type": "raw",
+            "metadata": {
+                "vscode": {
+                    "languageId": "raw"
+                }
+            },
+            "source": [
+                "<a target=\"_blank\" href=\"https://colab.research.google.com/github/PrunaAI/pruna/blob/v|version|/docs/tutorials/sana_diffusers_int8.ipynb\">\n",
```
llcnt: Should we change the link to a different colab ? :)

src/pruna/engine/save.py
```diff
@@ -229,13 +227,17 @@ def save_model_hqq_diffusers(model: Any, model_path: str, smash_config: SmashCon
     hf_quantizer = HQQDiffusersQuantizer()
     AutoHQQHFDiffusersModel = construct_base_class(hf_quantizer.import_algorithm_packages())
     if hasattr(model, "transformer"):
-        AutoHQQHFDiffusersModel.save_quantized(model.transformer, model_path + "/transformer_quantized")
-        del model.transformer
+        AutoHQQHFDiffusersModel.save_quantized(model.transformer, os.path.join(model_path, "backbone_quantized"))
+        transformer_backup = model.transformer
+        model.transformer = None
```
llcnt, line 233: Could we avoid setting it to None ?
It raises this warning when I test it : 
image
Maybe just `del model.transformer` would work, no ? ;)

llcnt (approved):
Only minor changes/questions, but otherwise looks good :)

docs/tutorials/diffusion_quantization_acceleration.ipynb
johnrachwan123: Lol nice catch

src/pruna/engine/save.py
johnrachwan123: I think this info message is being thrown regardless if we set it to None or del it. Both are equivalent. This is being thrown during the save function 😉

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -118,7 +122,11 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         if cacher_type in compilation_map:
             return compilation_map[cacher_type](model, smash_config)
 
-        # If the model does not have a helper, we compile the model itself
+        if hasattr(model, "transformer") and isinstance(model.transformer, tuple(get_diffusers_transformer_models())):
```
sharpenb: These lines can be factorized. Another simplification would be that if we can have respective "is_diffusers_trnasformer" and "is_unet_trnasformer" in the `model_checks`

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -209,6 +217,31 @@ def deepcache_logic(model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+        The model to compile.
+    smash_config : SmashConfigPrefixWrapper
+        Configuration settings for compilation.
+
+    Returns
+    -------
+    Any
+        The compiled model.
+    """
+    if hasattr(model, "transformer"):
```
sharpenb, line 237: The function is consistent with the previous compilation standard. A future task (already on the todo list) is to properly manage modules to avoid having custom function for each architecture type.

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -209,6 +217,31 @@ def deepcache_logic(model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
     return model
 
 
+def unet_transformer_pipeline_logic(model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
+    """
+    Apply compilation logic for HQQ models.
```
sharpenb: It is not clear if the function name and the comment are consistent. The function does not look specific to HQQ.

Since the `logic` functions actually compile the mode, a suggestion would be to rename them `compile_XYZ`

src/pruna/algorithms/quantization/hqq_diffusers.py
```diff
@@ -156,12 +157,10 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         )
 
         # Prepare the model for fast inference
-        try:
-            if smash_config["weight_bits"] == 4:
-                imported_modules["prepare_for_inference"](working_model, backend=smash_config["backend"])
-        except Exception as e:
-            pruna_logger.error(f"Error: {e}")
-            pass
+        if smash_config["weight_bits"] == 4 and next(iter(working_model.parameters())).dtype == torch.bfloat16:
```
sharpenb: Could we add a comment why the backend argument is not supported in the else case? :)

src/pruna/engine/load.py
```diff
@@ -285,31 +285,19 @@ def load_hqq_diffusers(path: str, **kwargs) -> Any:
     hf_quantizer = HQQDiffusersQuantizer()
     AutoHQQHFDiffusersModel = construct_base_class(hf_quantizer.import_algorithm_packages())
 
-    # if it is a diffusers model, it saves the model_index.json file
-    if not os.path.exists(os.path.join(path, "model_index.json")):
-        model = AutoHQQHFDiffusersModel.from_quantized(
-            path, **filter_load_kwargs(AutoHQQHFDiffusersModel.from_quantized, kwargs)
-        )
-    else:
+    if os.path.exists(os.path.join(path, "backbone_quantized")):
```
sharpenb, line 289: We lost the comments explaining the loading logic :)

src/pruna/engine/save.py
```diff
@@ -229,13 +227,17 @@ def save_model_hqq_diffusers(model: Any, model_path: str, smash_config: SmashCon
...
         model.save_pretrained(model_path)
+        model.transformer = transformer_backup
```
sharpenb, line 236: Could we add a comment saying "Save the quantized backbone" and "save the rest" or sthg like this

sharpenb (changes requested):
Nice PR! I left some comments. I would suggest to add in the PR description a disclaimer that the speedup were observed on H100 with Flux but not on other hardware and models. If it was actually tested on more setups, let's indicate it as well :)

For a future PR, @louis could take over and check torch compile on more models (like llms) and more quantization methods :)

src/pruna/algorithms/compilation/torch_compile.py
johnrachwan123: conditions are merged but i don't necessarily agree with the second point since its unecessary code since they are anw a subset of torch module

src/pruna/algorithms/compilation/torch_compile.py
johnrachwan123: Updated the docstring thanks for the catch.

I think changing from _logic to compile_xyz is its own PR since it hits pruna_pro and pruna. Let's keep it in mind for a new small feature!

src/pruna/algorithms/compilation/torch_compile.py
johnrachwan123: Not sure what you mean by this but from what i understand there is nothing to act on. Let's discuss what you mean in a quick 1on1 !

src/pruna/algorithms/quantization/hqq_diffusers.py
johnrachwan123: Done

src/pruna/engine/save.py
johnrachwan123: Done

src/pruna/engine/load.py
johnrachwan123: Added back

sharpenb (approved):
Thanks for the fixes! It looks good to me :)

llcnt (approved):
Thank you for the fixes :)

Issue #21, opened 2025-03-17
[BUG] Pruna installation fails with Python 3.12


johannaSommer:
## Describe the bug
Pruna installation fails with Python 3.12 due to auto-gptq.

## What I did

```
conda create -n myenv python=3.12
pip install pruna

Collecting pruna
  Using cached pruna-0.2.0.post1-py3-none-any.whl.metadata (3.6 kB)
Collecting ConfigSpace (from pruna)
  Using cached configspace-1.2.1.tar.gz (130 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Collecting DeepCache (from pruna)
  Using cached DeepCache-0.1.1-py3-none-any.whl.metadata (16 kB)
Collecting auto-gptq (from pruna)
  Using cached auto_gptq-0.7.1.tar.gz (126 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      Building cuda extension requires PyTorch (>=1.13.0) being installed, please install PyTorch first: No module named 'torch'
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```

## Expected behavior
The installation should work.

## Screenshots
None.

## Environment
 - pruna version: 0.2.0.post1
 - python version: 3.12
 - Operating System: Ubuntu

## Additional context
None.

johannaSommer:
fixed in #41

PR #41, opened 2025-03-27
feat: switch autogptq to gptqmodel

johnrachwan123:
## Description
AutoGPTQ is now deprecated. This PR switches our GPTQ algorithm to gptqmodel instead.

## Related Issue
autogptq already caused problems with supporting python 3.12.

## Type of Change

## How Has This Been Tested?
integration test still runs.

## Checklist

## Additional Notes
I did not investigate the extent of the new repo. This is only to make the switch. We might be able to find some nice speedups with some investigations.

### DIFF
--- a/src/pruna/algorithms/quantization/gptq_model.py
+++ b/src/pruna/algorithms/quantization/gptq_model.py
@@ -57,7 +57,7 @@ class GPTQQuantizer(PrunaQuantizer):
         return [
             OrdinalHyperparameter(
                 "weight_bits",
-                sequence=[2, 3, 4],
+                sequence=[2, 3, 4, 8],
                 default_value=4,
                 meta=dict(desc="Sets the number of bits to use for weight quantization."),
             ),
--- a/src/pruna/algorithms/quantization/gptq_model.py
+++ b/src/pruna/algorithms/quantization/gptq_model.py
@@ -22,7 +22,6 @@ from pruna.config.smash_config import SmashConfigPrefixWrapper
 from pruna.config.smash_space import Boolean
 from pruna.data.utils import recover_text_from_dataloader
 from pruna.engine.model_checks import is_causal_lm
-from pruna.engine.save import SAVE_FUNCTIONS
 from pruna.engine.utils import safe_memory_cleanup
 
 
@@ -44,6 +43,7 @@ class GPTQQuantizer(PrunaQuantizer):
     run_on_cuda = True
     dataset_required = True
     compatible_algorithms = dict()
+    required_install = "``pip install pruna[gptq]``"
 
     def get_hyperparameters(self) -> list:
         """
@@ -132,7 +132,6 @@ class GPTQQuantizer(PrunaQuantizer):
             model.quantize(calib_data, batch_size=smash_config["batch_size"])
             model.save(temp_dir)
             model = imported_modules["GPTQModel"].load(temp_dir)
-            model.model_local_path = "facebook/opt-125m"
 
         return model
 
--- a/src/pruna/algorithms/quantization/gptq_model.py
+++ b/src/pruna/algorithms/quantization/gptq_model.py
@@ -16,13 +16,13 @@ import tempfile
 from typing import Any, Dict
 
 from ConfigSpace import OrdinalHyperparameter
-from transformers import AutoModelForCausalLM, GPTQConfig
 
 from pruna.algorithms.quantization import PrunaQuantizer
 from pruna.config.smash_config import SmashConfigPrefixWrapper
 from pruna.config.smash_space import Boolean
 from pruna.data.utils import recover_text_from_dataloader
 from pruna.engine.model_checks import is_causal_lm
+from pruna.engine.save import SAVE_FUNCTIONS
 from pruna.engine.utils import safe_memory_cleanup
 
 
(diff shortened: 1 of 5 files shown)


src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -44,6 +43,7 @@ class GPTQQuantizer(PrunaQuantizer):
     run_on_cuda = True
     dataset_required = True
     compatible_algorithms = dict()
+    required_install = "``pip install pruna[gptq]``"
```
johannaSommer: can you add a comment here (and in the pyproject.toml) stating that this always has to be a seperate install due to https://github.com/ModelCloud/GPTQModel/issues/1466

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -102,35 +112,28 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
             The quantized model.
         """
+        imported_modules = self.import_algorithm_packages()
         with tempfile.TemporaryDirectory(prefix=smash_config["cache_dir"]) as temp_dir:
             # cast original model to CPU to free memory for smashed model
             if hasattr(model, "to"):
                 model.to("cpu")
                 safe_memory_cleanup()
             model.save_pretrained(temp_dir)
+            smash_config.tokenizer.save_pretrained(temp_dir)
```
johannaSommer, line 120: this is probably because the tokenizer needs to be present in the repository, can we add this as explanation here? as it doesnt follow the usual structure

tests/algorithms/testers/quantization.py
```diff
@@ -1,13 +1,13 @@
 import pytest
 
+from pruna.algorithms.quantization.gptq_model import GPTQQuantizer
```
johannaSommer, line 4: while you at it (sorry but it needs to be done) can you add a proper post_smash_hook for gptq please? thanks!

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -102,35 +112,28 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
             )
 
-            smashed_model = AutoModelForCausalLM.from_pretrained(
-                temp_dir,
-                quantization_config=gptq_config,
-                trust_remote_code=True,
-                device_map="auto",
-                torch_dtype="auto",
-            )
+            model = imported_modules["GPTQModel"].load(temp_dir, gptq_config)
```
johannaSommer, line 129: check with the handler - execute at least one evaluation. This will likely require a handler exception :/

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -57,17 +57,27 @@ def get_hyperparameters(self) -> list:
...
+                default=True,
+                meta=dict(desc="Whether to use exllama for quantization."),
+            ),
             OrdinalHyperparameter(
                 "group_size",
                 sequence=[64, 128, 256],
                 default_value=128,
                 meta=dict(desc="Group size for quantization."),
             ),
+            OrdinalHyperparameter(
```
johannaSommer: remove this for now and use batch size in smash config - we will adjust this better in the future

johannaSommer (changes requested):

johannaSommer:
Linking #21

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -57,17 +57,27 @@ def get_hyperparameters(self) -> list:
...
+                sequence=[2, 3, 4, 8],
+                default_value=4,
                 meta=dict(desc="Sets the number of bits to use for weight quantization."),
             ),
-            Boolean("use_exllama", default=True, meta=dict(desc="Whether to use exllama for quantization.")),
+            Boolean(
+                "use_exllama",
+                default=True,
+                meta=dict(desc="Whether to use exllama for quantization."),
+            ),
```
llcnt, line 71: I think we can remove this hyperparameter, if it is deprecated in gptqmodel ;)

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -44,6 +43,7 @@ class GPTQQuantizer(PrunaQuantizer):
     run_on_cuda = True
```
llcnt, line 43: gptqmodel [repo](https://github.com/ModelCloud/GPTQModel?tab=readme-ov-file#what-is-gptqmodel) mentions that it can run on cpu also. However I did not test it, do you think we should change to `run_on_cpu=True`?

llcnt (approved):
Thank you for the switch! I added only 2 minors comments, and we are good :)

src/pruna/algorithms/quantization/gptq_model.py
johannaSommer: if this is unused we should still deprecate it properly, meaning it stays here and there is a warning in the apply and the argument is not used anymore

johnrachwan123: I think it still exists

tests/algorithms/testers/quantization.py
johnrachwan123: Good ?

src/pruna/algorithms/quantization/gptq_model.py
johnrachwan123: I tried to run it on cpu and it failed during inference time. This needs more investigation as it should be possible to make it work but the way we quantize the models currently might not be the best.

johannaSommer (approved):

PR #43, opened 2025-03-31
ci: add installation tests for all supported OS and python versions

johannaSommer:
## Description
This PR adds workflows to check the installation of `pruna` on various OS and Python versions to avoid issues like #21 in the future. I am open for discussion on this but generally felt like this is not something that should happen on every commit, but rather on dispatch when the PR necessitates it and once as a sanity check when merging to main.

## Related Issue
Avoid #21 in the future

## Type of Change

## How Has This Been Tested?
Ran the workflows on this PR. Currently Python 3.12 on ubuntu is failing which is intended, the rest pass.

## Checklist

## Additional Notes
None.

### DIFF
--- a/.github/workflows/installation.yaml
+++ b/.github/workflows/installation.yaml
@@ -32,13 +32,17 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install Poetry
+      - name: Install Poetry on Windows
+        if: runner.os == 'Windows'
         run: |
-          curl -sSL https://install.python-poetry.org | python3 -
-          echo "${HOME}/.local/bin" >> $GITHUB_PATH
+          powershell -Command "(Invoke-WebRequest -Uri https://install.python-poetry.org -UseBasicParsing).Content | python - --version 1.8.5"
+          echo "$(python -m site --user-base)\Scripts" >> $env:GITHUB_PATH
 
-      - name: Configure Poetry to not create virtual environments
-        run: poetry config virtualenvs.create false
+      - name: Install Poetry on Unix
+        if: runner.os != 'Windows'
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/installation.yaml
```diff
@@ -0,0 +1,45 @@
...
+          echo "$(python -m site --user-base)\Scripts" >> $env:GITHUB_PATH
+
+      - name: Install Poetry on Unix
+        if: runner.os != 'Windows'
+        run: |
+          curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.8.5 python3 -
+          echo "$(python3 -m site --user-base)/bin" >> $GITHUB_PATH
+
+      - name: Install dependencies
+        run: pip install -e .
```
johnrachwan123, line 45: Can we try it on [full] ?

.github/workflows/installation.yaml
```diff
@@ -0,0 +1,45 @@
...
+    branches:
+      - main
+
+jobs:
+  installation:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-latest, windows-latest]
```
johnrachwan123: can we add ubuntu-latest here as well ?

johannaSommer: added 👍🏻

.github/workflows/installation.yaml
johannaSommer: I tried it but realized that all packages in [full] are CPU-only, so I'd suggest we skip full and just test the core version here

SaboniAmine (approved):
LGTM thanks!

johnrachwan123 (approved):
LGTM!

PR #26, opened 2025-03-20
feat: add model loading and saving functionality from Hugging Face hub

davidberenstein1957:
## Description
- Introduced `load_pruna_model_from_hub` to load models directly from the Hugging Face hub.
- Added `save_pruna_model_to_hub` to save models along with a model card template.
- Updated `PrunaModel` class to include methods for loading and saving from/to the hub.
- Improved code formatting and logging for better readability and debugging.
- 
## Related Issue
Closes #24 

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
tests WIP

Repo: https://huggingface.co/davidberenstein1957/my_smashed_model

```python
from transformers import AutoModelForCausalLM, AutoTokenizer

from pruna import PrunaModel, SmashConfig, smash

model_id = "facebook/opt-125m"
model = AutoModelForCausalLM.from_pretrained(
    model_id, torch_dtype="auto", device_map="mps"
)

text = "The 45th president of the United States of America is"

tokenizer = AutoTokenizer.from_pretrained(model_id)
ins = tokenizer(text, return_tensors="pt")

# Initialize the SmashConfig
smash_config = SmashConfig()
smash_config.add_tokenizer(model_id)
# smash_config.add_data("WikiText")
# smash_config["quantizer"] = "gptq"


# Smash the model
smashed_model = smash(model=model, smash_config=smash_config, experimental=True)

# Save the model to the hub
smashed_model.save_pretrained("smashed_model")
smashed_model.save_to_hub(
    repo_id="davidberenstein1957/my_smashed_model",
    folder_path="smashed_model",
)

# Load the model from the hub
pruna_model = PrunaModel.from_hub(repo_id="davidberenstein1957/my_smashed_model")
```

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #44, opened 2025-04-01
feat: Add Hugging Face integration to save and load models

### LINKED ISSUE #24: [FEATURE] Add `push_to_hub` method to `smash`
I want to push smashed models to the Hub.

```
from pruna import smash

# Smash the model
smashed_model = smash(
    model=model,
    smash_config=smash_config,
)
smashed_model.push_to_hub
```

## Additional context
from_pretrained is already available

davidberenstein1957:
## Description
- Introduced `load_pruna_model_from_hub` to load models directly from the Hugging Face hub.
- Added `save_pruna_model_to_hub` to save models along with a model card template.
- Updated `PrunaModel` class to include methods for loading and saving from/to the hub.
- Improved code formatting and logging for better readability and debugging.
- 
## Related Issue
Closes #24 
Related to #26 

## Type of Change

## How Has This Been Tested?
tests WIP

Repo: https://huggingface.co/davidberenstein1957/my_smashed_model

```python
from transformers import AutoModelForCausalLM, AutoTokenizer

from pruna import PrunaModel, SmashConfig, smash

model_id = "facebook/opt-125m"
model = AutoModelForCausalLM.from_pretrained(
    model_id, torch_dtype="auto", device_map="mps"
)

text = "The 45th president of the United States of America is"

tokenizer = AutoTokenizer.from_pretrained(model_id)
ins = tokenizer(text, return_tensors="pt")

# Initialize the SmashConfig
smash_config = SmashConfig()
smash_config.add_tokenizer(model_id)
# smash_config.add_data("WikiText")
# smash_config["quantizer"] = "gptq"


# Smash the model
smashed_model = smash(model=model, smash_config=smash_config, experimental=True)

# Save the model to the hub
smashed_model.save_pretrained("smashed_model")
smashed_model.save_to_hub(
    repo_id="davidberenstein1957/my_smashed_model",
    folder_path="smashed_model",
)

# Load the model from the hub
pruna_model = PrunaModel.from_hub(repo_id="davidberenstein1957/my_smashed_model")
```

## Checklist

### DIFF
--- a/src/pruna/engine/model_card_template.md
+++ b/src/pruna/engine/model_card_template.md
@@ -1,0 +1,45 @@
+---
+library_name: {library_name}
+tags:
+- pruna-ai
+---
+
+# Model Card for {repo_id}
+
+This model was created using the [pruna](https://github.com/PrunaAI/pruna) library. Pruna is a model optimization framework built for developers, enabling you to deliver more efficient models with minimal implementation overhead.
+
+## Usage
+
+You can load this model using the following code:
+
+```python
+from pruna import PrunaModel
+
+loaded_model = PrunaModel.from_hub("{repo_id}")
+```
+
+After loading the model, you can use the inference methods of the original model.
+
+## Model Configuration
+
+The configuration of the model is stored in the `config.json` file.
+
+```bash
+{model_config}
+```
+
+## Smash Configuration
+
+The configuration of the model is stored in the `smash_config.json` file.
+
+```bash
+{smash_config}
+```
+
+## 🌍 Join the Pruna AI community!
+
+[![Twitter](https://img.shields.io/twitter/follow/PrunaAI?style=social)](https://twitter.com/PrunaAI)
+[![GitHub](https://img.shields.io/github/followers/PrunaAI?label=Follow%20%40PrunaAI&style=social)](https://github.com/PrunaAI)
+[![LinkedIn](https://img.shields.io/badge/LinkedIn-Connect-blue)](https://www.linkedin.com/company/93832878/admin/feed/posts/?feedType=following)
+[![Discord](https://img.shields.io/badge/Discord-Join%20Us-blue?style=social&logo=discord)](https://discord.com/invite/rskEr4BZJx)
+[![Reddit](https://img.shields.io/reddit/subreddit-subscribers/PrunaAI?style=social)](https://www.reddit.com/r/PrunaAI/)
--- a/src/pruna/engine/save.py
+++ b/src/pruna/engine/save.py
@@ -11,17 +11,19 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 import json
 import os
 import shutil
+import tempfile
 from copy import deepcopy
 from enum import Enum
 from functools import partial
-from typing import Any
+from pathlib import Path
+from typing import Any, List
 
 import torch
 import transformers
+from huggingface_hub import upload_large_folder
 
 from pruna.config.smash_config import SMASH_CONFIG_FILE_NAME, SmashConfig
 from pruna.engine.load import (
@@ -69,7 +71,9 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) -> 
 
     # in the case of multiple, specialized save functions, we default to pickled
     else:
-        pruna_logger.debug(f"Several save functions stacked: {smash_config.save_fns}, defaulting to pickled")
+        pruna_logger.debug(
+            f"Several save functions stacked: {smash_config.save_fns}, defaulting to pickled"
+        )
         save_fn = SAVE_FUNCTIONS.pickled
         smash_config.load_fn = LOAD_FUNCTIONS.pickled.name
     # execute selected save function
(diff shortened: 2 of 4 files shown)


src/pruna/engine/model_card_template.md
```diff
@@ -0,0 +1,45 @@
+---
```
johannaSommer: looks amazing! I was thinking it might be better to put this into `engine/hf_hub_utils/model_card_template.md`, what do you think? :)

src/pruna/engine/model_card_template.md
```diff
@@ -0,0 +1,45 @@
...
+from pruna import PrunaModel
+
+loaded_model = PrunaModel.from_hub("{repo_id}")
+```
+
+After loading the model, you can use the inference methods of the original model.
+
+## Smash Configuration
+
+The configuration of the model is stored in the `smash_config.json` file.
```
johannaSommer: Should we say here that this isnt the "model configuration" but rather "compression configuration"? Otherwise it doubles with the model configuration below

src/pruna/engine/model_card_template.md
```diff
@@ -0,0 +1,45 @@
...
+- pruna-ai
+---
+
+# Model Card for {repo_id}
+
+This model was created using the [pruna](https://github.com/PrunaAI/pruna) library. Pruna is a model optimization framework built for developers, enabling you to deliver more efficient models with minimal implementation overhead.
+
+## Usage
+
+You can load this model using the following code:
```
johannaSommer: should we put a `pip install pruna` here?

src/pruna/engine/save.py
```diff
@@ -79,6 +81,89 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) ->
...
+        The ignore patterns.
+    num_workers : int | None, optional
+        The number of workers.
+    print_report : bool, optional
+        Whether to print the report.
+    print_report_every : int, optional
+        The print report every.
+    """
+    # Create a temporary directory within the specified folder path to store the model files
+    with tempfile.TemporaryDirectory(dir=folder_path) as temp_dir:
```
johannaSommer: such a clean function, I love it! 🙌

src/pruna/config/smash_config.py
```diff
@@ -167,6 +167,10 @@ def save_to_json(self, path: str) -> None:
         for name in ADDITIONAL_ARGS:
             config_dict[name] = getattr(self, name)
 
+        # do not save the old cache directory or device
+        config_dict.pop("cache_dir")
+        config_dict.pop("device")
```
johannaSommer: why is it necessary to remove the device here? This is breaking quite some thing at the moment, maybe a bit of reasoning behind it - if we save a model with some specific configurations, it is easier for us to reapply them (because they are very quick to apply) rather than maintain a lot of complexity saving it in its smashed state. For this, we apply a quick smash after loading, which needs the device to be set to its original value. I would propose to keep it in the config and thing of it like "compressed for this device"

johannaSommer: also for me at the moment lines 141 and 142 in this file fail because it can not find the cache dir, but I think dropping the cache dir when saving is completely fine and we can just adjust lines 141 and 142 to be
```python
if name == "cache_dir":
      continue
```

tests/engine/test_load.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+
+from pruna.engine.pruna_model import PrunaModel
+
+
+@pytest.mark.parametrize(
```
johannaSommer, line 5: sorry this might not be well documented at the moment, but our pytests rely on having either a `@pytest.mark.cpu` or `@pytest.mark.cuda` to be included in our runs, I think both the loading and the saving tests can be marked as cpu :)

docs/user_manual/save_load.rst
```diff
@@ -66,7 +100,7 @@ you should also load the smashed model as follows:
 
     loaded_model = PrunaModel.from_pretrained("saved_model/", torch_dtype=torch.float16)
 
-Depending on the saving function of the algorithm combination not all keyword arguments are required for loading (e.g. some are set by the algorithm combination itself). 
+Depending on the saving function of the algorithm combination not all keyword arguments are required for loading (e.g. some are set by the algorithm combination itself).
 In that case, we discard and log a warning about unused keyword arguments.
```
johannaSommer, line 104: since we now have so much cool functionality in the Pruna Model, it might make sense to use autodoc to document them? How about adding somewhere (similar to the smashconfig and smash docs):
```
.. autoclass:: pruna.engine.pruna_model.PrunaModel
   :members: from_pretrained, from_hub, save_to_hub, save_pretrained
```

johannaSommer (changes requested):
Looks amazing already! I have a few small comments and one thing we should change about the smash config saving, otherwise we are good to go!

src/pruna/config/smash_config.py
davidberenstein1957: Done. I did not think about the device specific compresion directly, but reasoned from a user perspective, where the device might come across as weird, although we do get a warning about devices not mathcing, so that should be fine.

src/pruna/config/smash_config.py
```diff
@@ -167,6 +168,8 @@ def save_to_json(self, path: str) -> None:
         for name in ADDITIONAL_ARGS:
             config_dict[name] = getattr(self, name)
 
+        # do not save the old cache directory or device
+        config_dict.pop("cache_dir")
```
gsprochette: Shouldn't this be the same as lines142-143, check if "cache_dir" is in config dict and del?

src/pruna/engine/save.py
```diff
@@ -79,6 +81,89 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) ->
     smash_config.save_to_json(model_path)
 
 
+def save_pruna_model_to_hub(
+    model: Any,
+    smash_config: SmashConfig,
+    repo_id: str,
+    folder_path: str | Path | None = None,
```
gsprochette: The function looks great, I just have a question about the folder_path argument. We save the model to folder_path/temp_dir_name with save_pruna_model, is there a case where the user would want to control where the temp dir is created ? My understanding is that if the user provides a path they probably want to find the model there afterwards, so it shouldn't be a temporary dir, right? I would be in favor of saving to a temp dir if we have no path, and saving directly to the path without a temp dir otherwise:
- rename folder_path to model_path to match save_pruna_model arguments
- by default, we create a tempdir using smash_config.cache_dir_prefix
- if the user provided a path, we make a true directory without the need to clean it up afterwards.

This way if the user wants to both save locally and to the hub, this function takes care of both without the need to actually save the model twice.

gsprochette (commented):
The feature looks great already, just one style change and one discussion about the save_to_hub arguments :+1: :sparkles:

johannaSommer (approved):
LGTM! 🌻

src/pruna/engine/save.py
davidberenstein1957: Done :)

gsprochette (approved):
LGTM! There's still the typing issue where you need to add PATH as accepted types for the other saving functions, but once it's done it looks ready to merge :sparkles:

davidberenstein1957:
@johannaSommer, it seems something from the CICD is stuck. Perhaps you have god powers to override?

PR #45, opened 2025-04-01
feat: enable bnb with torch compile

llcnt:
## Description
Small PR that enable the combination of `diffusers_int8` quantizer and `torch_compile` for diffusers model.
The change in the code is minimal, but the results are interesting:
- following the previous work on HQQ (#23), we quantize and compile the denoiser inside the diffusers pipeline;
- I tested to compile the forward function of the denoiser instead (ie the call function of the unet or transformer), but the inference is not accelerated;
- I tested the inference time for 50 denoising steps on a Flux-8B-freepik model on a L40S GPU and obtained $23.63s$. After Bnb quantization, this number increases to $28.48s$. After quantization+compilation, we can reduce this number to $23.69s$;
- Note that the default compilation (ie capturing the full graph) will raise an error on a bnb-quantized model (because bnb will dynamically look for outliers in the inputs. JIT compilation does not support dynamic input with the default torch.compile parameters). To bypass this, we should set `smash_config['torch_compile_fullgraph'] = False`
- I also tested to compile (both the model and the forward) of a model quantized with `quanto`, but it fails because quanto uses 'fake_tensors' that are not compatible with torch.compile.

## Type of Change

## How Has This Been Tested?

I added a combination test on a sd model: 
image
Other unit test are still valid, and I have tested to save and re-load the model that was quantized+compiled in this [notebook](https://drive.google.com/file/d/1Wf623h9Lzjl-XnUF--jHj_Edk7XxRBKi/view?usp=sharing).

## Checklist

## Additional Notes

Similar results can be obtained in `pruna_pro` with the combination of HIGGS quantizer and torch.compile (but with better memory reduction! you can go up to $3$ or $2$ bits):
- Initial inference time after quantization is $31.40s$, and can be reduced to $25.65s$ after compilation;
- current experiments are run to improve this speedup, and also enjoy speedup for batch inference.

### DIFF
--- a/src/pruna/algorithms/compilation/torch_compile.py
+++ b/src/pruna/algorithms/compilation/torch_compile.py
@@ -45,7 +45,7 @@ class TorchCompileCompiler(PrunaCompiler):
     run_on_cuda = True
     dataset_required = False
     compatible_algorithms = dict(
-        quantizer=["half", "hqq_diffusers"],
+        quantizer=["half", "hqq_diffusers", "diffusers_int8"],
         cacher=["deepcache"],
     )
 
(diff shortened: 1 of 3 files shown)


sharpenb (approved):
Thanks for the PR! Looks lgtm :)

PR #51, opened 2025-04-07
feat: compile forward pass llm

llcnt:
## Description

This PR makes `torch.compile` more suitable to LLMs. We do not compile the whole model (that contains complex dynamic functions), but only the forward pass that computes one token. This change is particularly useful for quantized LLMs: we can obtain speedups by composing `quantization+torch.compile`.
The main code changes are the following:
- add a `causal_lm_logic` in our existing torchcompile file;
- define a `StaticCache` (dynamic KV cache is not compatible with torch compile), and re-define the `generate` function of the LLM with our custom one (max_kv_cache_size is the main bottleneck here: as we can not use dynamic cache, we have to use static cache but with a large enough size so that the user can generate up to max_kv_cache_size tokens);
- create a `utils.py` file where we design the `generate' function. By now we only support 2 input args (input_ids, max_new_tokens). We could add other args (the default generate function from hf contains dozens of args), but it would take time to test all of it.

The results of compatibility tests are the following:
- quanto is not compatible with torchcompile;
- Edit: HQQ is now compatible with torch.compile (see comment below);
- AWQ is not compatible (because AWQ uses a dedicated AWQforcausalLM that wraps the default function generate from hf);
- BnB is compatible when we set fullgraph=False (see #45 );
- GPTQ is compatible ;
- HIGGS (`pruna_pro` feature) is compatible.

## Type of Change

## How Has This Been Tested?

Latency for running inference with $56$ tokens generated for the prompt "This is a test of this large language model", on a meta-llama/Llama-3.2-1b-Instruct model:
| **Method**      | **GPU**       | **Before Quantization** | **After Quantization** | **After Quantization + Compilation** |
|-----------------|------------------|--------------------------|-------------------------|-------------------------------|
| **BitsAndBytes**| L40S       |        0.83                  |         2.20                |           4.89(fullgraph=false)                    |
|                 | A10G     |           0.94               |           2.41              |             5.24(fullgraph=false)                  |
| **GPTQ**        | L40S       |         0.83                 |          0.76               |             0.75                  |
|                 | A10G     |              0.94            |         0.90                |            0.89                   |
| **HIGGS**       | L40S       |          0.83                |          1.04               |            0.42                   |
|                 | A10G     |           0.94               |          1.23               |            0.48                   |

I have re-tested all methods (w/ and w/o manual cudagraphs after the code refactoring, and the speedups are the same). I have also tested HQQ and HIGGS on a larger llama-3.1-8B model o a single L40S GPU:
| **Method**      |  **Before Quantization** | **After Quantization + Compilation** |
|-----------------|------------------|----------------------------------------------------|
| **HQQ**|       1.59                  |          0.50                   |
| **HIGGS**       |         1.59              |            0.54                  |

I added unit tests on a llama-1B (because opt125 is not compatible)
image

Experiments can be reproduced with this [notebook](https://drive.google.com/file/d/1GaQTVdWuF8RcjhP2n-ZcjuqomQogmP9B/view?usp=sharing).

## Checklist

## Additional Notes

- opt models are not compatible (because they dont have a 'cached_postion' arg), so an exception check was created to avoid compiling an opt model that is quantized (IMO it is ok as opt models are currently outdated)


src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -45,7 +49,7 @@ class TorchCompileCompiler(PrunaCompiler):
     run_on_cuda = True
     dataset_required = False
     compatible_algorithms = dict(
-        quantizer=["half", "hqq_diffusers", "diffusers_int8"],
+        quantizer=["half", "hqq_diffusers", "diffusers_int8", "gptq", "llm_int8"],
```
sharpenb: The PR description says that Higgs is compatible with it. Should it then be here? :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,151 @@
...
+    if top_k is not None:
+        v, _ = torch.topk(logits, min(top_k, logits.size(-1)))
+        pivot = v.select(-1, -1).unsqueeze(-1)
+        logits = torch.where(logits < pivot, -float("Inf"), logits)
+    probs = torch.nn.functional.softmax(logits, dim=-1)
+    return probs
+
+
+@torch.compile
+def sample(logits, temperature: float = 1.0, top_k: Optional[int] = None):
```
sharpenb: I see now that it is restricted to one sampling strategy while there could be multiple ones (see [here](https://huggingface.co/docs/transformers//generation_strategies)). I wuold suggest ot check if we cna inherit from theirs and use the same trick to compile all the options.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            input_ids (torch.Tensor): The input ids to generate from
+            max_new_tokens (int): The maximum number of new tokens to generate
+
+        Returns:
+        -------
+            torch.Tensor: The generated ids
+        """
+        batch_size, seq_length = input_ids.shape
```
sharpenb: This function would be enfit from commenting to explain the logic. E.g. it is unclear where the past_kv is updated. It looks like when generating a new token the kv for this token is not update in the past kv.

tests/fixtures.py
```diff
@@ -107,7 +107,15 @@ def get_automodel_transformers(model_id: str) -> tuple[Any, SmashConfig]:
     except Exception:
         smash_config.add_tokenizer("bert-base-uncased")
 
-    smash_config.add_data("WikiText")
+    # quantizers that require calibration will throw an error if the tokenizer does not have a pad token
+    tokenizer_for_calibration = AutoTokenizer.from_pretrained(model_id)
```
sharpenb: What would be the reasoning behind this test change? :) I was wondering if this should be applied to all causallm tests.
- Why do we need to add a pad token?
- Why do we need to limit the dataset size?

sharpenb (changes requested):
Thanks for the PR! The PR description with the benchmark is very promising :) In a future PR, I feel that we would benefit to have more llm tests on small llms (beyond opt).

src/pruna/algorithms/compilation/torch_compile.py
llcnt: No we dont need, compatibilities are un-directed in between pruna and pruna_pro ;)

tests/fixtures.py
llcnt: There is no need to change this if we test quantizer that requires calibration on opt125 (that is the case for our previous tests). However if you test a llama, it will fail beacause the tokenizer does not have a pad token (we did not have any trouble before because when we were testing llama, eg on HQQ, we did not use calibration, hence no tokenizer) :)
- default llama1B tokenizer does not have a pad token;
- we can delete this limit, but the unit will be slower (considering that we test here a llama1b model and not an opt125)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes we can :) But as I mentioned in the main comment, we would have to write by ourself each if statement (corresponding to each generation strategy; as it is done in`transformers` [here](https://github.com/huggingface/transformers/blob/7ecc5b88c0328aea91a3c9f8763f56b3b1e26767/src/transformers/generation/utils.py#L2101) ) and to test them all. Should we spend time on this now, or postpone this work for a future PR? (I would be in favor of the latter for keeping this PR as simple as possible ahah ;) )

src/pruna/algorithms/compilation/utils.py
llcnt: The KV cache is handled automatically by the model.forward, no need to manually update it ;)
I have added some comments in the code

tests/fixtures.py
llcnt: Edit: I integrated the fix from begum when rebasing, but I still need to limit the dataset size (otherwise it kills my cpu)

llcnt:
Edit2: I have added support for HQQ !
- We don't use our default forward compilation, but the one provide in HQQ where cuda graphs are managed manually;
- during my experiments I experienced cuda errors when using a cache_size for HQQ+compile < 400. I also experienced some bugs when the number of generated tokens is small (<100)
- I also experienced errors (reported [here](https://github.com/mobiusml/hqq/issues/157)) when using hqq-quantized from Automodel from hf, or when using a model in torch.float16: we need torch.bfloat16 and a model that uses AutoHQQ;
- The generator provided in HQQ directly converts input/output with the tokenizer corresponding to the given LLM. So I added some lines to convert this when necessary so that the user can follow the same pipeline `model.generate(input_tokenized, max_new_tokens)`, as it is the case in `transformers`.

I have also modified the hqq.py file because some users (including me ahah) have experienced some bugs (impossible to move the model with .to):
- I removed the `try:` statement that tries to use .to to the hqq-quantized model: it is un-valid by nature;
- I added a torch_dtype (in Automodel call) and a compute_dtype (in AutoHQQ call) arg to be able to use torch.bfloat16;
- I exchanged the AutoHQQ and AutoModel in the try statement so that, by default, we first try to apply AutoHQQ now. I did this because, for some reason, `prepare_for_infererence` did not change the HQQmAtmul when the model is quantized with AutoModel from `transformers`;
- I re-launched the unit tests on HQQ to be sure my change did not break anything and there are successful.

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -241,6 +265,67 @@ def unet_transformer_pipeline_logic(model: Any, smash_config: SmashConfigPrefixW
...
+        else:
+            pruna_logger.warning("No generation config found, using default values for top_k and temperature.")
+            # https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig.top_k
+            top_k = 50
+            # https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig.temperature
+            temperature = 1.0
+        past_kv = StaticCache(
+            model.config,
+            smash_config["batch_size"],
+            smash_config["max_kv_cache_size"],
```
sharpenb: This parameter simpply be the context length?

src/pruna/algorithms/compilation/utils.py
sharpenb: I feel now that it is a bit too many comments. Could we siply say (1) init generated token id and their position, (2) sample/decode tokens. One question is also why we cannot use compiled decoding to generate the first token?

src/pruna/algorithms/compilation/utils.py
sharpenb: Makes sense. I am still wondering if there is no function in hf that we can simply import/inherit and does the similar/same things? Now we reimplement the function to the lowest details.

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -241,6 +265,67 @@ def unet_transformer_pipeline_logic(model: Any, smash_config: SmashConfigPrefixW
...
+                    "`smash_config.add_tokenizer(model_id)`."
+                )
+        gen = imported_algorithms["HFGenerator"](
+                model,
+                tokenizer,
+                max_new_tokens=smash_config["max_kv_cache_size"],
+                do_sample=True,
+                compile="partial",
+                compile_options={"mode": smash_config["mode"], "fullgraph": smash_config["fullgraph"]},
+            ).enable_cuda_graph()
```
sharpenb: If I understand correctly, the compilation happens here now. When looking at [this](https://github.com/mobiusml/hqq/blob/1f052eb5a0aab0572d380d48b708ae1c74936d23/hqq/utils/generation_hf.py#L112), it looks like there is quite some redundant code with what we have in utils. Could we simply have our own equivalent of HFGenerator class that we use for all quantization methods? It would avoid this if HQQ and make maintenance easier.

sharpenb (changes requested):
Thanks for the nice udpate! I left some comments on the modified files. I feel that some code factorization is possible :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes I guess I can see what you have in mind. :) My initial plan was to get the "smallest" function to be compiled so that we can after "put it back into the model", and have something like "function_before_generate = our_custom_compiled_function", so that we can blindly use the default `generate()` function afterwards. Unfortunately it is not possible: we compile the function that gets as input the (previous) tokens, and generate a new token. The limitation comes from the fact that `transformers` does not re-uses any function between the different sampling strategies. Currently we have done `model.generate = our_custom_compiled_generate`, what we could do as a future work is to do :
- `model._sample = our_custom_compiled_generate_sample` (this is the function we currently "overwrite" by [default](https://github.com/huggingface/transformers/blob/e5ac23081ec4021818a21d7442d396f31de8c30c/src/transformers/generation/utils.py#L2452) when we do the model.generate replacement);
-  `model._beam_search= our_custom_compiled_generate_beam`
-  `model._assisted_decoding= our_custom_compiled_generate_assisted`
-  `model._dola_decoding= our_custom_compiled_generate_dola`
-  `model._contrastive_search= our_custom_compiled_generate_contrastive`
- `model._group_beam_search= our_custom_compiled_generate_groupbeam`
- `model._constrained_beam_search= our_custom_compiled_generate_constrainedbeam`

src/pruna/algorithms/compilation/utils.py
llcnt: I completely refactor so that now we have a HFGenerator and some functions I tried to comment :)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: I refactored the code to follow your idea ;) we have now a single generator class :)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: I want to answer Yes, but to be fair I must say that it is more strict: in full generality, a standard LLM with contextlength=10 can generate 20 tokens (it will just forget the first 10 tokens while generating the last tokens). However here we can not generate more than the KV cache can accept (because it is staticCache and we can not use DynamicCache with compilation...)

src/pruna/algorithms/compilation/utils.py
sharpenb: Thanks for the explanation! I think that this implementation is nice enough for this PR. For a future PR that will be able to compile any specified function I feel that another implementation would be possible. The implementation could make a new function that takes as input any function that would need to be compiled, create a replacement compiled function from it without detailed reimplementation (e.g. by wrapping the input function and compiling it, and using monkey patching) and replace it in the model. There would be two advantage from this implementation: easier maintenance through no code boilerplate, and capacity to compile functions at different levels :)

src/pruna/algorithms/compilation/torch_compile.py
sharpenb: Makes sense. Since LLM usually come with a `max_context_length` variable, I could see that we remove this as hyperparameter and simply set it to this value. If a user wants to restrict the cache size, they can simply cahnge the value of this vairable in the LLM they give as input :)

src/pruna/algorithms/compilation/torch_compile.py
sharpenb: Amazing, thanks! Did you rerun the quick benchmark to see that it does not change the speedup? It should not change anything but could be nice to check a couple of the table rows :)

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -241,6 +271,49 @@ def unet_transformer_pipeline_logic(model: Any, smash_config: SmashConfigPrefixW
...
+    gen = HFGenerator(
+            model,
+            max_kv_cache_size=smash_config["max_kv_cache_size"],
+            temperature=temperature,
+            top_k=top_k,
+            compile_mode=smash_config["mode"],
+            compile_fullgraph=smash_config["fullgraph"],
+            batch_size=smash_config["batch_size"],
+        )
+    # If we are using max-autotune-no-cudagraphs, we need to handle the cudagraphs manually.
```
sharpenb, line 310: This part is a bit unclear. It feels that by enabling the cuda graph we would do sthg similar to `max-autotune` (also when reading the [docs](https://pytorch.org/docs/stable/generated/torch.compile.html)). Could you elaborate on what is the benefit of this no cudagraphs mode? :)

src/pruna/algorithms/compilation/torch_compile.py
sharpenb: Could we also add the HQQ number to the table? :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+        torch.Tensor
+            The generated tokens.
+        """
+        for i in range(1, max_new_tokens):
+            next_token = self.gen_next_token(next_token)
+        output_tokens = self.generated_ids
+
+        if cleanup:
+            del self.inputs, self.generated_ids, self.cache_position
+            torch.cuda.empty_cache()
```
sharpenb, line 457: We might need as well `gc.collect()` to finish the cleaning.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+        self.generated_ids[:, self.cache_position] = next_token.int()
+        return next_token
+
+    def next_token_iterator(self, next_token: torch.Tensor, max_new_tokens: int, cleanup: bool = True) -> torch.Tensor:
+        """
+        Generate the next token.
+
+        Parameters
+        ----------
+        next_token : torch.Tensor
```
sharpenb: That sounds a bit unatural to me that the current token is called next_token. Also l.530 looks to input here prefill() which intuitvely gives the feeling that these are the previous tokens. We can probably make these variables name more clear :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+                cache_position=self.cache_position + 1,
+                past_key_values=self.past_key_values,
+                temperature=self.temperature,
+                top_k=self.top_k,
+        )
+        self.cache_position += 1
+        self.generated_ids[:, self.cache_position] = next_token.int()
+        return next_token
+
+    def gen_next_token(self, next_token: torch.Tensor) -> torch.Tensor:
```
sharpenb: It looks like we could merge gen_next_token and gen_next_token_raw :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+
+import contextlib
+from typing import Callable
+
+import torch
+from torch.nn.attention import SDPBackend, sdpa_kernel
+from transformers.cache_utils import StaticCache
+
+
+class HFGenerator:
```
sharpenb: It is not clear that we should call it HF Generator. Is the function specialized to CausalLM or transformers? IN general, it feels that this class/code is a good base to adapt but we can probably think how to clean it :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+
+        self.model = model
+        self.device = model.device
+        self.temperature = temperature
+        self.top_k = top_k
+        self.use_cache = True
+        self.compile_mode = compile_mode
+        self.compile_fullgraph = compile_fullgraph
+        self.batch_size = batch_size
+        self.decode_one_token = self.decode_one_token_sampled
```
sharpenb: Could we merge the functions decode_on_token and decode_one_token_sampled? I was searching for the decode_one_token function but was surprised not to find it right away.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+        self.cache_size = max_kv_cache_size
+
+        self.setup_cache()
+
+        self.is_compiled = False
```
sharpenb: I would suggest to replace l103-105 with
```
self.is_compiled=False
self.decode_one_token = torch.compile(decode_one_token, mode=self.compile_mode, fullgraph=self.compile_fullgraph)
self.is_compiled=True
```
It would make the overall code much shorter :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+
+    def enable_cuda_graph_(self) -> None:
+        """Enable the CUDA graph."""
+        self.do_capture_graph = True
+        self.gen_next_token = self.gen_next_token_withgraph  # type: ignore
+
+    def enable_cuda_graph(
+        self,
+        iters: int = 2,
+        prompt_tokenized: list[int] = [596, 8830, 315, 6913, 19476, 11, 1778, 439, 279, 12939],
```
sharpenb, line 356: This is the random prompt, right?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+        -------
+        None
+            This method modifies the internal state of the generator to use CUDA graphs
+            but does not return any value.
+        """
+        _ = self.generate(
+            torch.tensor(prompt_tokenized, device=self.model.device).unsqueeze(0),
+            max_new_tokens=max_kv_cache_size
+        )
+        for _ in range(iters):
```
sharpenb, line 381: Why do we need mulitple iterations? It looks like it might be not needed to capture the cuda graph.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+        max_kv_cache_size : int
+            The maximum KV cache size.
+
+        Returns
+        -------
+        None
+            This method modifies the internal state of the generator to use CUDA graphs
+            but does not return any value.
+        """
+        _ = self.generate(
```
sharpenb, line 377: Why do we need this call without enabling the cuda graph before?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+            max_new_tokens=max_kv_cache_size
+        )
+        for _ in range(iters):
+            self.enable_cuda_graph_()
+            _ = self.generate(
+                torch.tensor(prompt_tokenized, device=self.model.device).unsqueeze(0),
+                max_new_tokens=max_kv_cache_size
+            )
+
+    def gen_next_token_withgraph(self, next_token: torch.Tensor) -> torch.Tensor:
```
sharpenb: I think that all these cuda_graph function are done to support max auto tune nograph mode. It could be nice to discuss when is this mode particularly useful (see my comment in torch compile) :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,530 @@
...
+            The current token.
+
+        Returns
+        -------
+        torch.Tensor
+            The next token generated by the model.
+        """
+        return self.gen_next_token_raw(next_token)
+
+    def enable_cuda_graph_(self) -> None:
```
sharpenb: I feel that we could merhe `enable_cuda_graph_` and `enable_cuda_graph`

sharpenb (changes requested):
Thanks for the udpate! It is pretty nice to factorize the compilation code for multiple quantization methods. While HFGenerator class is a nice base, I believ that we can clean the code so that it will be shorter and easier to maintain. Happy to dicsuss :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes :)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: I have added a small table with HIGGS and HQQ number on a llama8B model quantized with 4 bits. HQQ is slightly faster (thanks to the torchaoint4 kernels we use in this case). However:
- HQQ + torch_compile is not working when the number of bits is different from 4. It yields error when tested on 8 bits. This make sense because we have pinned the HQQ backend to torchaoint4. I guess we could fix this limitation by enabling gemlite backend (it was deleted from our previous version of hqq.py). We can discuss this point and work on it on a next PR.
- HQQ will be slower (with torchaoint4 backend) for large batch size as detailed [here](https://github.com/mobiusml/hqq/issues/138)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: I was not able (I tested on qwen7b, llama1B and llama8B) to spot any errors when using "max-autotune". However, some models are supposed to not be able with torch_compile automatic cuda graphs. in this case, we would have to manage them manually. I tested with and without these manual cudagraphs, and the speed results are exactly the same. But as the additional code lines for manual cudagraphs are minimal I would be in favor to keep them ;)

src/pruna/algorithms/compilation/utils.py
llcnt: Indeed! I now use our `safe_memory_cleanup` function from the utils ;)

src/pruna/algorithms/compilation/utils.py
llcnt: I see your point. We indeed re-write the sampling process of the LLM, and this is not depending on `transformers`. However, all the implementation is currently tied to the management of the `StaticCache`, which is a 'HF' feature, and is not working (to best of my knowledge) with other LLM implementations

src/pruna/algorithms/compilation/utils.py
llcnt: Agrre ;) I cleaned all this current vs next token, to make the code more consistent

src/pruna/algorithms/compilation/utils.py
llcnt: Done :)

src/pruna/algorithms/compilation/utils.py
llcnt: Done :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes! I have also deleted the is_compiled :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes I technically experienced no difference, but could be some, see my answer above ;)

src/pruna/algorithms/compilation/utils.py
llcnt: done !

src/pruna/algorithms/compilation/utils.py
llcnt: When dealing with manual cudagraphs, we need to run a static input without graph first (of shape 1), and then run multiple iterations of graph captures, as derailed in [this thread](https://github.com/huggingface/transformers/issues/30351)

src/pruna/algorithms/compilation/utils.py
llcnt: When dealing with manual cudagraphs, we need to run a static input without graph first (of shape 1), and then run multiple iterations of graph captures, as detailed in [this thread](https://github.com/huggingface/transformers/issues/30351)

sharpenb: I was checking the thread and it mostly mentions forward and warmup after compilation. Could you link to the specific part showing that it needs it without graphs? :)
image

src/pruna/algorithms/compilation/utils.py
sharpenb: Makes sense. It looks to me that `StaticCache` is imported from transformers. If this class is tied to some package/model format, I would then suggest to rename it to represent this specification for now e.g. `TransformersGenerator`. This is not blocking the PR but would be more informative and the current implementation constraints.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -0,0 +1,479 @@
...
+        -------
+        None
+            This method initializes the model by setting it to evaluation mode and configuring
+            the cache implementation and usage settings.
+        """
+        self.model.eval()
+        self.model.generation_config.cache_implementation = "static"
+        self.model.config.use_cache = True
+
+    def multinomial_sample_one_no_sync(self, probs_sort: torch.Tensor) -> torch.Tensor:
```
sharpenb, line 157: It is good for this PR to have our own code. However, it still feels that we could reimport some of their functions and compile it e.g. the list of smapling methods defined [here](https://github.com/huggingface/transformers/blob/4602059aaee7b075ca51c21941ab2d27980bef7c/src/transformers/generation/utils.py#L322). Again, that is good for now but if it is possible that would be much ncier to import all samplers and so that they can all be compiled :)

src/pruna/algorithms/compilation/torch_compile.py
sharpenb: Thanks a lot for it, that is cool :)

src/pruna/algorithms/compilation/utils.py
sharpenb: It is also look that we use the same random prompt for all the passes before/after the compilation. Out of curiosity, what happens if we do only one forward pass on this single prompt? 
image

sharpenb (approved):
Thanks for the udpate! I think that we converged to sthg that we can merge :) As discussed, there would be a list of updates that can be done to optimize and extend the generality of the code :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes, this is my bad, here is the full explanation:
- if you run directly the model.generate with the graph capture, you will get the following error `InternalTorchDynamoError: RuntimeError: Cannot call CUDAGeneratorImpl::current_seed during CUDA graph capture.`, you can reproduce this error by running this [notebook](https://drive.google.com/file/d/1HkcOS3O-RHJJrefDv_LXkHlsmDgHumEt/view?usp=sharing) and just comment out the line `_ = self.generate` without cuda graphs (I left the error in the cell output so that you can see ;) )
- my explanation (but tbh I am not sure, I dont know the subtilities about cuda graphs) is that we have to run a 'normal' generate before activating manually the cuda graphs capture, maybe to fill the KV cache with some values ? I guess this is what they mean by 'Compile with inputs of size [batch_size, 1]' before doing the warmup? What I still dont understand is that this step is not compiled
- what is implemented [here](https://github.com/mobiusml/hqq/blob/1c47dfc2e58bac717087cd215cf45d1d1afd3eea/hqq/utils/generation_hf.py#L354), also do the call without enabling the cuda graph before. Maybe because they experienced the same error while directly capturing the cuda graph?

src/pruna/algorithms/compilation/utils.py
llcnt: Yes maybe some functions (like this multinomial_sample, or the logits compute) can be re-used across other sampling schemes. However, we will still need to define and compile a given function for each sampling strategy, because currently in `transformers`, the generate function is made of a long stack of `if` statements (one for each sampling strategy), and each sampling has its own logic (it seems that only the model.forward is re-used across sampling strategies): 
image

src/pruna/algorithms/compilation/utils.py
llcnt: I changed the class :)

johnrachwan123 (approved):
Thanks a lot for the great work! Let's merge 🚀

johannaSommer (approved):

PR #67, opened 2025-04-11
feat: `max_batch_size` refactoring

johannaSommer:
## Description
This PR refactors the usage of a batch size arguments for various methods. The user now sets the expected inference batch size once in the SmashConfig and this is used throughout the algorithms. The batch_size hyperparameters in the algorithms are deprecated accordingly. Additionally, I deprecated the naming of `max_batch_size` and renamed it to `batch_size`, as "max" might be misleading.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Ran tests for all affected methods and tested the deprecation locally - works as intended.

## Checklist

## Additional Notes
None.

### DIFF
--- a/src/pruna/algorithms/batching/ifw.py
+++ b/src/pruna/algorithms/batching/ifw.py
@@ -114,6 +114,7 @@ class IFWBatcher(PrunaBatcher):
 
         # ignore mypy warnings here because we ensure beforehand that processor is not None
         pruna_logger.info(f"Preparing model for inference with batch size {smash_config.batch_size}")
+        smash_config.lock_batch_size()
         pipe = pipeline(
             "automatic-speech-recognition",
             model=model,
--- a/src/pruna/algorithms/batching/ifw.py
+++ b/src/pruna/algorithms/batching/ifw.py
@@ -22,6 +22,7 @@ from transformers.utils import is_flash_attn_2_available
 from pruna.algorithms.batching import PrunaBatcher
 from pruna.algorithms.compilation.c_translate import WhisperWrapper
 from pruna.config.smash_config import SmashConfigPrefixWrapper
+from pruna.logging.logger import pruna_logger
 
 
 class IFWBatcher(PrunaBatcher):
@@ -31,6 +32,8 @@ class IFWBatcher(PrunaBatcher):
     Insanely Fast Whisper is an optimized version of Whisper models that significantly speeds up transcription.
     It achieves lower latency and higher throughput through low-level code optimizations and efficient batching,
     making real-time speech recognition more practical.
+    Note: IFW prepares the model for inference with the batch size specified in the smash config. Make sure to set the
+    batch size to a value that corresponds to your inference requirements.
     """
 
     algorithm_name = "ifw"
@@ -110,6 +113,7 @@ class IFWBatcher(PrunaBatcher):
         torch_dtype = torch.float16 if smash_config["weight_bits"] == 16 else torch.float32
 
         # ignore mypy warnings here because we ensure beforehand that processor is not None
+        pruna_logger.info(f"Preparing model for inference with batch size {smash_config.batch_size}")
         pipe = pipeline(
             "automatic-speech-recognition",
             model=model,
--- a/src/pruna/algorithms/batching/ifw.py
+++ b/src/pruna/algorithms/batching/ifw.py
@@ -58,12 +58,6 @@ class IFWBatcher(PrunaBatcher):
                 default_value=16,
                 meta=dict(desc="Sets the number of bits to use for weight quantization."),
             ),
-            OrdinalHyperparameter(
-                "batch_size",
-                sequence=[1, 2, 4, 8, 16, 32, 64],
-                default_value=16,
-                meta=dict(desc="The batch size to use for inference. Higher is faster but needs more memory."),
-            ),
             Constant(name="chunk_length", value=30),
         ]
 
@@ -122,7 +116,7 @@ class IFWBatcher(PrunaBatcher):
             tokenizer=smash_config.processor.tokenizer,  # type: ignore[attr-defined]
             feature_extractor=smash_config.processor.feature_extractor,  # type: ignore[attr-defined]
             chunk_length_s=smash_config["chunk_length"],
-            batch_size=smash_config["batch_size"],
+            batch_size=smash_config.batch_size,
             torch_dtype=torch_dtype,
             model_kwargs=(
                 {"attn_implementation": "flash_attention_2"}
--- a/src/pruna/algorithms/quantization/gptq_model.py
+++ b/src/pruna/algorithms/quantization/gptq_model.py
@@ -127,7 +127,7 @@ class GPTQQuantizer(PrunaQuantizer):
             )
 
             model = imported_modules["GPTQModel"].load(temp_dir, gptq_config)
-            model.quantize(calib_data, batch_size=smash_config.max_batch_size)
+            model.quantize(calib_data, batch_size=smash_config.batch_size)
             model.save(temp_dir)
             model = imported_modules["GPTQModel"].load(temp_dir)
 
--- a/src/pruna/config/smash_config.py
+++ b/src/pruna/config/smash_config.py
@@ -592,6 +592,7 @@ class SmashConfig:
                 "ifw_batch_size",
                 "higgs_example_batch_size",
                 "diffusers_higgs_example_batch_size",
+                "torch_compile_batch_size",
             ]
             if name in deprecated_hyperparameters:
                 warn(
--- a/src/pruna/config/smash_config.py
+++ b/src/pruna/config/smash_config.py
@@ -35,7 +35,7 @@ from pruna.data.pruna_datamodule import PrunaDataModule, TokenizerMissingError
 from pruna.logging.logger import pruna_logger
 
 ADDITIONAL_ARGS = [
-    "max_batch_size",
+    "batch_size",
     "device",
     "cache_dir",
     "save_fns",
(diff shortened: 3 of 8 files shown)


src/pruna/config/smash_config.py
```diff
@@ -76,7 +79,15 @@ def __init__(
...
         )
         self.config_space: ConfigurationSpace = self._configuration.config_space
-        self.max_batch_size = max_batch_size
+        if max_batch_size is not None:
+            warn(
+                "max_batch_size is soon to be deprecated. Please use batch_size instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            self.batch_size = max_batch_size
```
gsprochette, line 88: Should we check whether both max_batch_size and batch_size arguments are being used? I think it's fine as is, it's just extra security, feel free to ignore this comment.

src/pruna/config/smash_config.py
```diff
@@ -250,10 +273,7 @@ def train_dataloader(self, **kwargs) -> torch.utils.data.DataLoader | None:
         torch.utils.data.DataLoader | None
```
gsprochette, line 292: Could we take this opportunity to have the same return type in the docstrings of all dataloader methods?

src/pruna/config/smash_config.py
```diff
@@ -231,9 +242,21 @@ def flush_configuration(self) -> None:
...
         self.reset_cache_dir()
 
+    def __get_dataloader(self, dataloader_name: str, **kwargs) -> torch.utils.data.DataLoader | None:
+        if self.data is None:
+            return None
+
+        if "batch_size" in kwargs and kwargs["batch_size"] != self.batch_size:
+            pruna_logger.warning(
+                f"Batch size {kwargs['batch_size']} is not the same as the batch size {self.batch_size}"
+                f"set in the SmashConfig. Using the {self.batch_size} from the SmashConfig."
```
gsprochette: The last sentence could be rephrased since it will become e.g. "Using the 4 from the SmashConfig", maybe "Using batch_size={self.batch_size} from the SmashConfig" would be clearer, or just "Using {self.batch_size}." since the detail was already provided in the previous sentences.

gsprochette (commented):
Looks pretty much good to me, I'm looking forward to using this unified batch_size and having access to data from outside the smash_config :) I only suggested a couple of minor changes which should take just a minute.

src/pruna/config/smash_config.py
johannaSommer: I completely get what you mean, I wanted to do the same thing originally... the problem is that we cant definitely know what is "being used" because the `batch_size` is never "None"... Worst case someone uses both to set the batch size to 1 - we wouldnt really know if they are "using" `batch_size` or if its just the default value of the argument

src/pruna/config/smash_config.py
johannaSommer: we definitely can 🤓 done!

src/pruna/config/smash_config.py
johannaSommer: absolutely, thanks!

src/pruna/algorithms/quantization/gptq_model.py
```diff
@@ -127,7 +127,7 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
             )
 
             model = imported_modules["GPTQModel"].load(temp_dir, gptq_config)
-            model.quantize(calib_data, batch_size=smash_config.max_batch_size)
+            model.quantize(calib_data, batch_size=smash_config.batch_size)
```
begumcig, line 130: Does the batch_size argument here override the calib_data's batch_size?

src/pruna/algorithms/batching/ifw.py
```diff
@@ -116,13 +113,14 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
         # ignore mypy warnings here because we ensure beforehand that processor is not None
+        pruna_logger.info(f"Preparing model for inference with batch size {smash_config.batch_size}")
         pipe = pipeline(
             "automatic-speech-recognition",
             model=model,
             tokenizer=smash_config.processor.tokenizer,  # type: ignore[attr-defined]
             feature_extractor=smash_config.processor.feature_extractor,  # type: ignore[attr-defined]
             chunk_length_s=smash_config["chunk_length"],
-            batch_size=smash_config["batch_size"],
+            batch_size=smash_config.batch_size,
```
begumcig, line 124: Okay, just wanted to flag two things here: 

1 Since this batch_size can directly impact inference performance (latency, memory, etc.), I'm a bit concerned that changes here (or in any future algorithm that plays with this setting) could unintentionally affect our evaluation metrics, especially when comparing against the base model. Would it make sense to pass a config to the evaluation agent as well, so we're running everything under the same conditions?

2 This is not a problem but more of a question but what happens if the data we pass later (e.g., from a DataLoader) is already batched differently? Does this lead to re-batching under the hood? Might be worth double-checking that the batch sizes align, or that we’re not unintentionally introducing extra batching logic from the pipeline itself.

begumcig (changes requested):
Everything already looks super solid! Just left a small comment regarding batch_size mismatches btw the pipeline & dataloader and how it could affect evaluation. 
Great job overall 🥹

src/pruna/config/smash_config.py
gsprochette: I guess we could set the default of `batch_size` to None, but the reason for doing this seems unimportant so let's leave it as is I would say.

src/pruna/algorithms/quantization/gptq_model.py
johannaSommer: yeah good question, in this case the calib data is the string of all text snippets from the dataset as a whole and doesnt have an inherent batch size... and then GPTQ slices and embedds as necessary

src/pruna/algorithms/batching/ifw.py
johannaSommer: after our discussion, I added the following warning:

```python
from pruna.evaluation.evaluation_agent import EvaluationAgent
from pruna.evaluation.task import Task
from pruna.data.pruna_datamodule import PrunaDataModule
import torch

pipe = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", torch_dtype=torch.float16)

config = SmashConfig() # batch size is 1 by default
config["cacher"] = "deepcache"
smashed_pipe = smash(pipe, config)

task = Task(["gpu_memory"], datamodule=PrunaDataModule.from_string("LAION256", dataloader_args={"batch_size": 3}))
agent = EvaluationAgent(task)
agent.evaluate(smashed_pipe)
```

would now output:
```
INFO - Starting cacher deepcache...
INFO - cacher deepcache was applied successfully.
INFO - Loaded only training, splitting train 80/10/10 into train, validation and test...
INFO - Testing compatibility with functools.partial(<function image_generation_collate at 0x7f6950b86f80>, img_size=512)...
INFO - Creating metrics from names: ['gpu_memory']
INFO - Evaluating a smashed model.
INFO - Detected diffusers model. Using DiffuserHandler with fixed seed.
- The first element of the batch is passed as input.
- The generated outputs are expected to have .images attribute.
WARNING - Batch size mismatch between evaluation datamodule and smashed model's smash config. This may lead to incorrect metric computation due to compression algorithms being batch size specific. Adjust the datamodule creation to match the smashed model's batch size, e.g., datamodule = PrunaDataModule.from_string(dataset_name, dataloader_args={'batch_size': 1})
INFO - Evaluating stateful metrics.
INFO - Evaluating isolated inference metrics.
```

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -124,6 +124,18 @@ def prepare_model(self, model: Any) -> PrunaModel:
             pruna_logger.info("Evaluating a base model.")
 
         model.inference_handler.log_model_info()
+        if (
+            "batch_size" in self.task.datamodule.dataloader_args
+            and self.task.datamodule.dataloader_args["batch_size"] != model.smash_config.batch_size
```
begumcig, line 129: This is a step in the right direction! My only concern is that since the smash_config always includes a batch_size attribute by default, we might end up showing this warning every time—even when the model itself isn't changing in a way that would affect inference. Ideally, the warning should only appear if the model will actually run inference with a different batch size internally.

I'm not entirely sure how to reliably detect that. But if this issue only occurs with specific batching algorithms (is this the case???), maybe we could check whether the smashing algorithm is "batcher" instead of just relying on the batch_size in the config, for instance. What do you think?

begumcig (changes requested):

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -124,6 +124,19 @@ def prepare_model(self, model: Any) -> PrunaModel:
...
+            and not is_base
+            and model.smash_config.is_batch_size_locked()
```
begumcig, line 131: 💅💅💅

begumcig (approved):

src/pruna/config/smash_config.py
```diff
@@ -76,7 +79,15 @@ def __init__(
             SMASH_SPACE.get_default_configuration() if configuration is None else configuration
         )
         self.config_space: ConfigurationSpace = self._configuration.config_space
-        self.max_batch_size = max_batch_size
+        if max_batch_size is not None:
+            warn(
+                "max_batch_size is soon to be deprecated. Please use batch_size instead.",
```
gsprochette: Why "soon to be" and not "is"?

johannaSommer: yeah youre completely right 🙈 thanks!

PR #69, opened 2025-04-16
feat: support portable torch compilation

johannaSommer:
## Description
This PR prepares for the release of torch 2.7 which supports portable compilation. It includes extensions to the loading and saving mechanisms for torch compile and adds a tutorial to guide users through this new feature.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Added a tutorial notebook which is included in the tests.

## Checklist

## Additional Notes
For any guidance on how to use this feature, please refer to the tutorial.

### DIFF
--- a/src/pruna/algorithms/compilation/torch_compile.py
+++ b/src/pruna/algorithms/compilation/torch_compile.py
@@ -148,6 +148,11 @@ class TorchCompileCompiler(PrunaCompiler):
             The model to compile.
         smash_config : SmashConfig
             The configuration for the compilation.
+
+        Returns
+        -------
+        Any
+            The compiled model.
         """
         if smash_config["torch_compile_make_portable"]:
             os.environ["TORCHINDUCTOR_FX_GRAPH_CACHE"] = "1"
--- a/src/pruna/algorithms/compilation/torch_compile.py
+++ b/src/pruna/algorithms/compilation/torch_compile.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
 from typing import Any, Callable, Dict
 
 import torch
@@ -19,7 +20,7 @@ from ConfigSpace import CategoricalHyperparameter, OrdinalHyperparameter
 
 from pruna.algorithms.compilation import PrunaCompiler
 from pruna.algorithms.compilation.utils import TransformersGenerator
-from pruna.config.smash_config import SmashConfigPrefixWrapper
+from pruna.config.smash_config import SmashConfig, SmashConfigPrefixWrapper
 from pruna.config.smash_space import Boolean
 from pruna.engine.model_checks import (
     get_diffusers_transformer_models,
@@ -27,6 +28,7 @@ from pruna.engine.model_checks import (
     is_causal_lm,
     is_opt_model,
 )
+from pruna.engine.save import SAVE_FUNCTIONS
 from pruna.logging.logger import pruna_logger
 
 # This allows for torch compile to use more cache memory to compile the model
@@ -107,6 +109,15 @@ class TorchCompileCompiler(PrunaCompiler):
                     "CUDA graph capture overhead."
                 ),
             ),
+            Boolean(
+                "make_portable",
+                meta=dict(
+                    desc=(
+                        "Whether to make the model compiled model portable or not, "
+                        "and significantly reduce the warmup time of the model on a different machine."
+                    ),
+                ),
+            ),
         ]
 
     def model_check_fn(self, model: Any) -> bool:
@@ -126,6 +137,27 @@ class TorchCompileCompiler(PrunaCompiler):
         # opt models have no cache_position, so will raise error like
         # TypeError: OPTForCausalLM.forward() got an unexpected keyword argument 'cache_position'
         return callable(model) and not is_opt_model(model)
+
+    def apply(self, model: Any, smash_config: SmashConfig) -> Any:
+        """
+        Apply the compilation algorithm to the model.
+
+        Parameters
+        ----------
+        model : Any
+            The model to compile.
+        smash_config : SmashConfig
+            The configuration for the compilation.
+        """
+        if smash_config["torch_compile_make_portable"]:
+            os.environ["TORCHINDUCTOR_FX_GRAPH_CACHE"] = "1"
+
+        output = super().apply(model, smash_config)
+
+        # importantly, the torch artifacts saving need to be done *after* the before-compile-save
+        if smash_config["torch_compile_make_portable"]:
+            smash_config.save_fns.append(SAVE_FUNCTIONS.torch_artifacts.name)
+        return output
 
     def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         """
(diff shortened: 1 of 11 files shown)


src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -82,6 +84,15 @@ def get_hyperparameters(self) -> list:
                 default_value=None,
                 meta=dict(desc="Whether to use dynamic shape tracing or not."),
             ),
+            Boolean(
```
sharpenb, line 112: Why would people not make it portable? :)

src/pruna/engine/load.py
```diff
@@ -58,10 +58,18 @@ def load_pruna_model(model_path: str, **kwargs) -> tuple[Any, SmashConfig]:
...
         raise ValueError("Load function has not been set.")
 
-    model = LOAD_FUNCTIONS[smash_config.load_fn](model_path, **kwargs)
+    if LOAD_FUNCTIONS.torch_artifacts.name in smash_config.load_fns:
+        load_torch_artifacts(model_path, **kwargs)
+        smash_config.load_fns.remove(LOAD_FUNCTIONS.torch_artifacts.name)
+
+    if len(smash_config.load_fns) > 1:
+        pruna_logger.error(f"Load functions not used: {smash_config.load_fns[1:]}")
+        smash_config.load_fns = smash_config.load_fns[:1]
```
sharpenb: Do we need this line? It looks like in any case we will use the index 0 at l.72.

src/pruna/engine/load.py
```diff
@@ -153,8 +161,12 @@ def load_diffusers_model(path: str, **kwargs) -> Any:
     Any
         The loaded diffusers model.
     """
-    # if it is a diffusers model, it saves the model_index.json file
-    model_index = load_json_config(path, "model_index.json")
+    if os.path.exists(os.path.join(path, "model_index.json")):
+        # if it is a diffusers pipeline, it saves the model_index.json file
+        model_index = load_json_config(path, "model_index.json")
+    else:
```
sharpenb, line 274: Why did not we get this problem before? It is unclear how it is related to the save compiled model.

src/pruna/engine/load.py
```diff
@@ -260,6 +272,29 @@ def load_quantized(model_path: str, **kwargs) -> Any:
...
+    **kwargs : Any
+        Additional keyword arguments to pass to the model loading function.
+    """
+    with open(os.path.join(model_path, "artifact_bytes.bin"), "rb") as f:
+        artifact_bytes = f.read()
+
+    # check if the bytes are empty
+    if artifact_bytes == b"\x00\x00\x00\x00\x00\x00\x00\x01":
+        pruna_logger.error(
+            "Model has not been run before. Please run the model before saving to construct the compilation graph."
```
sharpenb: I am not sure if this is the time where we should alert the user about that. When he/she reaches that stage, the problem cannot be solved on the saved model. Such warning should probably be done when saving the model. Another point is whether we can simply force the run before sving the model.

src/pruna/engine/save.py
```diff
@@ -49,6 +48,11 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) ->
     """
     if not os.path.exists(model_path):
         os.makedirs(model_path)
+
+    if SAVE_FUNCTIONS.torch_artifacts.name in smash_config.save_fns:
```
sharpenb, line 56: One comment to explain why we do this would be nice :)

src/pruna/engine/save.py
```diff
@@ -153,7 +157,8 @@ def save_before_apply(model: Any, model_path: str, smash_config: SmashConfig) ->
     # load json directly from file
     with open(os.path.join(save_dir, SMASH_CONFIG_FILE_NAME), "r") as f:
         old_smash_config = json.load(f)
-    smash_config.load_fn = deepcopy(old_smash_config["load_fn"])
+    smash_config.load_fns.extend(old_smash_config["load_fns"])
```
sharpenb, line 250: It is not clear why this change. Smash config is not returned or used later in the function also. I might miss sthg.

src/pruna/engine/save.py
```diff
@@ -261,7 +266,31 @@ def save_quantized(model: Any, model_path: str, smash_config: SmashConfig) -> No
...
+        The directory to save the model to.
+    smash_config : SmashConfig
+        The SmashConfig object containing the save and load functions.
+    """
+    artifacts = torch.compiler.save_cache_artifacts()
+
+    assert artifacts is not None
+    artifact_bytes, cache_info = artifacts
+
+    with open(os.path.join(model_path, "artifact_bytes.bin"), "wb") as f:
```
sharpenb, line 386: See my comment in the respective load function :)

sharpenb (changes requested):
Very cool feature! It looks like the PR description is empty. Left some comments as well :)

src/pruna/engine/load.py
johannaSommer: Yes good point sorry, this is not related to saving the compiled model directly. When testing this feature I did one of the common scenarios which is to only compile the unet of a pipeline, i.e.g `smash(pipe.unet, smash_config)`. I discovered that this case isnt covered, luckily, it only needed a small extension. Full example for context:
```python
pipe = StableDiffusionPipeline()
config = SmashConfig()
config["compiler"] = "torch_compile"
smashed_unet = smash(pipe.unet, config)
smashed_unet.save_pretrained("path/")
# Loading this would have not worked previously
````

src/pruna/algorithms/compilation/torch_compile.py
johannaSommer: Hm i would say because it requires additinooal overhead (both the addiitonal running of the model before saving as well as the need to save more files) would likely break current workflows of users using torch compile

src/pruna/engine/load.py
johannaSommer: completely agree, I think this was originally my intention not sure why it ended up here

src/pruna/engine/save.py
johannaSommer: see explanation on adjust load_fns in smash config

src/pruna/engine/save.py
johannaSommer: If you remember, in the saving logic of pruna we have a queue of saving functions that take care of different aspects of the saving process that we go through step by step. Previously this was not the case with the loading, it would be just one function. To keep it more general and enable this torch compilation saving, we also make the load functions a queue that takes care of different things, e.g. loading the model *and* the cache in seperate functions that can be used independently

johannaSommer: also added a comment anyway!

johannaSommer:
Thanks a lot @sharpenb for the review! I was able to address all of your comments :) This will remain a draft PR until the new torch version is "officially" available and in its final stage, but it would be great to already have the changes approved and basically ready to go 👍🏻

johnrachwan123 (approved):
This is amazing! Let's discuss these things before merging to create a game plan though:

1. How can we make it even simpler for the user
2. Can we make lora switching for free?
3. Can we reduce the time to import pruna from scratch ? (I noticed that just importing pruna takes 5-7s maybe we can identify what is taking so long or follow a more lazy import stratefy like diffusers?

But overall amazing job 🚀

src/pruna/engine/save.py
sharpenb: Thanks! I might miss the comment above line 56 though. I feel that it should mention that this is to handle compiled saving :)

PR #70, opened 2025-04-20
ci: cache poetry installation

johannaSommer:
## Description
This PR introduces cached virtual environments to the internal CPU tests and the linting workflow. In short, the workflow creates a poetry "lock" file which specifies the *exact* versions of all installed packages. This lock file is then hashed and the cache of virtual environments is checked for a hit. In case of a hit, the cached environment is loaded and the workflow continues. In case of a miss, it falls back to an old environment and does updates where necessary. Finally, in case of a miss, the new environment is uploaded under a new hash.
In short, this reduces the "setup time" of a workflow from 2.5 minutes to 1 minute, of which 30 seconds are used to create the lock file and 30 seconds are used to retrieve the cached files. In my understanding, using uv would speed up specifically the process of creating a lock file significantly.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Tested with the workflows on GitHub.

## Checklist

## Additional Notes
@johnrachwan123 let's discuss the changes to stable-fast and see what we can do here.
Also while I was anw adjusting the workflow files, I adjusted the permissions as recommended by GitHub in the security scans.

### DIFF
--- a/.github/workflows/external_contributions_tests.yaml
+++ b/.github/workflows/external_contributions_tests.yaml
@@ -46,9 +46,4 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -m "cpu and not slow and not high_cpu" --cov=pruna --cov-report=xml
-
-      - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v3
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+          pytest -m "cpu and not slow and not high_cpu"
--- a/.github/workflows/external_contributions_tests.yaml
+++ b/.github/workflows/external_contributions_tests.yaml
@@ -1,5 +1,9 @@
 name: Fast CPU Tests for external contributions
 
+permissions:
+  contents: read
+  pull-requests: read
+
 on:
   pull_request_target:
     types: [ review_requested ]
@@ -22,7 +26,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/internal_contributions_tests.yaml
+++ b/.github/workflows/internal_contributions_tests.yaml
@@ -1,5 +1,9 @@
 name: Fast CPU Tests
 
+permissions:
+  contents: read
+  pull-requests: read
+
 on:
   push:
     branches:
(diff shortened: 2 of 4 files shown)


.github/workflows/external_contributions_tests.yaml
```diff
@@ -42,9 +46,4 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -m "cpu and not slow and not high_cpu" --cov=pruna --cov-report=xml
-
-      - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v3
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+          pytest -m "cpu and not slow and not high_cpu"
```
sharpenb, line 49: It is probably fine. Could you explainthe lgoic of removing codecov here? :)

.github/workflows/internal_contributions_tests.yaml
```diff
@@ -25,29 +29,29 @@ jobs:
...
+          restore-keys: |
+            poetry-${{ runner.os }}-${{ matrix.python-version }}
 
-      - name: Install dependencies
-        run: pip install -e .[tests] --extra-index-url https://prunaai.pythonanywhere.com/
+      - name: Install dependencies using Poetry
+        run: poetry install --extras tests
 
       - name: Run tests
         run: |
```
sharpenb, line 54: Same question fo codecov here :)

.github/workflows/linting.yaml
```diff
@@ -13,35 +17,42 @@ jobs:
...
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Poetry
-        run: |
-          curl -sSL https://install.python-poetry.org | python3 -
-          echo "${HOME}/.local/bin" >> $GITHUB_PATH
+        uses: abatilo/actions-poetry@v2
```
sharpenb, line 30: Could you explain why do we need it for the linting workflow as well? It is not clear why a poetry env is needed in that case.

johannaSommer: we need it here as well because poetry caches the environment and we need the environment to execute the pytests for style, etc.

.github/workflows/internal_contributions_tests.yaml
johannaSommer: we now record codecov via the nightly tests, which gives us the whole coverage and not just that of the CPU test subset

PR #75, opened 2025-04-25
feat: qkv fusing

llcnt:
## Description

This PR introduces a new family of algorithm: ~~FUSER~~ FACTORIZER !
The general idea behind matrix fusing is to concatenate $n$ matrices together (eg $W_{large}=[W_{small}, \dots, W_{small}]$) and perform a single $W_{large} \cdot x$ product instead of $n$ smaller $W _{small} \cdot x$ products.
There is no memory gains. And, in general, there is no latency gains.
However, when combined with quantization, matrix fusing can enable speedups (see https://github.com/huggingface/diffusers/pull/9185)

The main changes in the code are:
- Add a new ~~`FUSER`~~ `FACTORIZER` in the new ~~`fusing`~~ `factorizing` folder;
- Add a new ~~fuser~~ factorizer for qkv matrices for `diffusers` models;
- Add a new check function that look at the models in the `diffusers` package that are compatible with fusing (ie have some FusedAttention processors);
- Add qkv_fusing for all compatible algorithms;
- Delete the original model (fixture) loaded during the unit test to avoid OOM errors.

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?

I have added a unit test for the new qkv fuser on sdv1.4.

## Checklist


src/pruna/engine/utils.py
```diff
@@ -200,3 +201,70 @@ def set_to_train(model: Any) -> None:
...
+        ----------
+        exc_type : Exception
+            The exception type.
+        exc_value : Exception
+            The exception value.
+        traceback : Exception
+            The traceback.
+        """
+        if hasattr(self.pipeline, "transformer"):
+            self.pipeline.transformer = self.pipeline.working_model
```
nifleisch, line 359: I think this should be 
`self.pipeline.transformer = self.working_model`

nifleisch (approved):
LGTM! 🙌

src/pruna/algorithms/factorizing/qkv_factorizing.py
```diff
@@ -0,0 +1,111 @@
...
+        ----------
+        model : Any
+            The model to check.
+
+        Returns
+        -------
+        bool
+            True if the model has a fused attention processor, False otherwise.
+        """
+        return has_fused_attention_processor(model)
```
sharpenb: I was wondering if this is only for diffusers in view of the aaply funciton that have pipelines mentions. We should either make it work for all models or update the checks :)

sharpenb: If this is specific to diffusers, I would recommend to call the algo `qkv_diffusers` similarly to some other quant methods :)

src/pruna/engine/model_checks.py
```diff
@@ -426,6 +426,50 @@ def get_diffusers_unet_models() -> list:
...
+
+    Returns
+    -------
+    bool
+        True if the model can use a Fused attention processor, False otherwise.
+    """
+    if not hasattr(model, "attn_processors"):
+        return False
+    fusing_possible_list = [
+        processor_name.replace("Fused", "") for processor_name in dir(diffusers.models.attention_processor)
```
sharpenb, line 531: Why is the list of fusing possible elements composed of the processor names that had Fused in them? It sounds not intuitive that they were already denoted as fused.

src/pruna/engine/model_checks.py
```diff
@@ -426,6 +426,50 @@ def get_diffusers_unet_models() -> list:
...
+    ----------
+    pipeline : Any
+        The diffusers pipeline to check.
+
+    Returns
+    -------
+    bool
+        True if the pipeline can use a Fused attention processor, False otherwise.
+    """
+    if not is_unet_pipeline(pipeline) and not is_transformer_pipeline(pipeline):
```
sharpenb: It sounds that this functions tests more thatn simply has fused attention. It tests if these are unet/transformer pipelines as well. I think though that we might have already some code for that somewhere. I would then recommend that model checks check fist the unet/trasnformer part and then the fused attention. The fused attention should not implement this code again.

tests/common.py
```diff
@@ -70,6 +71,12 @@ def run_full_integration(algorithm_tester: Any, device: str, model_fixture: tupl
         model = algorithm_tester.cast_to_device(model, device=smash_config["device"])
         smashed_model = algorithm_tester.execute_smash(model, smash_config)
         algorithm_tester.execute_save(smashed_model)
+        if hasattr(model, "to"):
```
sharpenb: Why is cpu/desrtroy/cleanup part of this PR? :)

sharpenb (changes requested):
Thanks for the PR! I left some comments :)

src/pruna/engine/model_checks.py
llcnt: By default, the attn_processors attribute of a pipeline.transformer is not made of the Fused ones, but of the default as you can see in the following: 
image

So we need to look in `diffusers` for all Fused_attn_processors, and create a list of the 'parent' processors they depend on (eg [FusedFluxAttnProcessor2_0](https://github.com/huggingface/diffusers/blob/06fee551e9c33b2504a379982d2d3a76501ad8a9/src/diffusers/models/attention_processor.py#L2473) is the extension of FluxAttnProcessor2_0), to check if the model will be able to swap from attn_processor to Fusedattn_processor ;)

src/pruna/algorithms/factorizing/qkv_factorizing.py
llcnt: I see your point and I agree :)
Indeed currently, qkv fusing is not available in `transformers` (except for awq quantized models).
I was hesitating in between the naming conventions: `qkv_factorizer` (lack the diffuers mention), `qkv_factorizer_diffusers` (informative but maybe too long). I am also ok with `qkv_diffusers` (but qkv by itself is a bit misleading as it does not refer to any standard family of algo)
What do you think?

src/pruna/engine/model_checks.py
llcnt: I have deleted this line, and I use in addition the `is_diffusers_model()` function in the model_check_fn :)

tests/common.py
llcnt: We could do a separate PR for this, but I think it is a small enough change in code to do it here ;)
I added a new test on Flux in this [PR](https://github.com/PrunaAI/prunatree/pull/83), and it leads to OOM on a L40S GPU if I don't change the above lines: in the `run_full_integration` function, the initial model was moved onto device with `model = algorithm_tester.cast_to_device`, but never un-loaded

src/pruna/engine/utils.py
nifleisch: Maybe I am missing something especially since I am not super familiar with how these context classes work but in the `__enter__` method working model is set as an attribute of the ModelContext and not as an attribute of the pipeline. So I think `__exit__` should be 
```suggestion
        if hasattr(self.pipeline, "transformer"):
            self.pipeline.transformer = self.working_model
            elif hasattr(self.pipeline, "unet"):
            self.pipeline.unet = self.working_model
        else:
            self.pipeline = self.working_model
        del self.working_model
```
Alternatively you can set `working_model` as an attribute of `self.pipeline` in the `__enter__` method.

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -46,7 +46,11 @@ class DiffusersInt8Quantizer(PrunaQuantizer):
     dataset_required = False
     run_on_cpu = False
     run_on_cuda = True
-    compatible_algorithms = dict(cacher=["deepcache"], compiler=["torch_compile"])
+    compatible_algorithms = dict(
+        factorizer=["qkv_factorizing"],
```
nifleisch: This is probably highly subjective, but I would also prefer something like `qkv_diffusers` or even `qkv_fusing_diffusers` instead of `qkv_factorizing`.

nifleisch (changes requested):
Still awesome PR! I only have one small remark about the `ModelContext`.

src/pruna/engine/utils.py
llcnt: Yeah I see what you mean ;) tbh I am not familiar with context manager, and during my test I found that modifying the working_model breaks the attribute link in the context manager. I tried to explain this in [this comment](https://github.com/PrunaAI/pruna/blob/9e5f43bf5fa992fd8f345c321f9b0d7d98395658/src/pruna/algorithms/factorizing/qkv_factorizing.py#L98) :(
Ideally, I agree with you I would like to just modify the working_model, and do self.pipeline.transformer = self.working_model in the `__exit__`.
Instead I have to define the working_model (once it is smashed) as an attribute of the pipeline like [this](https://github.com/PrunaAI/pruna/blob/9e5f43bf5fa992fd8f345c321f9b0d7d98395658/src/pruna/algorithms/factorizing/qkv_factorizing.py#L100), and then do self.pipeline.transformer = self.pipeline.working_model  in the `__exit__`

nifleisch: I completely missed your request for my review—sorry about that. Overall, this looks fine to me. However, like you, I’m not entirely happy with the current implementation, since it requires you to set pipeline.working_model inside the context block, and will fail otherwise. I’m not sure that’s always desirable. Moreover, if the working model is itself an nn.Module, attaching one nn.Module as an attribute of another could introduce side effects.

I also wonder whether you really need the __exit__ method at all, since qkv-fusing is an in-place operation of the transformer and may not require any overwrite. But if you’d prefer to keep the version as-is, that works for me too.

nifleisch (approved):

src/pruna/engine/model_checks.py
sharpenb: Thanks that is very clear :)

src/pruna/engine/model_checks.py
sharpenb: Perfect, thanks :)

tests/common.py
sharpenb: Makes sense. I am still wondering if these things should not be done by default in the safe_memory_clean_up since they all look to do some clean up :)

sharpenb: I think that these should be moved to safe_memory_clean_up :)

sharpenb (changes requested):
Looks very good. It can be merged after renaming and fixing the safe_memory_cleanup :)

tests/common.py
llcnt: I think I see what you have in mind :) Unfortunately `safe_memory_clean_up` is used in many places in `pruna` where no model can be cleaned (see eg [here](https://github.com/PrunaAI/pruna/blob/bc1ece9b77f4fd426fbaf43e03b2f5eb66f2dc96/tests/algorithms/testers/base_tester.py#L64) or [here](https://github.com/PrunaAI/pruna/blob/bc1ece9b77f4fd426fbaf43e03b2f5eb66f2dc96/src/pruna/evaluation/metrics/metric_memory.py#L142) ), so we can not put the model cleaning inside this function.
Maybe I can create another function called `safe_model_clean_up(model)` that moves the model to cpu, then try to destroy it if available and then run safe_memory_clean_up, what do you think ? (tbh this will increase the complexity of this PR because `safe_memory_cleanup` is referenced 39 times in pruna oss/pro in 19 different files)

davidberenstein1957:
@sharpenb @llcnt wrote a fix here. It seemed to be something broader that also impacted general loading.

https://github.com/PrunaAI/pruna/pull/119

tests/common.py
sharpenb: I think that safe_moemory clena_up coudl be added list as arguments of objects that should be deleted with del `safe_memory_cleanup(, objects_to_delete=[])`. In this way the other calls would be unchanged and it would make the clean up function complete.

sharpenb (commented):
I only had a last comment :)

tests/common.py
llcnt: Oh yes ! thats smart :) I have modified accordingly

llcnt: I re-ran the unit test with this latest modification and it works : 
image

sharpenb (approved):
Super cool, let's merge it :)

PR #78, opened 2025-04-28
fix: `GPUMemory` metric memory bug

johannaSommer:
## Description
This PR fixes a bug in the GPUMemory metric where two models are kept on the GPU at the same time. While this does not affect the correctness of the memory measurement, it can lead to OOM issues with models like e.g. Flux despite the model fitting on the GPU (as we are keeping it twice).

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Executed the metric on a previously failing Flux model.

## Checklist

## Additional Notes
None.

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -103,7 +103,7 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
             # there is anyway no way to recover from this error
             # raise it here for better traceability
             raise e
-        except (ValueError, RecursionError, RuntimeError) as e:
+        except (ValueError, RecursionError, RuntimeError, AttributeError) as e:
             if raise_error:
                 raise ValueError(f"Could not move model to device: {str(e)}")
             else:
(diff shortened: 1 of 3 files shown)


src/pruna/evaluation/metrics/metric_memory.py
```diff
@@ -135,6 +135,7 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
         save_path = model.smash_config.cache_dir + "/metrics_save"
         model_cls = model.__class__
         model.save_pretrained(save_path)
+        model.move_to_device("cpu")
```
johnrachwan123, line 138: I think I understand why we do this but I have to admit that cpu memory issues could also arise from this. I don't have a better solution in mind but just a thought to keep in mind.

johnrachwan123 (approved):
I think we can merge but left a comment to keep something in mind

src/pruna/engine/utils.py
```diff
@@ -99,6 +99,10 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
     if hasattr(model, "to"):
         try:
             model.to(device)
+        except torch.cuda.OutOfMemoryError as e:
```
begumcig, line 102: slay

src/pruna/evaluation/metrics/metric_memory.py
begumcig: do you think destroying it would make more sense?

begumcig (approved):

PR #81, opened 2025-04-28
refactor: metric attributes

begumcig:
## Description

General evaluation refactoring for better user experience: 
- Atomic metrics: Each metric returns a single float value as a result. 
- MetricResult: results are returned with MetricResult objects, which carry the metric name, metric metadata and the result value
- Added various class attributes to metrics for clarity like `higher_is_better`, `metric_units`,
- Refactored call types for metrics, now any metric can be called with just `single` or `pairwise`, and cannot be called with non supported call types.
- Metric names in the metric class and metric results match
- Adapted the EvaluationAgent to this paradigm so that even when the metrics are atomic, inference computations are shared when possible

## Type of Change

## Checklist


docs/user_manual/evaluation.rst
```diff
@@ -135,24 +134,27 @@ Metrics help quantify different aspects of model performance, from output qualit
...
+  | ``timing_type``: Type of timing to use ("sync" or "async", default "sync").
 
-Base Metrics
-^^^^^^^^^^^^^
-Simple metrics that compute values directly from inputs without maintaining state across batches. Examples include:
-- Model Architecture metrics
-- Energy consumption metrics
-- Memory usage metrics
 
-elapsed_time
```
johannaSommer: can you elaborate a bit to what extent users can still use the previous metrics / naming? I am a bit worried that we are changing too much and will break existing setups without prior deprecation warnings

src/pruna/evaluation/metrics/metric_cmmd.py
```diff
@@ -102,7 +103,7 @@ def update(self, x: List[Any] | torch.Tensor, gt: torch.Tensor, outputs: torch.T
         self.ground_truth_embeddings.append(gt_embeddings)
         self.output_embeddings.append(output_embeddings)
 
-    def compute(self) -> torch.Tensor:
+    def compute(self) -> MetricResult:
```
johannaSommer, line 107: i am a bit confused about this return type, dont we assume in the evaluation agent that we get the raw results from the `.compute()` function?

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -123,10 +128,166 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
...
+            The model to evaluate.
+        dataloader : DataLoader
+            The dataloader to evaluate the model on.
+
+        Returns
+        -------
+        MetricResult
+            The latency for model inference.
+        """
+        raw_results = super().compute(model, dataloader)
```
johannaSommer, line 214: I definitely see what you are trying to do here with the structure, but wouldnt this lead to double computations if I am requesting both throughput and latency for example?

src/pruna/evaluation/metrics/metric_energy.py
```diff
@@ -133,3 +135,110 @@ def _get_data(self, source: Any, attribute: str, name: str) -> float:
...
+            The model to evaluate.
+        dataloader : DataLoader
+            The dataloader to evaluate the model on.
+
+        Returns
+        -------
+        MetricResult
+            The CO2 emissions of the model.
+        """
+        raw_results = super().compute(model, dataloader)
```
johannaSommer, line 226: same comment as above on double computatinos

docs/user_manual/evaluation.rst
johannaSommer: okay after looking further, i definitely see that you already took this into account. It's a lot which makes it difficult to understand, can you elaborate / summarize briefly again for me here what will still be possible / what can we simply not support anymore?

src/pruna/evaluation/metrics/result.py
```diff
@@ -0,0 +1,66 @@
...
+    result : float | int
+        The result of the metric.
+    """
+
+    name: str
+    params: Dict[str, Any]
+    result: float | int
+
+    @classmethod
+    def from_results_dict(
```
johannaSommer, line 51: out of curiosity, do we expect more functions to be added here?

johannaSommer: I am overall wondering if this metric result might be very similar to the individual metric types that you introduced. What is the conceptual difference between a metric result and a metric? I am asking because i think attributes like higher-is-better and benchmark-metric could also be used here. 
To be clear, I am not disagreeing with the structure, I am just wondering if this needs a seperate data class?

johannaSommer (changes requested):
That's really amazing work you did Begüm! Let's iterate one more time, I have a few questions, but I don't think it will take long then we finally have this refactooooor 🚀

docs/user_manual/evaluation.rst
begumcig: Thanks for the great question! I summarized the metric refactor in the table below. In the first column, you can see the old metric class and string identifier. These are now replaced by new internal parent classes (second column), which are primarily used for inference-time grouping and shared computation. The third and fourth columns show the new atomic metric classes and their string identifiers, which are what users are expected to use going forward.

If someone tries to instantiate an old metric class, they'll now get an instance of the corresponding new internal parent class instead. This won’t break anything! the new internal parents still expose the .compute() method and return a dictionary just like the originals.

When a task is initialized using old class names or string identifiers, we also still support this, but issue a deprecation warning. Internally, the Evaluation Agent maps these to their respective new atomic metrics and groups them in the new internal parent classes only for inference efficiency.

| **Old Metric Class & Identifier**         | **New Internal Parent Class**     | **New Metric Classes**           | **New String Identifiers**   |
|------------------------------------------|-----------------------------------|----------------------------------|-------------------------------|
| `GPUMemoryMetric` `"gpu_memory"`         | `GPUMemoryStats`                    | `DiskMemoryMetric`               | `"disk_memory"`              |
|                                          |                                   | `InferenceMemoryMetric`          | `"inference_memory"`         |
|                                          |                                   | `TrainingMemoryMetric`           | `"training_memory"`          |
| `ElapsedTimeMetric` `"elapsed_time"`     | `InferenceTimeStats`                      | `TotalTimeMetric`                | `"total_time"`               |
|                                          |                                   | `LatencyMetric`                  | `"latency"`                  |
|                                          |                                   | `ThroughputMetric`               | `"throughput"`               |
| `ModelArchitectureMetric` `"model_architecture"` | `ModelArchitectureStats`           | `TotalParamsMetric`             | `"total_params"`             |
|                                          |                                   | `TotalMacsMetric`                | `"total_macs"`               |
| `EnergyMetric` `"energy"`                | `EnvironmentalImpactStats`               | `CO2ConsumedMetric`              | `"co2_emissions"`            |
|                                          |                                   | `EnergyConsumedMetric`           | `"energy_consumed"`          |

src/pruna/evaluation/metrics/metric_cmmd.py
begumcig: You're absolutely right that some part of the EvaluationAgent expects raw results, but that applies only  to `compute_stateless_metrics`, which handles inference sharing via a shared parent metric class. For stateful metrics, inference sharing is default, so there's no need to return raw results. In `compute_stateful_metrics`, the agent directly appends the output of `metric.compute()` to the results list. so for cmmd (or any stateful metric), we do need to return a MetricResult.

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: That was also my nightmare scenario!! I tried to handle this by sharing inference for new new metrics that internally share this computation logic. If you checkout https://github.com/PrunaAI/pruna/blob/81fcd3f4059accd3896f561696a0f4a12d6405ea/src/pruna/evaluation/evaluation_agent.py#L215

I tried to share inference for stateless metics whenever possible 🥹
So if you're using the EvaluationAgent you're doing all your computations efficiently

src/pruna/evaluation/metrics/result.py
begumcig: That’s a really good point, and honestly, I wasn’t 100% sure myself at first whether this warranted a separate data class. I talked a bit with Gabriel about it, especially since these results will eventually be logged, and we landed on dataclass as a middle ground between using a plain dictionary and bringing in something heavier like pydantic.

The main reason I kept MetricResult as a separate class is that even with its simplicity now (since the metrics themselves became more atomic, the MetricResult evolved into something simpler and simpler as this refactoring progressed), it’s still useful to have a structured place for validation. The small assertions I added here already helped catch a couple of subtle bugs during development, and I imagine we’ll want to add more checks or logging-related logic in the future.

That said, you're right,  it does overlap conceptually with some of the attributes in the metrics themselves (like higher_is_better), we can rethink the separation if you have a suggestion!!

guennemann: I personally think that a separate class is a good solution here. I actually could see to **additionally** add the higher_is_better is better here (it adds a bit redundancy but would make it much simply in downstream tasks to link the result to the property of the metric). As an alternative to just duplicating some attributes, one could add a pointer to the corresponding class the metric result is based on (and then the user can get the higher is better via the class).

One should potentially also discuss whether the pointer in general is nicer (in contrast to the string you are using currently in the metric result to identify what metric is used).

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -214,7 +211,14 @@ def compute_stateless_metrics(self, model: PrunaModel, stateless_metrics: List[A
         Dict[str, Any]
             The results of the stateless metrics.
         """
-        results = {}
-        for stateless_metric in stateless_metrics:
-            results.update(stateless_metric.compute(model, self.task.dataloader))
+        results = []
```
guennemann, line 227: It would be nice to add some comments explaining what this block is supposed to do (i.e. the context of avoiding redundant computation via grouping together...)

src/pruna/evaluation/metrics/utils.py
```diff
@@ -77,3 +88,208 @@ def metric_data_processor(
...
+    param_names = get_param_names_from_signature(sig)
+    return {name: getattr(instance, name, None) for name in param_names}
+
+
+def get_direct_parents(list_of_instances: List[Any]) -> Tuple[Dict[Any, List[Any]], List[Any]]:
+    """
+    Get the direct parents of a list of instances.
+
+    Returns a dictionary where the keys are the direct parents and the values are the direct children.
+    Also returns a list of instances that directly inherit from BaseMetric.
```
guennemann: Does it make sense to change the name of the function? Since you not only get the parents but also list of instances from the basemetric. It was a bit confusing to read the code in the agent (since there function returned two objects)

guennemann: Effectively you are subdividing the metrics in one that have now parent and once which have a parent -- and for the latter you then also directly do the grouping. Would be nice to reflect this a bit.

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -214,7 +211,14 @@ def compute_stateless_metrics(self, model: PrunaModel, stateless_metrics: List[A
...
+        parent_to_children, children_of_base = get_direct_parents(stateless_metrics)
+        for (parent, config), children in parent_to_children.items():
```
guennemann: it seems that config is actually not used?

src/pruna/evaluation/metrics/metric_base.py
```diff
@@ -19,6 +19,8 @@
 class BaseMetric(ABC):
     """The base class for all Pruna metrics."""
 
+    metric_name: str = ""
+
```
guennemann: Shouldn't we add "higher is better" and similar also already in the base metric?

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -123,10 +128,163 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
                     break
 
         total_elapsed_time = sum(list_elapsed_times)
-        batch_size = dataloader.batch_size
+        self.batch_size = cast(int, dataloader.batch_size)
```
guennemann, line 164: Why do we need to cast? Probably we should update the dataloader.

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -15,21 +15,26 @@
...
+from pruna.evaluation.metrics.result import MetricResult
 from pruna.logging.logger import pruna_logger
 
+LATENCY = "latency"
+THROUGHPUT = "throughput"
+TOTAL_TIME = "total_time"
 
-@MetricRegistry.register("elapsed_time")
-class ElapsedTimeMetric(BaseMetric):
+
```
guennemann, line 34: As a comment it might be helpful to add that this metric should not / will not be created/requested explicitely by the user -- but only implecitely via the child metrics.

src/pruna/evaluation/metrics/metric_elapsed_time.py
guennemann: For code understanding, it would be good to add this discussion as a comment (that the execution of the metrics via the agent will avoid redundancy)

src/pruna/evaluation/metrics/metric_energy.py
```diff
@@ -23,11 +24,14 @@
...
 from pruna.evaluation.metrics.registry import MetricRegistry
+from pruna.evaluation.metrics.result import MetricResult
 from pruna.logging.logger import pruna_logger
 
+ENERGY_CONSUMED = "energy_consumed"
+CO2_EMISSIONS = "co2_emissions"
 
-@MetricRegistry.register("energy")
-class EnergyMetric(BaseMetric):
+
```
guennemann, line 33: Same as for inferenceTime. Please clarify that the parent class will not be requested by the user.

src/pruna/evaluation/metrics/metric_energy.py
```diff
@@ -119,10 +123,8 @@ def _collect_emissions_data(self, tracker: EmissionsTracker) -> Dict[str, Any]:
...
             )
 
-        emissions_data["tracker_emissions"] = self._get_data(tracker.final_emissions_data, "emissions", "tracker")
-        emissions_data["tracker_energy_consumed"] = self._get_data(
-            tracker.final_emissions_data, "energy_consumed", "tracker"
-        )
+        emissions_data[CO2_EMISSIONS] = self._get_data(tracker.final_emissions_data, "emissions", "tracker")
+        emissions_data[ENERGY_CONSUMED] = self._get_data(tracker.final_emissions_data, "energy_consumed", "tracker")
 
         return emissions_data
```
guennemann, line 135: I am now wondering why we return a dict here and not also a list of metric results for consistency. The downside would be that the children will not be able to access easily/efficiently "its own" result (since we loose the hashMap). So probably we keep it in the current form.

src/pruna/evaluation/task.py
```diff
@@ -15,19 +15,36 @@
...
 from pruna.evaluation.metrics.registry import MetricRegistry
+from pruna.evaluation.metrics.utils import get_hyperparameters
 from pruna.logging.logger import pruna_logger
 
 AVAILABLE_REQUESTS = ("image_generation_quality",)
+PARENT_TO_CHILD_MAP = {
+    "ModelArchitectureStats": [TOTAL_MACS, TOTAL_PARAMS],
+    "InferenceTimeStats": [LATENCY, THROUGHPUT, TOTAL_TIME],
+    "EnvironmentalImpactStats": [ENERGY_CONSUMED, CO2_EMISSIONS],
+}
```
guennemann: Do we really have to explicily make this mapping here? Can't we gather this from the code/class structure automatically?

guennemann (commented):
Thanks for the nice refactoring. You can find detailed comments inline.

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: The dataloader is a pytorch dataloader, the casting is not really for functionality but rather for static type-checking.

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: Added to the Evaluation Agent under the computation of stateless metrics

src/pruna/evaluation/evaluation_agent.py
begumcig: Yes, it's making sure the keys in parent_to_children are unique, so we map the metrics with same hyperparameter together!

src/pruna/evaluation/metrics/metric_energy.py
begumcig: Yes exactly! Also for the people who are using the metric classes directly and not via the Agent I think it's nice that they still get their results in a dictionary form. We can deprecate it in the future but I didn't want to change how we return the results in the "legacy" class too much.

src/pruna/evaluation/metrics/metric_elapsed_time.py
guennemann: I think it might be good/better to also add it to the metric, since one stumbles upon this when reading _this_ file.

src/pruna/evaluation/evaluation_agent.py
guennemann: But config isn't used here, right? So can't you just replace it with `(parent, _)`

docs/user_manual/evaluate.rst
```diff
@@ -393,4 +423,4 @@ When first setting up evaluation, limit the dataset size with ``datamodule.limit
 Use pairwise metrics for comparison
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-When comparing an optimized model against the baseline, use pairwise metrics to get direct comparison scores.
\ No newline at end of file
+When comparing an optimized model against the baseline, use pairwise metrics to get direct comparison scores... _evaluation:
```
davidberenstein1957: is this processed correctly or should it be a new line?

docs/user_manual/evaluate.rst
```diff
@@ -196,24 +222,28 @@ The following table shows the different call types supported by |pruna| metrics
...
+The ``MetricResult`` is a class that contains the result of a metric evaluation.
 
-        .. code-block:: python
+Each metric returns a ``MetricResult`` instance, which contains the outcome of a single evaluation.
 
-            from pruna.evaluation.metrics import CMMD
+The ``MetricResult`` class stores the metric's name, any associated parameters, and the computed result value:
 
-            metric = CMMD() # or ["cmmd"]
+.. container:: hidden_code
```
davidberenstein1957, line 235: why hidden code?

docs/user_manual/evaluate.rst
```diff
@@ -158,7 +158,33 @@ Metric Call Types
...
+
+            from pruna.evaluation.metrics import CMMD
+
+            metric = CMMD(call_type="single") # or [CMMD() since single is the default call type]
+
+    .. tab:: Pairwise mode
+
+        .. code-block:: python
+
+            from pruna.evaluation.metrics import CMMD
```
davidberenstein1957, line 177:
```suggestion
            from pruna.evaluation.metrics import CMMD
            
```

docs/user_manual/evaluate.rst
```diff
@@ -196,24 +222,28 @@ The following table shows the different call types supported by |pruna| metrics
...
+.. code-block:: python
 
-            from pruna.evaluation.metrics import CMMD
-            metric = CMMD(call_type="pairwise")
+  # Example output
+  MetricResult(
+    name="clip_score",
+    params={"param1": "value1", "param2": "value2"},
+    result=28.0828
+  )
```
davidberenstein1957:
```suggestion
    MetricResult(
        name="clip_score",
        params={"param1": "value1", "param2": "value2"},
        result=28.0828
    )
```

docs/user_manual/evaluate.rst
```diff
@@ -196,24 +222,28 @@ The following table shows the different call types supported by |pruna| metrics
...
 
-            from pruna.evaluation.metrics import CMMD
+The ``MetricResult`` class stores the metric's name, any associated parameters, and the computed result value:
 
-            metric = CMMD() # or ["cmmd"]
+.. container:: hidden_code
 
-    .. tab:: Pairwise mode
+    .. code-block:: python
+      from pruna.evaluation.metrics.result import MetricResult
```
davidberenstein1957:
```suggestion

      from pruna.evaluation.metrics.result import MetricResult
```

docs/user_manual/evaluate.rst
davidberenstein1957: deoes this work correctly? also this `.. _evaluation` was also added to yor previous PR.

src/pruna/evaluation/metrics/metric_base.py
```diff
@@ -15,25 +15,33 @@
...
+    metric_units: str
+    higher_is_better: bool
 
     @abstractmethod
-    def compute(self, *args, **kwargs) -> Any:
+    def compute(
+        self,
+        model: PrunaModel,
+        dataloader: DataLoader,
+    ) -> Any:
```
davidberenstein1957, line 37: perhaps we should use string based type hinting here? as we are not using the classes?

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -85,48 +152,172 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
...
+        The number of warmup batches to evaluate the model.
+    device : str | torch.device, default="cuda"
+        The device to evaluate the model on.
+    timing_type : str, default="sync"
+        The type of timing to use.
+    """
+
+    higher_is_better: bool = False
+    metric_name: str = LATENCY
+    metric_units: str = "ms/num_iterations"
```
davidberenstein1957, line 194: should we add something like `MS_NUM_ITER`?

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -85,48 +152,172 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
...
+    n_warmup_iterations : int, default=10
+        The number of warmup batches to evaluate the model.
+    device : str | torch.device, default="cuda"
+        The device to evaluate the model on.
+    timing_type : str, default="sync"
+        The type of timing to use.
+    """
+
+    higher_is_better: bool = True
+    metric_units: str = "num_iterations/ms"
```
davidberenstein1957, line 237: same usage here

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -85,48 +152,172 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any]:
...
+        """Forwards to InferenceTimeStats."""
+        warn(
+            "Class ElapsedTimeMetric is deprecated and will be removed in a future release. \n"
+            "It has been replaced by InferenceTimeStats, \n"
+            "which is a shared parent class for 'LatencyMetric', 'ThroughputMetric' and 'TotalTimeMetric'. \n"
+            "In the future please use 'LatencyMetric', 'ThroughputMetric' or 'TotalTimeMetric' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return InferenceTimeStats(*args, **kwargs)
```
davidberenstein1957: When I deprecate, I normally add a specific version where it is going to be removed, +  I add a test that uses this deprecated class but will be removed during that version. @johannaSommer I also noticed you recently deprecated something, so perhaps it would be good to start doing this to make it easier to maintain.

src/pruna/evaluation/metrics/metric_memory.py
```diff
@@ -100,23 +106,37 @@ class GPUMemoryMetric(BaseMetric):
...
+                    "Please use 'DiskMemoryMetric', 'InferenceMemoryMetric' or 'TrainingMemoryMetric' instead.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                warn(
+                    f"Mode '{mode}' is deprecated and will be removed in a future release. "
+                    f"Please use '{DEPRECATED_MODES_MAP[mode]}' instead.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
davidberenstein1957: arent we using `logger.warn`?

src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -162,23 +198,28 @@ class TorchMetricWrapper(StatefulMetric):
...
+
+        Parameters
+        ----------
+        metric_name : str
+            Name of the metric.
+        call_type : str
+            Specifies the order and type of inputs to use for metric calculation.
+        """
+        # Special case for clip_score until new torchmetrics version.
+        if metric_name == "clip_score" and call_type.startswith(PAIRWISE):
```
davidberenstein1957, line 218: should we import the variable for the clip score name here?

docs/user_manual/evaluation.rst
johannaSommer: why was i even asking, of course you took care of everything perfectly duh

src/pruna/evaluation/metrics/metric_elapsed_time.py
johannaSommer: makes sense, thanks Begüm!

src/pruna/evaluation/metrics/result.py
johannaSommer: I thought about it again and I agree with you both now, let's keep them as seperate classes... In the future we might want to have different normalizations or scalings of the same metric, where this class would then come in handy for e.g. storing units (latency in ms or s) or normalizing some score to 0-100

src/pruna/evaluation/metrics/metric_elapsed_time.py
johannaSommer: thanks a lot for this feedback, yes, let's definitely do this! I will also take this into account

src/pruna/evaluation/metrics/metric_memory.py
johannaSommer: we did a bit of research on this and it was recommended to use `warn.warn` for deprecations regardless of your "own" logger, wdyt?

davidberenstein1957: I agree. I just thought I saw logger.warn somewhere else :) So, let's use `warn.warn`:)

docs/user_manual/evaluate.rst
begumcig: I just wanted to show an example of what a MetricResult looks like, didn't want to give the impression that users need to import it to use it when metrics return it :)

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: Sounds great! Added the version we stop supporting in the warning message + added tests for the deprecated classes in the tests.

src/pruna/evaluation/metrics/metric_base.py
begumcig: I actually think this is a really good discussion point and something we’ve also gone back and forth on. In other parts of the codebase, we’ve been importing annotations from __future__ to delay evaluation and ensure backward compatibility.. I am adding the import here as well, now.
I’m not totally sure if this is the best solution, though. From what I understand, using from __future__ import annotations generally removes the need for string type hints unless we’re supporting Python <3.7. what do you think?? :)

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: I’m only using this unit in this specific metric, so I wasn’t sure if it still makes sense to define a global constant for it. I went through all the metrics, and for the ones where I reuse the units across multiple places, I did create global strings. What do you think; would you prefer we keep all units as globals for consistency, or only when they’re reused?

src/pruna/evaluation/metrics/utils.py
```diff
@@ -77,3 +88,211 @@ def metric_data_processor(
...
+    ----------
+    call_type : str
+        The call type to get the pairing for.
+
+    Returns
+    -------
+    str
+        The pairing for the call type.
+    """
+    if call_type.startswith(PAIRWISE):
```
guennemann, line 225: Just to double check: This is correct, right? If it starts with PAIRWISE we return the single pairing?

src/pruna/evaluation/task.py
```diff
@@ -122,25 +138,58 @@ def get_metrics(request: str | List[str | BaseMetric]) -> List[BaseMetric]:
...
         elif all(isinstance(item, str) for item in request):
-            pruna_logger.info(f"Creating metrics from names: {request}")
-            metric_names: List[str] = cast(List[str], request)
-            return MetricRegistry.get_metrics(metric_names)
+            return _process_metric_names(cast(List[str], request))
         else:
-            pruna_logger.error("List must contain either all strings or all BaseMetric instances.")
-            raise ValueError("List must contain either all strings or all BaseMetric instances.")
+            pruna_logger.error("List must contain either all strings or all BaseMetric or StatefulMetric instances.")
+            raise ValueError("List must contain either all strings or all BaseMetric or StatefulMetric instances.")
```
guennemann: Super minor: Can you add brackets around (BaseMetric or StatefulMetric) to make sure that one can mix BaseMetric and StatefulMetric in one list.

src/pruna/evaluation/task.py
```diff
@@ -122,25 +138,58 @@ def get_metrics(request: str | List[str | BaseMetric]) -> List[BaseMetric]:
...
+            new_requests.append(cast(str, metric_name))
+    return MetricRegistry.get_metrics(new_requests)
+
+
+def _process_single_request(request: str) -> List[BaseMetric | StatefulMetric]:
+    if request == "image_generation_quality":
+        pruna_logger.info("An evaluation task for image generation quality is being created.")
+        return [
+            TorchMetricWrapper("clip_score"),
+            TorchMetricWrapper("clip_score", call_type="pairwise"),
```
guennemann, line 190: Is this still "pairwise" or PAIRWISE?

guennemann (approved):
We nice work! See comment for some very minor points.

src/pruna/evaluation/task.py
begumcig: I think when we’re creating an instance, we should stick to "pairwise", but within the metric class itself, it makes sense to use PAIRWISE as a kind of pseudo-literal to make sure everything aligns. That said, it’s mostly an aesthetic preference on my end :)

PR #85, opened 2025-04-30
fix: generation functions for `torch.compile` llm combination

johnrachwan123:
## Description
This PR fixes several issues in the torch_compile + quantization combo. When this combo was activated before this PR:

- The model can only take inputs and `max_new_tokens` as input and no other input
- The model always generates `max_new_tokens` number of tokens because the decoding function doesnt stop at EOS token
- The HQQ quantization was happening on the GPU and keeping the original model on GPU as well which could cause OOMs

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Tested on large Qwen 3 model and all unit tests still pass.

## Checklist

## Additional Notes
None.


src/pruna/algorithms/compilation/utils.py
```diff
@@ -87,21 +90,20 @@ def __init__(
...
-        self.device = model.device
+        self.device = device
         self.temperature = temperature
         self.top_k = top_k
         self.use_cache = True
         self.compile_mode = compile_mode
         self.compile_fullgraph = compile_fullgraph
         self.batch_size = batch_size
         self.cache_size = max_kv_cache_size
+        self.eos_token_id = getattr(model.config, "eos_token_id", None)
```
sharpenb, line 104: For some models (like chat), we will likely have problems with other specials tokens. In this case it will be likely easier to handle it via the tokenizer object.

src/pruna/algorithms/quantization/hqq.py
```diff
@@ -116,22 +117,27 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
 
         quant_config_hqq = imported_modules["BaseQuantizeConfig"](nbits=weight_quantization_bits, group_size=group_size)
         quant_config_hf = imported_modules["HqqConfig"](nbits=weight_quantization_bits, group_size=group_size)
-
+        if hasattr(model, "to"):
+            model_device = model.device
+            model.to("cpu")
```
sharpenb: We should use the `device` specified in the smash_config in order to mode the model on the device that should be sued during smashing :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -459,21 +462,28 @@ def next_token_iterator(
...
 
         Returns
         -------
         torch.Tensor
-            The generated tokens.
+            The generated tokens, including the input prompt and potentially an EOS token.
         """
-        self.setup(inputs=input_ids, max_new_tokens=max_new_tokens)
-        return self.next_token_iterator(self.prefill(), max_new_tokens)
+        self.setup(
```
sharpenb, line 544: I would suggest oa dd a warning saying that other arguments are ignored.

sharpenb (approved):
Nice quick fix! It should do the job for the next release even though I left a couple of comments that could be quickly fixed :) It would be nice in the PR description:
- to add a before/after to explain what was failing in terms of code snippet and now works e.g. a call with the a different generate signature, a call with large max length taking a lot of time, a call of HQQ on a small GPU. 
- to add future works including better other special token management, batch inference, memory management,

PR #87, opened 2025-05-02
fix: make kv cache dynamic based on input

johnrachwan123:
## Description
This PR makes the kv cache dynamic and the user doesnt need to worry about defining this anymore. It also means we are using the most efficient kv cache possible.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Unit tests + notebook.

## Checklist

## Additional Notes
We should remove batch size and cache_size after this change.

### DIFF
--- a/src/pruna/algorithms/compilation/utils.py
+++ b/src/pruna/algorithms/compilation/utils.py
@@ -295,7 +295,7 @@ class TransformersGenerator:
 
         # Check if batch size changed compared to the cache configuration
         # Round up max_new_tokens to the nearest 1000 for better memory allocation
-        rounded_cache_size = ((max_new_tokens + 999) // 1000) * 1000
+        rounded_cache_size = ((inputs.shape[1] + max_new_tokens + 999) // 1000) * 1000
         if new_batch_size != self.cache_batch_size or self.cache_size != rounded_cache_size:
             if new_batch_size != self.cache_batch_size:
                 pruna_logger.info(
--- a/src/pruna/algorithms/compilation/utils.py
+++ b/src/pruna/algorithms/compilation/utils.py
@@ -297,25 +297,21 @@ class TransformersGenerator:
         # Round up max_new_tokens to the nearest 1000 for better memory allocation
         rounded_cache_size = ((inputs.shape[1] + max_new_tokens + 999) // 1000) * 1000
         if new_batch_size != self.cache_batch_size or self.cache_size != rounded_cache_size:
-            if new_batch_size != self.cache_batch_size:
-                pruna_logger.info(
-                    f"Batch size changed from {self.cache_batch_size} to {new_batch_size}. Re-initializing StaticCache."
-                )
-            else:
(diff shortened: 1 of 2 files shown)


src/pruna/algorithms/compilation/utils.py
```diff
@@ -294,12 +294,20 @@ def setup(self, inputs: torch.Tensor, max_new_tokens: int):
         new_batch_size = inputs.shape[0]
 
         # Check if batch size changed compared to the cache configuration
-        if new_batch_size != self.cache_batch_size:
-            pruna_logger.info(
-                f"Batch size changed from {self.cache_batch_size} to {new_batch_size}. Re-initializing StaticCache."
-            )
+        # Round up max_new_tokens to the nearest 1000 for better memory allocation
+        rounded_cache_size = ((inputs.shape[1] + max_new_tokens + 999) // 1000) * 1000
```
sharpenb, line 298: To confirm the better memory allocation comes from the fact that we will not reinit the cache size too frequently. Is that correct? :)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -294,12 +294,20 @@ def setup(self, inputs: torch.Tensor, max_new_tokens: int):
...
+        if new_batch_size != self.cache_batch_size or self.cache_size != rounded_cache_size:
+            if new_batch_size != self.cache_batch_size:
+                pruna_logger.info(
+                    f"Batch size changed from {self.cache_batch_size} to {new_batch_size}. Re-initializing StaticCache."
+                )
+            else:
+                pruna_logger.info(
+                    f"Cache size changed from {self.cache_size} to {rounded_cache_size}. Re-initializing StaticCache."
+                )
```
sharpenb: I would suggest this

```
if new_batch_size != self.cache_batch_size or self.cache_size != rounded_cache_size:
  pruna_logger.info(
                      f"Cache size changed from {self.cache_batch_size}x{self.cache_size} to {new_batch_size}x{rounded_cache_size}. Re-initializing StaticCache."
                  )
```

src/pruna/algorithms/compilation/utils.py
```diff
@@ -294,12 +294,20 @@ def setup(self, inputs: torch.Tensor, max_new_tokens: int):
...
             self.batch_size = new_batch_size
             self.cache_batch_size = new_batch_size
+            self.cache_size = rounded_cache_size
             self.setup_cache()
 
             # If CUDA graph was used, it's now invalid
```
sharpenb: If we use `"max-autotune-no-cudagraphs"`, it will deactivate the compilation completely after the static cache reinit. The cuda graph is only activate once here so far: https://github.com/PrunaAI/pruna/blob/3aebee757a0e844537c235004fa693105d4e6a8d/src/pruna/algorithms/compilation/torch_compile.py#L350

sharpenb (approved):
Looks good! Left a couple of comments :)

PR #92, opened 2025-05-06
feat: add fastercache and pab

nifleisch:
## Description
This PR adds the PAB and FasterCache algorithms from diffusers (https://huggingface.co/docs/diffusers/main/api/cache).

## Type of Change

## How Has This Been Tested?
I manually tested both caching mechanisms across all supported Diffusers pipelines, visually inspected the resulting images, videos, and measured the inference time (relative speedups match [this benchmark](https://github.com/huggingface/diffusers/pull/9562). For FLUX, I evaluated every supported combination of algorithms. I implemented new tests for each algorithm and all of their combinations with other algorithms.

## Checklist

## Additional Notes
I tried to make these cachers work with compilation but ran into errors. The main problem is that these methods introduce a condition for the attention layer, which hinders compilation.

### DIFF
--- a/src/pruna/algorithms/caching/fastercache.py
+++ b/src/pruna/algorithms/caching/fastercache.py
@@ -1,0 +1,210 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+from typing import Any, Dict, Optional, Tuple
+
+from ConfigSpace import OrdinalHyperparameter
+
+from pruna.algorithms.caching import PrunaCacher
+from pruna.config.smash_config import SmashConfigPrefixWrapper
+from pruna.engine.model_checks import (
+    is_allegro_pipeline,
+    is_cogvideo_pipeline,
+    is_flux_pipeline,
+    is_hunyuan_pipeline,
+    is_latte_pipeline,
+    is_mochi_pipeline,
+    is_wan_pipeline,
+)
+from pruna.logging.logger import pruna_logger
+
+
+class FasterCacheCacher(PrunaCacher):
+    """
+    Implement FasterCache.
+
+    FasterCache is a method that speeds up inference in diffusion transformers by:
+    - Reusing attention states between successive inference steps, due to high similarity between them
+    - Skipping unconditional branch prediction used in classifier-free guidance by revealing redundancies between
+      unconditional and conditional branch outputs for the same timestep, and therefore approximating the unconditional
+      branch output using the conditional branch output
+    """
+
+    algorithm_name = "fastercache"
+    references = {"GitHub": "https://github.com/Vchitect/FasterCache", "Paper": "https://arxiv.org/abs/2410.19355"}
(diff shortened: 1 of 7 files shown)


johnrachwan123 (approved):
LGTM

sharpenb (changes requested):
Looks good! Mostly small comments before ti can be merged.

src/pruna/algorithms/caching/pab.py
```diff
@@ -0,0 +1,180 @@
...
+            The configuration for the caching.
+
+        Returns
+        -------
+        Any
+            The smashed model.
+        """
+        imported_modules = self.import_algorithm_packages()
+        # set default values
+        temporal_attention_block_skip_range: Optional[int] = None
```
sharpenb, line 115: I would still recommend to put them in the smash config as constant and mention that these can be overwritten for different architecture with a link to the code file or the diffuser PR so that the documentation is complete.

src/pruna/algorithms/caching/fastercache.py
```diff
@@ -0,0 +1,210 @@
...
+        unconditional_batch_timestep_skip_range: Tuple[int, int] = (-1, 641)
+        spatial_attention_block_identifiers: Tuple[str, ...] = (
+            "blocks.*attn1",
+            "transformer_blocks.*attn1",
+            "single_transformer_blocks.*attn1"
+        )
+        temporal_attention_block_identifiers: Tuple[str, ...] = ("temporal_transformer_blocks.*attn1",)
+        attention_weight_callback = lambda _: 0.5  # noqa: E731
+        tensor_format: str = "BFCHW"
+        is_guidance_distilled: bool = False
```
sharpenb, line 134: I would still recommend to put them in the smash config as constant and mention that these can be overwritten for different architecture with a link to the code file or the diffuser PR so that the documentation is complete :)

src/pruna/algorithms/caching/fastercache.py
```diff
@@ -0,0 +1,210 @@
...
+        smash_config : SmashConfigPrefixWrapper
+            The configuration for the caching.
+
+        Returns
+        -------
+        Any
+            The smashed model.
+        """
+        imported_modules = self.import_algorithm_packages()
+        # set default values
```
sharpenb: Here the link to the why these are default is not given :)

src/pruna/algorithms/caching/fastercache.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Returns
+        -------
+        list
+            The hyperparameters.
+        """
+        return [
+            OrdinalHyperparameter(
+                "interval",
+                sequence=[2, 3, 4, 5],
```
sharpenb: Could we doubled check that the interval parameter has the same sequence option for all caching methods with this hyperparameters. I have seen some of them having more or less flexibility e.g. sequence=[1, 2, 3, 4, 5].

sharpenb (commented):
I readded my comments that have been dropped in the last review ;)

src/pruna/algorithms/caching/fastercache.py
nifleisch: Good point. I will also support the value “1” for the sequence argument, which indicates that no caching is applied.

src/pruna/algorithms/caching/pab.py
nifleisch: I agree. The best solution would be to use the pipeline-specific defaults unless the user explicitly specifies a parameter. Unfortunately, our SmashConfig interface doesn’t support that logic right now. To implement it, we’d have to apply the same defaults across every pipeline. Given the large number of parameters—and the fact that most aren’t straightforward to tune—I’ll leave things as they are for now. This approach ensures users receive strong out-of-the-box results and a straightforward interface, albeit with fewer tuning options.

sharpenb: Makes sense. As discussed async, let's make clear in the PR descripiton that we have a pipeline specific default in this PR that would need some iterations :)

sharpenb (approved):
Let's go!

PR #98, opened 2025-05-06
fix: enhance model saving functionality to load multiple JSON configs present in diffusers

### LINKED ISSUE #96: [BUG] `PrunaModel.save_to_hub` does not work with `diffusers`
## Describe the bug

Diffusers uses a different `.json` file to stora model info.

## What I did

NA
```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

## Expected behavior

I would expect it to recognise arbitrary json.

## Environment
 - pruna version:  0.2.3.
 - python version: 3.12
 - Operating System: MacOS

davidberenstein1957:
- Updated `save_pruna_model_to_hub` to load all JSON files in the model directory, ensuring `smash_config.json` is present.
- Modified README template to reflect that model configuration is now stored in `*.json` files instead of a single `config.json` file.
- Adjusted code examples for clarity and consistency.
- Resolved bug where loading saved `SmashConfig` does not have locally stored `smash_config.json` in cache dir after loading from hub
- Added integration pytest marker

## Description

Error reproduction code using another branch
```python
from pruna import PrunaModel

model_id = "PrunaAI/tiny-stable-diffusion-pipe-smashed"
loaded_model = PrunaModel.from_hub(model_id)
loaded_model.save_to_hub(model_id)
```

## Related Issue

Fixes #96 
## Type of Change

## Checklist

### DIFF
--- a/src/pruna/engine/handler/handler_utils.py
+++ b/src/pruna/engine/handler/handler_utils.py
@@ -47,7 +47,9 @@ def register_inference_handler(model: Any) -> InferenceHandler:
     if handler is not None:
         return handler
 
-    model = model.model if hasattr(model, "model") else model
+    import pdb
+
+    pdb.set_trace()
     if "diffusers" in model.__module__:
         return DiffuserHandler(call_signature=inspect.signature(model.__call__))
     elif "transformers" in model.__module__:
--- a/src/pruna/engine/handler/handler_utils.py
+++ b/src/pruna/engine/handler/handler_utils.py
@@ -47,6 +47,7 @@ def register_inference_handler(model: Any) -> InferenceHandler:
     if handler is not None:
         return handler
 
+    model = model.model if hasattr(model, "model") else model
     if "diffusers" in model.__module__:
         return DiffuserHandler(call_signature=inspect.signature(model.__call__))
     elif "transformers" in model.__module__:
--- a/src/pruna/engine/handler/handler_utils.py
+++ b/src/pruna/engine/handler/handler_utils.py
@@ -47,9 +47,6 @@ def register_inference_handler(model: Any) -> InferenceHandler:
     if handler is not None:
         return handler
 
-    import pdb
-
-    pdb.set_trace()
     if "diffusers" in model.__module__:
         return DiffuserHandler(call_signature=inspect.signature(model.__call__))
     elif "transformers" in model.__module__:
--- a/src/pruna/engine/save.py
+++ b/src/pruna/engine/save.py
@@ -54,6 +54,12 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) -> 
     if not os.path.exists(model_path):
         os.makedirs(model_path)
 
+    # save old smash config (includes tokenizer and processor)
+    save_dir = os.path.join(smash_config.cache_dir, SAVE_BEFORE_SMASH_CACHE_DIR)
+    if not os.path.exists(save_dir):
+        os.makedirs(save_dir)
+    smash_config.save_to_json(save_dir)
+
     if SAVE_FUNCTIONS.torch_artifacts.name in smash_config.save_fns:
         save_torch_artifacts(model, model_path, smash_config)
         smash_config.save_fns.remove(SAVE_FUNCTIONS.torch_artifacts.name)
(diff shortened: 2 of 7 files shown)


src/pruna/engine/save.py
```diff
@@ -138,21 +145,29 @@ def save_pruna_model_to_hub(
...
+        # Load all JSON files in the model directory
+        config_files = {}
+        for json_file in model_path_pathlib.glob("*.json"):
+            with json_file.open() as f:
+                config_files[json_file.stem] = json.load(f)
+
+        # Ensure smash_config is loaded
+        smash_config_file = Path(SMASH_CONFIG_FILE_NAME).stem
+        if smash_config_file not in config_files:
+            raise ValueError("smash_config.json not found in model directory")
```
davidberenstein1957:
```suggestion
            raise ValueError(f"{SMASH_CONFIG_FILE_NAME} not found in model directory")
```

src/pruna/engine/handler/handler_utils.py
```diff
@@ -47,6 +47,7 @@ def register_inference_handler(model: Any) -> InferenceHandler:
     if handler is not None:
         return handler
 
+    model = model.model if hasattr(model, "model") else model
```
begumcig: I am a bit worried about this line 🥹 In engine.pruna_model we initialize the handler using `model.model`, already, and we use some pipelines that also has an internal `.model` attribute, (like automatic speech recognition pipelines). I am a bit worried that this may break how inference handler works. Maybe we can find another way, what's the purpose of this line?

begumcig (changes requested):

tests/engine/test_save.py
```diff
@@ -1,20 +1,38 @@
 import os
-import shutil
-
 import pytest
-
 from pruna.engine.pruna_model import PrunaModel
-from transformers import AutoModelForCausalLM
+from transformers import AutoModelForCausalLM, AutoTokenizer
```
gsprochette, line 4:
```suggestion
from transformers import AutoModelForCausalLM
```

gsprochette:
```suggestion
from transformers import AutoModelForCausalLM
```

tests/engine/test_load.py
```diff
@@ -5,7 +5,8 @@
 @pytest.mark.parametrize(
     "model_name, expected_output, should_raise",
     [
-        ("PrunaAI/opt-125m-smashed", "PrunaModel", False),
+        ("PrunaAI/test-tiny-random-llama4-smashed", "PrunaModel", False),
+        ("PrunaAI/test-tiny-stable-diffusion-pipe-smashed", "PrunaModel", False),
         ("NonExistentRepo/model", None, True),
```
gsprochette: These are the same model_ids as those in the save tests. Shouldn't we have separate models to make sure the tests remain the same? Right now we risk having weight behaviour if `test_save.py` writes nonsense in there, of if the two tests run at the same time on different machines, no?

src/pruna/engine/save.py
```diff
@@ -134,25 +141,30 @@ def save_pruna_model_to_hub(
...
-        model_path_pathlib = Path(model_path)
 
-        # Open and load the model configuration and smash configuration data from their respective files
-        model_config_path = model_path_pathlib / "config.json"
-        smash_config_path = model_path_pathlib / "smash_config.json"
-        with model_config_path.open() as config_file, smash_config_path.open() as smash_config_file:
-            model_config = json.load(config_file)
-            smash_config_data = json.load(smash_config_file)
+        # Load the smash config
+        with open(os.path.join(model_path, SMASH_CONFIG_FILE_NAME), "r") as f:
```
gsprochette: If you define the `model_path_pathlib` can't we do something like this?
```suggestion
        with (model_path_pathlib / SMASH_CONFIG_FILE_NAME).open() as f:
```

src/pruna/engine/save.py
```diff
@@ -54,6 +54,12 @@ def save_pruna_model(model: Any, model_path: str, smash_config: SmashConfig) ->
     if not os.path.exists(model_path):
         os.makedirs(model_path)
 
+    # save old smash config (includes tokenizer and processor)
+    save_dir = os.path.join(smash_config.cache_dir, SAVE_BEFORE_SMASH_CACHE_DIR)
+    if not os.path.exists(save_dir):
+        os.makedirs(save_dir)
+    smash_config.save_to_json(save_dir)
+
```
gsprochette: I don't understand where this is being used. I assume this was added because save_before_apply was missing a smash_config, bug it seems like a different issue than what this PR addresses. Where and why is the smash_config in this dir being loaded later? The comment here should describe why we need to save the smash_config rather than the fact we save it.

gsprochette (commented):
Looks good already, but needs an iteration before it's good to go :) I left a couple of suggestions and there is the question about saving in `SAVE_BEFORE_SMASH_CACHE_DIR` which I didn't get.

gsprochette (approved):
LGTM! Thanks for adressing all the comments :+1: :sparkles:

begumcig (approved):
LGTM!!!

PR #103, opened 2025-05-06
feat: add device validation utility for improved device management

### LINKED ISSUE #97: [BUG] `SmashConfig` sets device to `cuda` even if it is not available.
## Describe the bug

I Smashed a model without specifying a device but it doe not want to load because thje config is set to cuda.

## What I did

```python

base_model = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4")

# Create and smash your model
smash_config = SmashConfig()
smash_config["cacher"] = "deepcache"
smashed_model = smash(model=base_model, smash_config=smash_config)
smashed_model.save_to_hub("dataset-viber/stable-diffusion-v1-4-smashed")

```

## Expected behavior

I would expect to only use cuda if it is available.

## Environment
 - pruna version: 0.2.3
 - python version: 3.12
 - Operating System: MacOS

davidberenstein1957:
* refactor: update SmashConfig to use validate_device for device assignment
* add: implement validate_device function to check and return the best available device
* enhance: log warnings for device availability and fallback scenarios

## Description

Reproduction of error with a device without `cuda`.
```python
from pruna import PrunaModel

loaded_model = PrunaModel.from_hub("PrunaAI/opt-125m-smashed")
```

## Related Issue

Fixes #97 

## Type of Change

## Checklist


tests/config/test_device.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@pytest.mark.cpu
+def test_device_none() -> None:
+    """Test that setting device to None defaults to 'cuda'."""
+    smash_config = SmashConfig(device=None)
+    assert smash_config.device == DEVICE
+
+@pytest.mark.cpu
+def test_device_available() -> None:
```
gsprochette: I may be wrong, but is this really testing what is described in the docstring? Because it's marked as cpu, `DEVICE` would (or at least could) be "cpu", right? Same thing for test_device_default. I don't like using `DEVICE` there, I would prefer tests marked as cuda that assume that `torch.cuda.is_available()` is True, instead of suspecting that the test passed because it was dispatched on a machine without a GPU.

src/pruna/config/smash_config.py
```diff
@@ -67,7 +67,7 @@ class SmashConfig:
     def __init__(
         self,
         max_batch_size: int = 1,
-        device: str = "cuda",
+        device: str | None = None,
```
gsprochette: The default value in the docstring should be updated following the logic of `check_device_compatibility`

src/pruna/config/smash_config.py
```diff
@@ -406,6 +406,48 @@ def check_argument_compatibility(self, algorithm_name: str) -> None:
...
+            The device to validate (e.g. 'cuda', 'mps', 'cpu').
+
+        Returns
+        -------
+        str
+            The best available device.
+        """
+        if device is None:
+            if torch.cuda.is_available():
+                pruna_logger.warning("No device specified. Using best available device: CUDA.")
```
gsprochette: This is the default (and backward compatible) behaviour so maybe in this specific case it should be an info instead of a warning?

gsprochette (commented):
Thanks a lot for taking care of this bug!
- Could you please provide a script reproducing the bug and making sure that it is solved with this fix? The bug mentionned uploading to the hub, which is not explicitely done in the tests added, so I need to make sure.
- I left a couple of comments, nothing too complicated, once they are addressed it should be good to go :)

src/pruna/config/smash_config.py
```diff
@@ -56,8 +57,9 @@ class SmashConfig:
     ----------
     max_batch_size : int, optional
         The maximum number of batches to process at once. Default is 1.
-    device : str, optional
-        The device to be used for smashing, e.g., 'cuda' or 'cpu'. Default is 'cuda'.
+    device : str | None, optional
```
gsprochette: The type changed again so now it should be "str | torch.device | None" :upside_down_face:

gsprochette (approved):
Looks great! The docstring needs one more update, once this is done it's good to go :)

src/pruna/config/smash_config.py
davidberenstein1957:
```suggestion
    device : str | torch.device | None, optional
```

src/pruna/engine/utils.py
```diff
@@ -241,3 +241,42 @@ def determine_dtype(pipeline: Any) -> torch.dtype:
 
     pruna_logger.warning("Could not determine dtype of model, defaulting to torch.float32.")
     return torch.float32
+
+
+def check_device_compatibility(device: str | torch.device | None) -> str:
```
begumcig, line 246: 🔥🔥🔥🔥

begumcig (approved):
LGTM!!! 🚀🚀🚀🚀

PR #105, opened 2025-05-07
feat: add option for compilation for module lists

johnrachwan123:
## Description
This PR adds the possibility to compile the module lists inside of a model instead of the full forward. The benefits are:

1. The compilation is 2x faster
2. There is less chance of it failing due to graph breaks

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Ran on different combinations.

## Checklist

## Additional Notes
None.

### DIFF
--- a/src/pruna/algorithms/compilation/torch_compile.py
+++ b/src/pruna/algorithms/compilation/torch_compile.py
@@ -126,7 +126,7 @@ class TorchCompileCompiler(PrunaCompiler):
                     desc=(
                         "Whether to compile the model itself or the module list. "
                         "Compiling the model itself has a longer warmup and could fail "
-                        "incase of graphbreaks but could lead to slightly faster compilation. "
+                        "in case of graphbreaks but could lead to slightly faster compilation. "
                         "Whereas compiling the module list has a shorter warmup and is more "
                         "robust to graphbreaks but could be slightly slower."
                     )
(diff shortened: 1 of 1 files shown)


src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -118,6 +118,20 @@ def get_hyperparameters(self) -> list:
...
             ),
+            OrdinalHyperparameter(
+                "target",
+                default_value="model",
+                sequence=["model", "module_list"],
+                meta=dict(
+                    desc=(
+                        "Whether to compile the model itself or the module list. "
+                        "Compiling the model itself has a longer warmup and could fail "
+                        "incase of graphbreaks but could lead to slightly faster compilation. "
```
johannaSommer: type: in case of graph breaks

johannaSommer (approved):
There is a small typo but approving already, looks great! 🥳

johannaSommer:
maybe one more idea: in case we can not find a module list, should we raise a warning?

nifleisch (approved):
Great work on this PR! 
One thing I’m worried about is that limiting compilation to only ModuleList instances means some modules may remain uncompiled. This could be problematic when applying quantization since the uncompiled quantized modules can be very slow. Still, since this compile mode is optional I it should be good for now.

johnrachwan123:
> Great work on this PR! One thing I’m worried about is that limiting compilation to only ModuleList instances means some modules may remain uncompiled. This could be problematic when applying quantization since the uncompiled quantized modules can be very slow. Still, since this compile mode is optional I it should be good for now.

Good point, maybe in the future we can extend this to compile all the known torch layers.

PR #106, opened 2025-05-07
feat: add fora cacher

nifleisch:
## Description
Add FORA caching for FLUX pipelines. 

## Type of Change

## How Has This Been Tested?
I tested all algorithm combinations for FLUX [dev]. I did a visual evaluation of the outputs and speedups. I tested the `set_params`, `enable` and `disable` methods of the cache helper. I implemented an integration test and combination tests, which all pass locally.

## Checklist

## Additional Notes
I did not add a tutorial for this new feature yet as there will be a tutorial for all recently added cachers including FORA, PAB, and FasterCache.

### DIFF
--- a/src/pruna/algorithms/caching/fora.py
+++ b/src/pruna/algorithms/caching/fora.py
@@ -1,0 +1,319 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import functools
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+from ConfigSpace import OrdinalHyperparameter
+
+from pruna.algorithms.caching import PrunaCacher
+from pruna.config.smash_config import SmashConfigPrefixWrapper
+from pruna.engine.model_checks import is_flux_pipeline
+
+
+class FORACacher(PrunaCacher):
+    """
+    Implement FORA for the Flux pipeline.
+
+    FORA reuses the outputs of the transformer blocks for N steps before recomputing them.
+    Different from the official implementation, this implementation exposes a start step parameter
+    that allows to obtain a higher fidelity to the base model.
+    """
+
+    algorithm_name: str = "fora"
+    references: dict[str, str] = {"Paper": "https://arxiv.org/abs/2407.01425"}
+    tokenizer_required: bool = False
+    processor_required: bool = False
+    run_on_cpu: bool = True
+    run_on_cuda: bool = True
+    dataset_required: bool = False
+    compatible_algorithms: dict[str, list[str]] = dict(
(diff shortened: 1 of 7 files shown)


src/pruna/algorithms/caching/fora.py
```diff
@@ -0,0 +1,319 @@
...
+            OrdinalHyperparameter(
+                "interval",
+                sequence=range(1, 6),
+                default_value=2,
+                meta=dict(desc="Interval at which the outputs are computed. Higher is faster, but reduces quality."),
+            ),
+            OrdinalHyperparameter(
+                "start_step",
+                sequence=range(11),
+                default_value=0,
```
johnrachwan123: Can we put the default value to 2 ? I know that 0 is the default in fora but its objectively really bad and I would prefer that the default is something that is nice for out users you know ?

johnrachwan123 (changes requested):
LGTM, this is a great contribution to the OSS community, Just a tiny detail change suggestion 😉

src/pruna/algorithms/caching/fora.py
```diff
@@ -0,0 +1,319 @@
...
+class FORACacher(PrunaCacher):
+    """
+    Implement FORA for the Flux pipeline.
+
+    FORA reuses the outputs of the transformer blocks for N steps before recomputing them.
+    Different from the official implementation, this implementation exposes a start step parameter
+    that allows to obtain a higher fidelity to the base model.
+    """
+
+    algorithm_name: str = "fora"
```
sharpenb, line 36: Is the old name included in the deprecated list now? :)

src/pruna/algorithms/caching/fora.py
```diff
@@ -0,0 +1,319 @@
...
+
+        Parameters
+        ----------
+        block : Any
+            The double stream transformer block to wrap.
+        layer : int
+            The layer of the double stream transformer block.
+        """
+        block_forward = block.forward
+        self.double_stream_blocks_forward.append(block_forward)
```
sharpenb: It sounds a bit dangerous to use a list with append here since it assumes that the wrap functions are called in the layer order. While it is the case in these functions it sounds safer to use array with indices access which would also be faster in tersm of complexity in wrapped_forward.

src/pruna/algorithms/caching/fora.py
```diff
@@ -0,0 +1,319 @@
...
+
+        Parameters
+        ----------
+        block : Any
+            The single stream transformer block to wrap.
+        layer : int
+            The layer of the single stream transformer block.
+        """
+        block_forward = block.forward
+        self.single_stream_blocks_forward.append(block_forward)
```
sharpenb: See above :)

sharpenb (changes requested):
Looks very nice! Some small comments :)

sharpenb (changes requested):
> Looks very nice! Some small comments :) 

Forgot also to check if the documentation has been updated accordingly ;)

src/pruna/algorithms/caching/fora.py
nifleisch: The legacy “flux_caching” algorithm remains in the Pruna Pro package but will be deprecated soon (users are already seeing deprecation warnings when using it). I didn’t add it to the deprecation list because Fora uses slightly different argument names, which would lead to errors under our current deprecation logic. To standardize argument naming for all cachers in Pruna, I’ve renamed “cache_interval” to “interval.”

src/pruna/algorithms/caching/fora.py
nifleisch: I agree. This change also makes `self.double_stream_blocks_forward` more consistent with the caches, which are dictionaries indexed by layer.

sharpenb (approved):
Thanks for the udpate! Looks good :)

PR #107, opened 2025-05-07
feat: add `configure_environment` function to `SmashConfig`

johannaSommer:
## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


src/pruna/config/smash_config.py
```diff
@@ -35,8 +37,6 @@
 from pruna.logging.logger import pruna_logger
 
 ADDITIONAL_ARGS = [
-    "batch_size",
```
sharpenb, line 38: It is still unclear if we shoudl use the batch_size from parameter to optimize over. It still feel that it can have a big impact e.g. on the throughput :)

src/pruna/config/smash_config.py
```diff
@@ -408,21 +399,84 @@ def add_processor(self, processor: str | ProcessorMixin) -> None:
...
+        max_seq_len: int = 512,
+        diffusion_steps_arg_name: str = "num_inference_steps",
+        diffusion_backbone_attr_name: str = "model",
+        diffusion_backbone_call_method: str = "__call__",
+        enable_cpu_offload: bool = False,
+        external_logging: str = "none",
+    ) -> None:
         """
-        Get a tokenizer object from a tokenizer name.
+        Configure the environment for the SmashConfig.
```
sharpenb: It would be nice to make clearer what we mean by "configure environment"

src/pruna/config/smash_config.py
```diff
@@ -408,21 +399,84 @@ def add_processor(self, processor: str | ProcessorMixin) -> None:
         else:
             self.processor = processor
 
-    def get_tokenizer_name(self) -> str | None:
+    def configure_environment(
+        self,
+        device: str = "cuda",
```
sharpenb, line 410: It feels like we could make this function less cluttered with all variables that belong to what we call the environment :)

src/pruna/algorithms/quantization/torch_static.py
```diff
@@ -134,7 +129,7 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
 
         # dataloader has been ensured to be set in the config
         for i, batch in enumerate(smash_config.train_dataloader()):  # type: ignore[arg-type]
-            if i >= smash_config["calibration_samples"]:
+            if i >= smash_config.calibration_samples:
```
sharpenb, line 132: From a user experience, it would be ncie that they can still access these fields with [""].

sharpenb (changes requested):
I left some early comments :) We could discuss some design that would differentiate environment, fixed, and free hyperparameters with tags :)

src/pruna/config/smash_config.py
```diff
@@ -408,21 +399,84 @@ def add_processor(self, processor: str | ProcessorMixin) -> None:
...
+        batch_size: int = 1,
+        saveable_model: bool = True,
+        calibration_samples: int = 16,
+        call_function_name: str = "__call__",
+        max_seq_len: int = 512,
+        diffusion_steps_arg_name: str = "num_inference_steps",
+        diffusion_backbone_attr_name: str = "model",
+        diffusion_backbone_call_method: str = "__call__",
```
nifleisch, line 418: I really like having information about the diffusers backbone in the smash config.

Right now, we duplicate the logic to figure out the backbone in several places. So I was thinking if we could instead infer the backbone automatically when `smash` is called (in case the backbone is not set manually by the user)? Then any algorithm that needs to know the backbone of a diffusers model could just get it from the config.

src/pruna/config/smash_config.py
```diff
@@ -408,21 +399,84 @@ def add_processor(self, processor: str | ProcessorMixin) -> None:
...
+            The device for which the model is optimized. Options are "cuda" and "cpu".
+        batch_size : int, default=1
+            The inference batch size to optimize the model for. Attention: some algorithms optimize the model for exactly
+            this batch size and changing this at inference time might lead to less performance gains.
+        saveable_model : bool, default=True
+            Whether the model needs to be saveable after optimization, e.g. with ``smashed_model.save_pretrained()``.
+            Setting this to false can lead to faster smashing.
+        calibration_samples : int, default=16
+            The number of calibration samples to be used for algorithms that perform calibration.
+        call_function_name : str, default="__call__"
```
nifleisch, line 437: Additionally to the existing diffusion arguments it could be helpful to have an argument like "is_guidance_distilled". Depending on whether a model is guidance distilled the backbone is called 1 or 2 times for every step. This information can be important for caching algorithms.

src/pruna/config/smash_config.py
```diff
@@ -607,6 +675,64 @@ def __str__(self) -> str:  # noqa: D105
...
+                f"The {name} attribute is deprecated and no longer needed, as they will be determined automatically.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            name = None
+
+        elif name == "_prepare_saving":
+            warn(
+                "The _prepare_saving attribute is deprecated. Please configure this setting by calling "
+                "smash_config.configure_environment(saveable_model=True/False).",
```
nifleisch: I am inconclusive if I like this way of configuring the environment or if I would like something like:
```
smash_config["savable_model"] = True
```
or 
```
smash_config["environment_savable_model"] = True
```
more.

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -174,9 +158,11 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         Any
             The compiled model.
         """
-        cacher_type = smash_config["cacher"]
-        if cacher_type in compilation_map:
-            return compilation_map[cacher_type](model, smash_config)
+        if smash_config["quantizer"] in compilation_map:
```
nifleisch, line 175: I think we should handle methods like `torchao.autoquant` differently -refer to the PR in Pruna Pro for more details.

nifleisch:
Really exciting PR! I believe adding options to customize the Pruna engine to diffusion models beyond the diffusers interface is very valuable. I have left some comments. I am not sure, if the added logic regarding the interplay between torch.compile and quantizers should be part of this PR. We might not even need it.

johnrachwan123 (approved):
As long as we can clearly document this for the user it LGTM!

src/pruna/config/smash_config.py
johannaSommer: yes of course! adjusted

src/pruna/config/smash_config.py
johannaSommer: you are referring to e.g. the unet or the transformer correct? I really love the idea! The user default could specify something like "auto" and then we try to determine it, if this fails we give an error and ask the user to specify it through the configure environment?

100% agree, however, I would suggest we make this a seperate task, as otherwise the PR becomes a bit large? For now my goal would be for the smashing to work somewhat the same and adjust the interface and then we can do what we just discussed now in a second step?

src/pruna/config/smash_config.py
johannaSommer: can you specify this a bit more? I don't thing I understand

src/pruna/config/smash_config.py
johannaSommer: completely agree! Is an argument like this already used somewhere where I can replace this?

src/pruna/config/smash_config.py
johannaSommer: added this, see sister PR in internal repository

src/pruna/algorithms/quantization/torch_static.py
johannaSommer: added

src/pruna/config/smash_config.py
johannaSommer: for now i went with option 1

src/pruna/config/smash_config.py
nifleisch: Yes, it is currently used in the custom cache helper.

nifleisch (commented):
One more idea that could fit into this PR is allowing users to exclude specific modules from quantization. Right now, you can’t specify module names in the smash config, so we maintain hard-coded lists of common norm and embedding module names (see the torchao PR for an example). A more flexible approach would be to let users provide their own exclusion list via the environment. Should we include this feature in the current PR, or include it in a separate PR?

src/pruna/config/smash_config.py
nifleisch: But I think I did not use that exact name for it. I think the argument is `num_backbone_calls_per_step` which is 1 if it is guidance distilled and 2 otherwise.

sharpenb (commented):
Thanks for the udpate. I left my thoughts in the sister PR :)

PR #109, opened 2025-05-09
test: use tiny random models to speed up tests

nifleisch:
## Description
Hugging Face offers “tiny-random” model versions for many popular models. This PR integrates these tiny random models into our test suite to speed up several test suites. On my g6 EC2 instance, the changes in this PR reduced the algorithm tests from **430 s → 215 s**, the combination tests from **322 s → 194 s**, and the evaluation tests from **570 s → 141 s**.

Moreover, because the random tiny models are much smaller, the initial download and setup for our tests is also significantly faster.

Since many of our tests (e.g., algorithm integration and combination tests) do not require models to generate meaningful outputs, using random tiny models is appropriate.

## Type of Change

## How Has This Been Tested?
I ran all affected tests multiple times and confirmed they pass consistently with the tiny random models.

## Checklist

## Additional Notes
This PR only modifies existing tests. Adopting tiny random models may also enable us to expand test coverage.


johnrachwan123:
thanks a lot for the PR! This is very exciting! I was wondering if you know why the algorithm tests take 50% less time and not more ?

nifleisch:
> thanks a lot for the PR! This is very exciting! I was wondering if you know why the algorithm tests take 50% less time and not more ?

To be honest, I was also surprised we didn’t see bigger speedups with these changes. Loading and smashing tiny random models in a script runs way more than twice as fast compared to doing the same with the original model. One factor that could have artificially improved the times for the old tests is that I ran the tests multiple times before measuring. Many of the larger models used in the test may have already been cached in RAM, so loading them was faster. We should investigate this further.

tests/evaluation/test_cmmd.py
```diff
@@ -13,8 +13,8 @@
 @pytest.mark.parametrize(
     "model_fixture, device, clip_model",
     [
-        pytest.param("stable_diffusion_v1_4", "cuda", "openai/clip-vit-large-patch14-336", marks=pytest.mark.cuda),
-        pytest.param("ddpm-cifar10", "cpu", "openai/clip-vit-large-patch14-336", marks=pytest.mark.cpu),
```
begumcig, line 17: Can we keep the ddpm model, but make it the CUDA test for this metric? :)

tests/fixtures.py
```diff
@@ -110,10 +110,9 @@ def get_automodel_transformers(model_id: str) -> tuple[Any, SmashConfig]:
     if hasattr(smash_config.tokenizer, "pad_token"):
         smash_config.tokenizer.pad_token = smash_config.tokenizer.eos_token
         dataset = PrunaDataModule.from_string(
-            "WikiText", collate_fn_args=dict(tokenizer=smash_config.tokenizer, max_seq_len=2048)
+            "WikiText", collate_fn_args=dict(tokenizer=smash_config.tokenizer, max_seq_len=64)
         )
-        # avoid too long calibration time, otherwise it kills the cpu
-        dataset.limit_datasets(256)
+        dataset.limit_datasets(16)
```
begumcig, line 116: slay

tests/fixtures.py
```diff
@@ -128,25 +127,41 @@ def get_torchvision_model(name: str) -> tuple[Any, SmashConfig]:
...
+        "dg845/tiny-random-stable-diffusion"
+    ),
     "sana": partial(
         get_diffusers_model,
         SanaPipeline,
         "Efficient-Large-Model/Sana_600M_512px_diffusers",
         variant="fp16",
         torch_dtype=torch.float16,
     ),
-    "ddpm-cifar10": partial(get_diffusers_model, DDIMPipeline, "google/ddpm-cifar10-32"),
```
begumcig, line 150: yes please let's keep this model for testing the unconditional image generation 🥹

begumcig (changes requested):
Amazing, show stopping, brilliant!!! left really smol comments! Let's make these tests FAST 💨💨💨💨

johannaSommer (approved):
I love it!! As a future suggestion (you know best whether now is the time or later), it would be great to somehow have a pairing between tiny and normal model. I.e. some function that is like
```python
get_fixture("stable_diffusion", tiny=True)
```
Because then we can run the tests in "full" or in "tiny" by just some marker, e.g. `pytest --tiny`. Nevertheless, definitely a contender for contribution of the quarter

johnrachwan123:
> > thanks a lot for the PR! This is very exciting! I was wondering if you know why the algorithm tests take 50% less time and not more ?
> 
> To be honest, I was also surprised we didn’t see bigger speedups with these changes. Loading and smashing tiny random models in a script runs way more than twice as fast compared to doing the same with the original model. One factor that could have artificially improved the times for the old tests is that I ran the tests multiple times before measuring. Many of the larger models used in the test may have already been cached in RAM, so loading them was faster. We should investigate this further.

I think you're intuition makes sense about the models being pre-downloaded

begumcig (approved):
LGTM!!!!! Amazing!!!

PR #110, opened 2025-05-10
feat: add torchao quantizer

nifleisch:
## Description
This PR introduces torchao as an additional quantizer for diffusion transformer models (e.g., FLUX). When combined with torch.compile, torchao delivers significant inference speedups for FLUX:

- Baseline (no optimization): 16 s
- torch.compile: 12.2 s
- torchao + torch.compile: 9.5 s

All timings were measured on an L40S GPU. Full benchmark (including visual comparison) are available [here](https://pruna.notion.site/torchao-bench-1ee70a039e5f80f88007d577e5fd4229?pvs=4).

The implementation closely follows the [diffusers-torchao](https://github.com/sayakpaul/diffusers-torchao) repository, which includes additional benchmarks and tests for the CogVideoX-5b video model.

## Type of Change

## How Has This Been Tested?

- Added a dedicated algorithm test
- Manually tested all quantization modes on FLUX, both with and without torch.compile.
- Manually tested quantization of LLama model.
- Verified model quality via a small benchmark (see link above).
Note: fp8dqrow mode was not tested due to its requirement for Hopper-series GPUs. Also the int4 quantization techniques are not compatible with FLUX.

## Checklist

### DIFF
--- a/src/pruna/algorithms/quantization/torchao.py
+++ b/src/pruna/algorithms/quantization/torchao.py
@@ -209,7 +209,8 @@ class TorchaoQuantizer(PrunaQuantizer):
             )
         # Only apply quantization on module list level if torch compile is also applied at that level
         if (
-            smash_config._base_config["torch_compile_target"] == "model"
+            smash_config["compiler"] == "torch_compile"
+            and smash_config._base_config["torch_compile_target"] == "model"
             or smash_config["compiler"] is None
             or smash_config["compiler"] != "torch_compile"
         ):
--- a/src/pruna/algorithms/quantization/torchao.py
+++ b/src/pruna/algorithms/quantization/torchao.py
@@ -22,6 +22,7 @@ from pruna.config.smash_config import SmashConfigPrefixWrapper
 from pruna.config.smash_space import CategoricalHyperparameter
 from pruna.engine.model_checks import get_diffusers_transformer_models, is_causal_lm
 from pruna.engine.save import SAVE_FUNCTIONS
+from pruna.logging.logger import pruna_logger
 
 # Based on common diffusers transformer architectures
 NORM_MODULES: list[str] = [
@@ -80,7 +81,10 @@ class TorchaoQuantizer(PrunaQuantizer):
     run_on_cpu: bool = True
     run_on_cuda: bool = True
     dataset_required = False
-    compatible_algorithms = dict(compiler=["torch_compile"])
+    compatible_algorithms = dict(
+        cacher=["fora"],
+        compiler=["torch_compile"],
+    )
 
     def get_hyperparameters(self) -> list:
         """
@@ -180,6 +184,26 @@ class TorchaoQuantizer(PrunaQuantizer):
             if not is_linear(module, fqn):
                 return False
             return all(name not in excluded_modules for name in fqn.split("."))
+
+        if (smash_config["compiler"] == "torch_compile" and
+            smash_config._base_config["torch_compile_mode"] != "max-autotune-no-cudagraphs"):
+            pruna_logger.warning(
+                "You are using torchao with torch.compile. "
+                "Please set `smash_config['torch_compile_mode']='max-autotune-no-cudagraphs'` for best results; "
+                "otherwise you may encounter undesirable outcomes."
+            )
+
+        if ("fp8" in smash_config["quant_type"] and
+            not (torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9))):
+            pruna_logger.warning(
+                "Float8 quantization requires an NVIDIA GPU with compute capability ≥ 8.9. "
+                "Your device does not meet this requirement."
+            )
+
+        if smash_config["quant_type"] == "fp8dqrow":
+            pruna_logger.warning(
+                "Row wise float8 dynamic quantization is still experimental and might not work on your hardware."
+            )
 
         imported_modules["quantize"](working_model, imported_modules[smash_config["quant_type"]], filter_fn=filter_fn)
         return model
--- a/src/pruna/algorithms/quantization/torchao.py
+++ b/src/pruna/algorithms/quantization/torchao.py
@@ -150,6 +150,8 @@ class TorchaoQuantizer(PrunaQuantizer):
             return True
         if isinstance(model, tuple(unet_models)):
             return True
+        if hasattr(model, "unet") and isinstance(model.unet, tuple(unet_models)):
+            return True
         if hasattr(model, "transformer") and isinstance(model.transformer, tuple(transformer_models)):
             return True
         if is_causal_lm(model):
@@ -172,7 +174,12 @@ class TorchaoQuantizer(PrunaQuantizer):
         Any
             The quantized model.
         """
-        working_model = model.transformer if hasattr(model, "transformer") else model
+        if hasattr(model, "unet"):
+            working_model = model.unet
+        elif hasattr(model, "transformer"):
+            working_model = model.transformer
+        else:
+            working_model = model
 
         excluded_modules = []
         if "norm" in smash_config["excluded_modules"]:
--- a/src/pruna/algorithms/quantization/torchao.py
+++ b/src/pruna/algorithms/quantization/torchao.py
@@ -20,7 +20,7 @@ import torch
 from pruna.algorithms.quantization import PrunaQuantizer
 from pruna.config.smash_config import SmashConfigPrefixWrapper
 from pruna.config.smash_space import CategoricalHyperparameter
-from pruna.engine.model_checks import get_diffusers_transformer_models, is_causal_lm
+from pruna.engine.model_checks import get_diffusers_transformer_models, get_diffusers_unet_models, is_causal_lm
 from pruna.engine.save import SAVE_FUNCTIONS
 from pruna.logging.logger import pruna_logger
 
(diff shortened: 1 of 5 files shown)


src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,132 @@
...
+        -------
+        list
+            The hyperparameters.
+        """
+        return [
+            CategoricalHyperparameter(
+                "quant_type",
+                choices=["int4dq", "int4wo", "int8dq", "int8wo", "fp8wo", "fp8dq", "fp8dqrow"],
+                default_value="int8dq",
+                meta=dict(desc="Quantization type to use."),
```
sharpenb: Could we provide slightly more info on the meaning of dq, wo, row in the docs? :)

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,132 @@
...
+            The configuration for the quantization.
+
+        Returns
+        -------
+        Any
+            The quantized model.
+        """
+        working_model = model.transformer if hasattr(model, "transformer") else model
+        imported_modules = self.import_algorithm_packages()
+        imported_modules["quantize"](working_model, imported_modules[smash_config["quant_type"]])
```
sharpenb: We do not support the parameters `module_to_not_convert`. This can have a big impact for some architectures.

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,132 @@
...
+from pruna.config.smash_space import CategoricalHyperparameter
+from pruna.engine.model_checks import get_diffusers_transformer_models
+from pruna.engine.save import SAVE_FUNCTIONS
+
+
+class TorchaoQuantizer(PrunaQuantizer):
+    """
+    Implement quantization using the torchao library.
+
+    This algorithm quantizes the weights and activations of linear layers in a model.
```
sharpenb: If we could give sligthly more info about what type of quantization is used e.g. linear quantization? what is the actiavtion quantization that they used? :)

src/pruna/algorithms/quantization/torchao.py
sharpenb: We could likely have even more options here. alt text

nifleisch: Really good point, I will add it to the docs. 👍

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,132 @@
...
+        Parameters
+        ----------
+        model : Any
+            The model to check.
+
+        Returns
+        -------
+        bool
+            True if the model is suitable for torchao quantization, False otherwise.
+        """
```
sharpenb, line 147: Why is that only for trasnformers in diffusers and nn.module? In particular, it looks like the test would not accept opt125m models.

sharpenb (changes requested):
Looks very promising :) Could we add in the benchmark autoquant+ torchao, and torch compile with their default torch ao backend? It would probably help to reduce confusion between the "torch ao" options :)

src/pruna/algorithms/quantization/torchao.py
nifleisch: I agree. I’m confident that for certain linear layers, quantization barely affects overall inference time, yet dynamic quantization can substantially alter their outputs. To address this, I’ll add a blacklist for popular pipelines that users can enable - ensuring caching is applied more conservatively when needed.

src/pruna/algorithms/quantization/torchao.py
nifleisch: Good catch. Torchao quantization also supports the 🤗 transformers library. Will add support for 🤗 transformers CausalLMs.

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,217 @@
...
+    "norm2_context",
+    "norm3",
+    "norm_k",
+    "norm_q",
+    "norm_v",
+    "norm_final",
+    "norm_out",
+    "skip_norm",
+]
+EMBEDDING_MODULES: list[str] = [
```
llcnt, line 44: Just a question because I already faced the same limitation: do you know if there is a way to let the user specify this list by him/herself?
(I have also hard-coded some list of layers for the `modules_to_not_convert`, but it is painful to define such huge lists for every type of models and give less freedom for the user...)

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,217 @@
...
+
+    algorithm_name = "torchao"
+    references = {"GitHub": "https://huggingface.co/docs/diffusers/quantization/torchao"}
+    save_fn = SAVE_FUNCTIONS.save_before_apply
+    tokenizer_required = False
+    processor_required = False
+    run_on_cpu: bool = True
+    run_on_cuda: bool = True
+    dataset_required = False
+    compatible_algorithms = dict(compiler=["torch_compile"])
```
llcnt: Did you test the compatibility with other family of algo, like cachers for eg ? :)
(I am curious to see if we can enjoy a combo speed-up)

src/pruna/algorithms/quantization/torchao.py
llcnt: And what about unet-based diffusers, do you think it is compatible ? :)

src/pruna/algorithms/quantization/torchao.py
nifleisch: I agree—this would be great. With our current smash_config API, it isn’t possible to specify custom inputs, although Johanna is working on adding that capability. That said, to some degree I really like the current solution, since it makes using these filters very convenient. They should work as intended for all Diffusers transformer models as they use the same naming conventions to refer to norm and embedding layers.

src/pruna/algorithms/quantization/torchao.py
nifleisch: The combination with our Pruna Pro cachers (“auto” and “periodic”) already works. Once Fora is merged, I’ll also test if torchao works with Fora/FasterCache and PAB straight out of the box.

src/pruna/algorithms/quantization/torchao.py
nifleisch: I tested it with StableDiffusion but was not able to obtain any speedup (torchao rather slowed it down). Still maybe this could be a different story for other diffusion models. I will add the compatibility.

johnrachwan123 (changes requested):
Looks great, but I tested it and having a module list option makes the warmup smaller and the overall model faster (10% in my measurements)

src/pruna/algorithms/quantization/torchao.py
sharpenb: The interface is now different to handle module not to convert in different algos. It would be great to unify them.

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,263 @@
...
+            True if the model is suitable for torchao quantization, False otherwise.
+        """
+        transformer_models = get_diffusers_transformer_models()
+        if isinstance(model, tuple(transformer_models)):
+            return True
+        if hasattr(model, "transformer") and isinstance(model.transformer, tuple(transformer_models)):
+            return True
+        if is_causal_lm(model):
+            return True
+        return isinstance(model, torch.nn.Module)
```
sharpenb, line 160: Why does it not work for SD? It is not critical though

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,263 @@
...
+                "Your device does not meet this requirement."
+            )
+
+        if smash_config["quant_type"] == "fp8dqrow":
+            pruna_logger.warning(
+                "Row wise float8 dynamic quantization is still experimental and might not work on your hardware."
+            )
+        # Only apply quantization on module list level if torch compile is also applied at that level
+        if (
+            smash_config["compiler"] == "torch_compile"
```
sharpenb, line 223: Cna this condition be simplified? It sounds like it would go int he else only if "torch_compile_target" is module_list.

src/pruna/algorithms/quantization/torchao.py
```diff
@@ -0,0 +1,263 @@
...
+            and smash_config._base_config["torch_compile_target"] == "model"
+            or smash_config["compiler"] is None
+            or smash_config["compiler"] != "torch_compile"
+        ):
+            # Apply quantization to the entire model
+            imported_modules["quantize"](
+                working_model, imported_modules[smash_config["quant_type"]], filter_fn=filter_fn
+            )
+        else:
+            # Apply quantization to individual submodules in ModuleLists
```
sharpenb, line 231: I was curious if it needsa recursive implementation. It can be that we have module list of module list. This is a bit the same logi as finding linear layers in a model recursviely.

sharpenb (changes requested):
Looks good. I still left a couple of comments :)

src/pruna/algorithms/quantization/torchao.py
johnrachwan123: Done

src/pruna/algorithms/quantization/torchao.py
johnrachwan123: Added

src/pruna/algorithms/quantization/torchao.py
johnrachwan123: That's a good idea but I guess let's do it in a separate PR for all the places where this is done especially because all models that i've seen don't currently need it

PR #127, opened 2025-05-16
docs: remove outdated Flux tutorial and add new image generation tutorial

davidberenstein1957:
- Deleted the `flux_small.ipynb` tutorial as it is no longer relevant.
- Introduced `image_generation.ipynb` tutorial to demonstrate optimizing and evaluating image generation models using the `pruna` package, including details on optimization algorithms and evaluation metrics.
- Updated the tutorial index to reflect these changes.

## Related Issue

Fixes #(issue number)

## Type of Change

- [X] This change requires a documentation update

## Checklist


davidberenstein1957:
@nifleisch, that looks good! I would leave the printed statements in the tutorials so people can actually read the output and understand better what is happening.

nifleisch:
> @nifleisch, that looks good! I would leave the printed statements in the tutorials so people can actually read the output and understand better what is happening.

I see your point. This way, users could follow the tutorial without running it on their own machines. I just spoke with Johanna, and we’re now clearing the outputs of all tutorial notebooks by default. So even if outputs remain in the tutorial, they won’t be included in the documentation.

davidberenstein1957:
That's good to know. Let's remove the output clearing, then. I think the output helps understanding, so it is fine to keep them, but feel free to verify with @johannaSommer since we initially decided to remove them.

nifleisch:
> That's good to know. Let's remove the output clearing, then. I think the output helps understanding, so it is fine to keep them, but feel free to verify with @johannaSommer since we initially decided to remove them.

Good point. I just checked with @johannaSommer, and she’s on board with this as well. Will add them back to the tutorial.

docs/tutorials/index.rst
```diff
@@ -7,6 +7,12 @@ This tutorial will guide you through the process of using |pruna| to optimize yo
 
 .. grid:: 1 2 2 2
 
+   .. grid-item-card:: Optimize and Evaluate Image Generation Models
```
sharpenb: "Compress and Evaluate" would be better

sharpenb (changes requested):
THanks for the tutorial. Here are my comments on the ntoebook :)
- I like the idea of a tutorial card at the top but it could be made cleaner (e.g. in a table?). I would also not mention the distinciton between metrics. The term "base/statefull" is more a implementation relevant rather than a user relevant. I would also remove the libraries from there.
- Do we need to show how ot install pruna? WE do not do this in the other tutorial and have a page for this.
- We could merge the set device cell with theload model cell
- Could we have "1., 2. 3." in subtitles like in other tutorials?
- We could mention how long is the smashing for this smash config so that peopl know what to expect.
- Define all metrics in a single list.
- `first_results` - > `base_model_results` (and `smashed_model_results`.
- Could we print all metrics in a nice way in a single cell?  i.e. merge cell 9. 10. 11.?
- Is this tutorial only with SD2.1? Make sense since it is quick to run but also unclear since a bit outdated.

sharpenb (changes requested):
Thanks for the udpate! Here are still some left comments:
- Let's remove the library row from the table
- Do we need to show how ot install pruna? We do not do this in the other tutorialx and have a doc page for it. I would suggest o redicrect to the installation page if people need.
- We could merge the "set device" cell with the "load model" cell
- Let's have numbers for sections like in other tutorials. Could we have "1., 2. 3." in subtitles like in other tutorials? :)
- I feel that we could find a nicer way to print the metrics e.g. the python print a small table with metrics before/after and relative difference in 3 columns.
- I feel that a small intro at the start of the tutorial to explain what we try to do and that it can work with other model than SD2.1 would be nice to give context
- We should also update the title of the notebook itself for "Compress and evaluate"

davidberenstein1957:
Hi @sharpenb, I think I once typed a response already but did not post it. Underneath are some of my reasons for the choices.

> Thanks for the udpate! Here are still some left comments:
> 
> * Let's remove the library row from the table

sure 

> * Do we need to show how ot install pruna? We do not do this in the other tutorialx and have a doc page for it. I would suggest o redicrect to the installation page if people need.

Yes, but if people want to run this as a standalone notebook, let's say in Jupyter or Google Colab or whatever other IDE, we need to install Pruna. 

> * We could merge the "set device" cell with the "load model" cell

I kept this separate as it is a bit more explicit and the device is used in multiple places throughout the notebook

> * Let's have numbers for sections like in other tutorials. Could we have "1., 2. 3." in subtitles like in other tutorials? :)

I left this out because I don't feel it adds value, and it makes it more challenging to maintain as it is a manual process.

> * I feel that we could find a nicer way to print the metrics, e.g. the Python print a small table with metrics before/after and relative difference in 3 columns.

I agree it could be nicer, but do you feel this is worth spending 30 minutes on (re-running the notebook evals and such)?

> * I feel that a small intro at the start of the tutorial to explain what we try to do and that it can work with other models than SD2.1 would be nice to give context

If you feel it is important to use a more modern model, I am happy to set that up, but I think it makes it
 less approachable to run end-to-end. Perhaps we can use stabilityai/stable-diffusion-3.5-medium or Efficient-Large-Model/SANA1.5_4.8B_1024px_diffusers?

> * We should also update the title of the notebook itself for "Compress and evaluate"

I think it is better to keep it more general, although the specific techniques we apply focus on compression.

davidberenstein1957:
@nifleisch @sharpenb @SaboniAmine could you review this PR, I think it would be nice to merge before the release.

sharpenb (changes requested):
Thanks for the update! A last update before merging would be to indicate the units with `througput (samples/s)`, `latency (s/samples)` and `Base Model`, `Compressed Model`, `Relative Difference` in the row/columns names to be complete :)

davidberenstein1957:
hi @sharpenb I resolved your comments.

nifleisch (approved):
I love the new tutorial, it feels very complete. Through it I learned about the amazing segmind/Segmind-Vega model that we probably should use more often for tutorials as it looks stunning!
Two minor things:
- At the beginning of the tutorial in section 1 there is a typo "howwever"
- In the last part you use both the throughput and the latency metric, which felt a bit redundant to me because throuput ~ 1 / latency. Instead I would have preferred a fidelity measure. (feel free to ignore this suggestion)

PR #128, opened 2025-05-16
feat: `accelerate` support

johannaSommer:
## Description
This PR introduces the option to smash a distributed model. It does so by
- extending expected SmashConfig devices to "accelerate"
- extending the `move_to_device` function
- adding functionality to determine the device of a model (plus tests)
- proof of concept adjustment for two methods to adapt to distributed models
- extending the loading of a pruna model to restore the original state

I would appreciate any suggestions on extending the tests!! I feel like the device casting functionality should be monitored thoroughly so any suggestions are welcome.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Added tests for device casting functionality which pass locally.

## Checklist

## Additional Notes
Examples:
```python
import torch
from pruna import smash, SmashConfig
from pruna.engine.utils import get_device

pipe = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", torch_dtype=torch.bfloat16, device_map="balanced")

config = SmashConfig(device="accelerate")
config["cacher"] = "deepcache"
smashed_pipe = smash(pipe, config)
get_device(smashed_pipe)
```
```bash
INFO - Starting cacher deepcache...
INFO - cacher deepcache was applied successfully.
'accelerate'
```
---------------------------
```python
import torch
from pruna import smash, SmashConfig
from pruna.engine.utils import get_device

pipe = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", torch_dtype=torch.bfloat16, device_map="balanced")

config = SmashConfig(device="cuda")
config["cacher"] = "deepcache"
smashed_pipe = smash(pipe, config)
get_device(smashed_pipe)
```
```bash
WARNING - SmashConfig specifies cuda but model is distributed.Casting model to cuda. If this is not desired, please use SmashConfig(device='accelerate') to continue with distributed model.
INFO - Starting cacher deepcache...
INFO - cacher deepcache was applied successfully.
'cuda'
```
---------------------------
```python
import torch
from pruna import smash, SmashConfig
from pruna.engine.utils import get_device

pipe = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", torch_dtype=torch.bfloat16, device_map="balanced")

config = SmashConfig(device="accelerate")
config["quantizer"] = "diffusers_int8"
smashed_pipe = smash(pipe, config)
get_device(smashed_pipe)
```
```bash
INFO - Starting quantizer diffusers_int8...
INFO - quantizer diffusers_int8 was applied successfully.
'accelerate'
```
---------------------------
```python
from transformers import AutoModelForCausalLM, AutoTokenizer
from pruna import smash, SmashConfig
from pruna.engine.utils import get_device

model_id = "meta-llama/Llama-3.2-1B"
base_model = AutoModelForCausalLM.from_pretrained(model_id, device_map="balanced")

config = SmashConfig(device="accelerate")
config["quantizer"] = "llm_int8"
smashed_model = smash(base_model, config)
get_device(smashed_model)
```
```bash
INFO - Starting quantizer llm_int8...
INFO - quantizer llm_int8 was applied successfully.
'accelerate'
```
---------------------------

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ wget = "*"
 torch = "2.7.0"
 torchvision = "0.22.0"
 torchmetrics = { version = "*", extras = ["image"] }
+accelerate = ">0.28.0"
 requests = ">=2.31.0"
 transformers = "*"
 pillow = ">=9.5.0"
--- a/src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
+++ b/src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
@@ -26,7 +26,7 @@ from pruna.engine.model_checks import (
     get_diffusers_transformer_models,
     get_diffusers_unet_models,
 )
-from pruna.engine.utils import safe_memory_cleanup
+from pruna.engine.utils import determine_dtype, move_to_device
 
 
 class DiffusersInt8Quantizer(PrunaQuantizer):
@@ -122,25 +122,29 @@ class DiffusersInt8Quantizer(PrunaQuantizer):
             The quantized model.
         """
         with tempfile.TemporaryDirectory(prefix=smash_config["cache_dir"]) as temp_dir:
-            # cast original model to CPU to free memory for smashed model
-            if hasattr(model, "to"):
-                input_device = model.device
-                model.to("cpu")
-                safe_memory_cleanup()
-
             # save the latent model (to be quantized) in a temp directory
             if hasattr(model, "transformer"):
-                model.transformer.save_pretrained(temp_dir)
-                latent_class = getattr(diffusers, type(model.transformer).__name__)
-                compute_dtype = next(iter(model.transformer.parameters())).dtype
+                working_model = model.transformer
+                device_map = {
+                    "": smash_config.device
+                    if smash_config.device != "accelerate"
+                    else smash_config.device_map["transformer"]
+                }
             elif hasattr(model, "unet"):
-                model.unet.save_pretrained(temp_dir)
-                latent_class = getattr(diffusers, type(model.unet).__name__)
-                compute_dtype = next(iter(model.unet.parameters())).dtype
+                working_model = model.unet
+                device_map = {
+                    "": smash_config.device if smash_config.device != "accelerate" else smash_config.device_map["unet"]
+                }
             else:
-                model.save_pretrained(temp_dir)
-                latent_class = getattr(diffusers, type(model).__name__)
-                compute_dtype = next(iter(model.parameters())).dtype
+                working_model = model
+                device_map = (
+                    {"": smash_config.device} if smash_config.device != "accelerate" else smash_config.device_map
+                )
+
+            move_to_device(working_model, "cpu")
+            working_model.save_pretrained(temp_dir)
+            latent_class = getattr(diffusers, type(working_model).__name__)
+            compute_dtype = determine_dtype(working_model)
 
             bnb_config = DiffusersBitsAndBytesConfig(
                 load_in_8bit=smash_config["weight_bits"] == 8,
@@ -159,6 +163,7 @@ class DiffusersInt8Quantizer(PrunaQuantizer):
                 temp_dir,
                 quantization_config=bnb_config,
                 torch_dtype=compute_dtype,
+                device_map=device_map,
             )
             # replace the latent model in the pipeline
             if hasattr(model, "transformer"):
@@ -167,10 +172,6 @@ class DiffusersInt8Quantizer(PrunaQuantizer):
                 model.unet = smashed_latent
             else:
                 model = smashed_latent
-
-            # move the model back to the original device
-            if hasattr(model, "to"):
-                model.to(input_device)
             return model
 
     def import_algorithm_packages(self) -> Dict[str, Any]:
--- a/src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
+++ b/src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
@@ -39,17 +39,14 @@ class DiffusersInt8Quantizer(PrunaQuantizer):
     This algorithm is specifically adapted for diffusers models.
     """
 
-    algorithm_name = "diffusers_int8"
-    references = {"GitHub": "https://github.com/bitsandbytes-foundation/bitsandbytes"}
-    tokenizer_required = False
-    processor_required = False
-    dataset_required = False
-    run_on_cpu = False
-    run_on_cuda = True
-    compatible_algorithms = dict(
-        factorizer=["qkv_diffusers"],
-        cacher=["deepcache", "fastercache", "fora", "pab"],
-        compiler=["torch_compile"],
+    algorithm_name: str = "diffusers_int8"
+    references: dict[str, str] = {"GitHub": "https://github.com/bitsandbytes-foundation/bitsandbytes"}
+    tokenizer_required: bool = False
+    processor_required: bool = False
+    dataset_required: bool = False
+    runs_on: list[str] = ["cuda", "accelerate"]
+    compatible_algorithms: dict[str, list[str]] = dict(
+        factorizer=["qkv_diffusers"], cacher=["deepcache", "fastercache", "fora", "pab"], compiler=["torch_compile"]
     )
 
     def get_hyperparameters(self) -> list:
--- a/src/pruna/algorithms/quantization/huggingface_llm_int8.py
+++ b/src/pruna/algorithms/quantization/huggingface_llm_int8.py
@@ -127,9 +127,8 @@ class LLMInt8Quantizer(PrunaQuantizer):
                 quantization_config=bnb_config,
                 trust_remote_code=True,
                 torch_dtype=smash_config["compute_dtype"],  # storage type of the non-int8 params
-                device_map=smash_config.device_map if smash_config.device == "accelerate" else "auto",
+                device_map=smash_config.device_map if smash_config.device == "accelerate" else smash_config.device,
             )
-            move_to_device(smashed_model, smash_config.device)
 
         return smashed_model
 
--- a/src/pruna/algorithms/quantization/huggingface_llm_int8.py
+++ b/src/pruna/algorithms/quantization/huggingface_llm_int8.py
@@ -35,14 +35,13 @@ class LLMInt8Quantizer(PrunaQuantizer):
     while 4-bit quantization further compresses the model and is often used with QLoRA for fine-tuning.
     """
 
-    algorithm_name = "llm_int8"
-    references = {"GitHub": "https://github.com/bitsandbytes-foundation/bitsandbytes"}
-    tokenizer_required = False
-    processor_required = False
-    run_on_cpu = False
-    run_on_cuda = True
-    dataset_required = False
-    compatible_algorithms = dict(compiler=["torch_compile"])
+    algorithm_name: str = "llm_int8"
+    references: dict[str, str] = {"GitHub": "https://github.com/bitsandbytes-foundation/bitsandbytes"}
+    tokenizer_required: bool = False
+    processor_required: bool = False
+    runs_on: list[str] = ["cuda", "accelerate"]
+    dataset_required: bool = False
+    compatible_algorithms: dict[str, list[str]] = dict(compiler=["torch_compile"])
 
     def get_hyperparameters(self) -> list:
         """
--- a/src/pruna/algorithms/quantization/huggingface_llm_int8.py
+++ b/src/pruna/algorithms/quantization/huggingface_llm_int8.py
@@ -23,7 +23,7 @@ from pruna.algorithms.quantization import PrunaQuantizer
 from pruna.config.smash_config import SmashConfigPrefixWrapper
 from pruna.config.smash_space import Boolean
 from pruna.engine.model_checks import is_causal_lm
-from pruna.engine.utils import safe_memory_cleanup
+from pruna.engine.utils import move_to_device
 
 
 class LLMInt8Quantizer(PrunaQuantizer):
@@ -107,10 +107,9 @@ class LLMInt8Quantizer(PrunaQuantizer):
         """
         with tempfile.TemporaryDirectory(prefix=smash_config["cache_dir"]) as temp_dir:
             # cast original model to CPU to free memory for smashed model
-            if hasattr(model, "to"):
-                model.to("cpu")
-                safe_memory_cleanup()
+            move_to_device(model, "cpu")
             model.save_pretrained(temp_dir)
+
             bnb_config = BitsAndBytesConfig(
                 load_in_8bit=smash_config["weight_bits"] == 8,
                 load_in_4bit=smash_config["weight_bits"] == 4,
@@ -128,7 +127,9 @@ class LLMInt8Quantizer(PrunaQuantizer):
                 quantization_config=bnb_config,
                 trust_remote_code=True,
                 torch_dtype=smash_config["compute_dtype"],  # storage type of the non-int8 params
+                device_map=smash_config.device_map if smash_config.device == "accelerate" else "auto",
             )
+            move_to_device(smashed_model, smash_config.device)
 
         return smashed_model
 
--- a/src/pruna/config/compatibility_checks.py
+++ b/src/pruna/config/compatibility_checks.py
@@ -94,9 +94,9 @@ def check_model_compatibility(
     for current_group in ALGORITHM_GROUPS:
         algorithm = smash_config[current_group]
         if algorithm is not None:
+            check_algorithm_availability(algorithm, current_group, algorithm_dict)
             # test if all required packages are installed, if not this will raise an ImportError
             algorithm_dict[current_group][algorithm].import_algorithm_packages()
-            check_algorithm_availability(algorithm, current_group, algorithm_dict)
             check_argument_compatibility(smash_config, algorithm)
             # check for model-algorithm compatibility with the model_check_fn
             if not algorithm_dict[current_group][algorithm].model_check_fn(model):
--- a/src/pruna/config/compatibility_checks.py
+++ b/src/pruna/config/compatibility_checks.py
@@ -1,0 +1,158 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+from typing import Any
+
+from pruna import SmashConfig
+from pruna.algorithms import PRUNA_ALGORITHMS
+from pruna.config.smash_space import SMASH_SPACE
+from pruna.engine.utils import get_device, move_to_device
+from pruna.logging.logger import pruna_logger
+
+
+def ensure_device_consistency(model, smash_config):
+    """
+    Ensure consistency between the device state of the model and the smash config.
+
+    Parameters
+    ----------
+    model : Any
+        The model to check for device consistency.
+    smash_config : SmashConfig
+        The smash config to check for device consistency.
+    """
+    model_device = get_device(model)
+
+    if smash_config.device in ["cpu", "cuda", "mps"]:
+        if model_device == smash_config.device:
+            pruna_logger.debug("Device consistency check passed.")
+        else:
+            if model_device != "accelerate":
+                pruna_logger.warning(
+                    (
+                        f"Model and SmashConfig have different devices. Model: {model_device}, "
+                        f"SmashConfig: {smash_config.device}. Casting model to {smash_config.device}."
+                        f"If this is not desired, please use SmashConfig(device='{model_device}')."
+                    )
+                )
+            else:
+                # in this case, the model_device is a device map and the model is on multiple GPUs
+                pruna_logger.warning(
+                    (
+                        f"SmashConfig specifies {smash_config.device} but model is distributed. "
+                        f"Casting model to {smash_config.device}. If this is not desired, please use "
+                        f"SmashConfig(device='accelerate') to continue with distributed model."
+                    )
+                )
+            move_to_device(model, smash_config.device)
+
+    elif smash_config.device == "accelerate":
+        if model_device == "accelerate":
+            pruna_logger.debug("Device consistency check passed.")
+            hf_device_map = get_device(model, return_device_map=True)
+            if not all(isinstance(v, int) for v in hf_device_map.values()):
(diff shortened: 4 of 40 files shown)


tests/engine/test_device.py
```diff
@@ -39,3 +40,35 @@ def test_device_available(device: str | torch.device, expected: str) -> None:
...
+    ],
+)
+def test_accelerate_to_single_casting(target_device: str | torch.device) -> None:
+    """Test that the device can be cast to the target device."""
+    model = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", device_map="balanced")
+    device_map = model.hf_device_map
+    move_to_device(model, target_device)
+    assert get_device(model) == target_device 
+    move_to_device(model, "accelerate", device_map=device_map)
+    assert get_device(model) == "accelerate"
```
guennemann: Do we want to expand the tests to actually do some computations in a distributed manner? And not just the move_to_device?

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -122,25 +122,29 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                    "": smash_config.device if smash_config.device != "accelerate" else smash_config.device_map["unet"]
+                }
             else:
-                model.save_pretrained(temp_dir)
-                latent_class = getattr(diffusers, type(model).__name__)
-                compute_dtype = next(iter(model.parameters())).dtype
+                working_model = model
+                device_map = (
+                    {"": smash_config.device} if smash_config.device != "accelerate" else smash_config.device_map
+                )
```
guennemann, line 139: The brackets () are done on purpose?

src/pruna/algorithms/quantization/huggingface_llm_int8.py
```diff
@@ -128,7 +127,9 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
                 quantization_config=bnb_config,
                 trust_remote_code=True,
                 torch_dtype=smash_config["compute_dtype"],  # storage type of the non-int8 params
+                device_map=smash_config.device_map if smash_config.device == "accelerate" else "auto",
```
guennemann: Why auto here? Isn't auto still distributing?

src/pruna/smash.py
```diff
@@ -71,6 +77,61 @@ def smash(
...
+    smash_config : SmashConfig
+        The smash config to check for device consistency.
+    """
+    model_device = get_device(model)
+
+    if smash_config.device in ["cpu", "cuda", "mps"]:
+        if model_device == smash_config.device:
+            pruna_logger.debug("Device consistency check passed.")
+        else:
+            if model_device != smash_config.device and model_device != "accelerate":
```
guennemann: `model_device != smash_config.device` should always hold in the else branch. please remove for clarity.

src/pruna/smash.py
```diff
@@ -71,6 +77,61 @@ def smash(
...
+                    (
+                        f"Model and SmashConfig have different devices. Model: {model_device}, "
+                        f"SmashConfig: {smash_config.device}. Casting model to {smash_config.device}."
+                    )
+                )
+            else:
+                # in this case, the model_device is a device map and the model is on multiple GPUs
+                pruna_logger.warning(
+                    (
+                        f"SmashConfig specifies {smash_config.device} but model is distributed."
```
guennemann: Why is the model distributed? Can't it be that smash_config is cuda and model is cpu?

src/pruna/smash.py
```diff
@@ -71,6 +77,61 @@ def smash(
...
+
+    if smash_config.device in ["cpu", "cuda", "mps"]:
+        if model_device == smash_config.device:
+            pruna_logger.debug("Device consistency check passed.")
+        else:
+            if model_device != smash_config.device and model_device != "accelerate":
+                pruna_logger.warning(
+                    (
+                        f"Model and SmashConfig have different devices. Model: {model_device}, "
+                        f"SmashConfig: {smash_config.device}. Casting model to {smash_config.device}."
```
guennemann: Shouldn't one also say here "If this is not desired...."

src/pruna/smash.py
```diff
@@ -71,6 +77,61 @@ def smash(
...
+            hf_device_map = get_device(model, return_device_map=True)
+            if not all(isinstance(v, int) for v in hf_device_map.values()):
+                raise ValueError("Device map indicates CPU offloading, this is not supported at this time.")
+            else:
+                smash_config.device_map = hf_device_map
+        else:
+            pruna_logger.warning(
+                (
+                    f"SmashConfig specifies 'accelerate' but model is not distributed and is on device {model_device}. "
+                    f"Updating SmashConfig to device='{model_device}'."
```
guennemann: Why do we override here the smashconfig but above we use the smash config when it is not the same?

src/pruna/engine/utils.py
```diff
@@ -95,10 +96,34 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
...
+        raise ValueError("Device must be a string in [cpu, cuda, mps, accelerate].")
+
+    # do not cast if the model is already on the correct device
+    if get_device(model) == device:
         return
-    if hasattr(model, "to"):
+
+    # logic for transformers pipelines, which have a device attribute but do not support casting directly
+    if hasattr(model, "task") and getattr(model, "task") == "automatic-speech-recognition":
+        target_model = model.model
```
guennemann: why here only for ASR?

guennemann (changes requested):
Very nice to get multi-GPU support. See my comments inline.

tests/engine/test_device.py
johannaSommer: Yes completely agree - my argument would be that this function specifically only tests the device casting function in an isolated manner (as this can already have a bunch of issues by itself). In a follow up PR, I am adding accelerate-models to the Algorithm Tests and with Begüm's extension of the tests, in particular by integrating the inference handler, we would actually do that for all smashed models. Do you think that would be enough or should we still test it here for completeness?

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
johannaSommer: yes that is necessary here because we enforce a certain line length, for which this expression is too long. With these brackets you can have expressions across multiples lines 👍🏻

tests/engine/test_device.py
guennemann: If it is covered in the inference handler, it should be fine.

src/pruna/algorithms/quantization/huggingface_llm_int8.py
johannaSommer: Great catch, thanks a lot! I mixed up the diffusers interface (it's complicated, you can't just specify `device_map="auto"` which is why we have to construct these device maps in the other method `diffusers_int8`) but transformers does actually allow `device_map="cuda".` Fixed this and was able to remove the device casting later in the function as a result of this. 😊

src/pruna/smash.py
johannaSommer: 🫡

src/pruna/smash.py
johannaSommer: Yes this can be the case, but from what I understand this is handled in the above statement. We first check on a mismatch of model device and smash config device:
- if the model device is not accelerate, its a cpu/cuda mismatch and we cast the model correctly
- if the model is on accelerate, we give a different warning

src/pruna/smash.py
johannaSommer: 🫡

src/pruna/smash.py
johannaSommer: So in this case the SmashConfig expects a distributed model but the input is not. Since there are many ways to distribute a model (which module goes where for example), I felt it would be best to just adjust the SmashConfig. In the above scenario its a bit easier because I can just recast (there is only one way on how to cast to CPU).

If you feel like we should somehow still try to _distribute a model for the user_ I think I can find a way but we would be making the decision on how to distribute for the user. What do you think would be best here?

src/pruna/engine/utils.py
johannaSommer: We had this already in the earlier version of the device movement function, it is only ASR at the moment as this is the only pipeline we really support at the moment. But you're right, let's take the time now to make this more general. Is adjusted 👍🏻

src/pruna/smash.py
guennemann: Your are right. I got confused with the different branches here.

src/pruna/smash.py
guennemann: I just feel that it might be confusing that sometimes we use one, sometimes the other. The warning is clear about it, but it is still a bit confusing. Would have been easier to say "if they disagree, then we always use X".

guennemann (approved):
Looks good to me.

Probably re-think/decide on the "overwrite" rule when the smash and model device disagree.

src/pruna/smash.py
```diff
@@ -71,6 +77,62 @@ def smash(
     return smashed_model
 
 
+def ensure_device_consistency(model, smash_config):
```
johnrachwan123: can we move this and other utility methods to a utils script ?

tests/engine/test_device.py
```diff
@@ -39,3 +40,35 @@ def test_device_available(device: str | torch.device, expected: str) -> None:
...
+@pytest.mark.parametrize(
+    "target_device",
+    [
+        "cuda",
+        "cpu",
+    ],
+)
+def test_accelerate_to_single_casting(target_device: str | torch.device) -> None:
+    """Test that the device can be cast to the target device."""
+    model = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", device_map="balanced")
```
johnrachwan123: can we change this to tiny sd ? and maybe add other tiny ones since its free ?

johnrachwan123 (approved):
LGTM! Thanks for the great work 🚀

tests/engine/test_device.py
johannaSommer: Not possible, I tried but in this case the tiny sd is really too tiny to be distributed 🥹

src/pruna/smash.py
johannaSommer: done 👍🏻

johnrachwan123 (approved):
Great work on this PR, I'm very excited for what's comming next 👀

pyproject.toml
```diff
@@ -43,6 +43,7 @@ wget = "*"
 torch = "2.7.0"
 torchvision = "0.22.0"
 torchmetrics = { version = "*", extras = ["image"] }
+accelerate = ">0.28.0"
```
begumcig, line 51: IT'S HAPPENING

src/pruna/engine/utils.py
```diff
@@ -110,32 +133,100 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
                 raise ValueError(f"Could not move model to device: {str(e)}")
             else:
                 pruna_logger.warning(f"Could not move model to device: {str(e)}")
-    elif hasattr(model, "task") and getattr(model, "task") == "automatic-speech-recognition":
-        model.model.to(device)
+    safe_memory_cleanup()
+
+
+def remove_all_accelerate_hooks(model: Any) -> None:
```
begumcig, line 166: dinner is SERVED

src/pruna/engine/utils.py
```diff
@@ -110,32 +133,100 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
...
+        # transformers models are all torch.nn.Module, which is what the hook removal expects
+        remove_hook_from_module(model, recurse=True)
     else:
-        if raise_error:
-            raise ValueError("Model does not support device movement.")
-        else:
-            pruna_logger.warning("Model does not support device movement.")
+        # diffusers pipelines e.g. are not torch modules, so we need to find all attributes that are modules
+        # we only do this at the first level, recurse will take care of the rest
+        for attr in model.components:
```
begumcig, line 195: do we know for sure that everything that's not an nn.Module is going to have a components argument?

tests/engine/test_device.py
```diff
@@ -39,3 +42,127 @@ def test_device_available(device: str | torch.device, expected: str) -> None:
...
+    move_to_device(model, input_device)
+    assert get_device(model) == input_device
+    move_to_device(model, target_device)
+    assert get_device(model) == target_device
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("target_device", ["cuda", "cpu"])
+@pytest.mark.parametrize("model_fixture", ["stable_diffusion_v1_4"], indirect=True)
+def test_accelerate_diffusers_casting(target_device: str | torch.device, model_fixture: Any) -> None:
```
begumcig, line 71: These tests are really valuable, thanks for covering so many scenarios! From what I understand, you're verifying that the model can be distributed across multiple GPUs, whether it's initially on CPU or already on a single GPU, right?
Would it also make sense to test the reverse, moving the model back to a single GPU or CPU after it's been spread across multiple devices?

tests/fixtures.py
```diff
@@ -93,15 +93,15 @@ def dummy_model() -> tuple[Any, SmashConfig]:
 
 def get_diffusers_model(cls: type[Any], model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, SmashConfig]:
     """Get a diffusers model for image generation."""
-    model = cls.from_pretrained(model_id, **kwargs)
+    model = cls.from_pretrained(model_id, device_map="balanced", **kwargs)
```
begumcig: Asking because I am curious, I was under the impression that balanced already moves a model to multi GPUs, why are we adding it here in the fixtures?

src/pruna/smash.py
```diff
@@ -49,6 +56,9 @@ def smash(
         Smashed model wrapped in a `PrunaModel` object.
     """
     with PrunaLoggerContext(verbose=verbose):
+        # check the device consistency of the model and the smash config
+        ensure_device_consistency(model, smash_config)
```
begumcig, line 60: slay

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -159,6 +163,7 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
                 temp_dir,
                 quantization_config=bnb_config,
                 torch_dtype=compute_dtype,
+                device_map=device_map,
```
begumcig, line 163: Oooh does this mean that rather than moving the model to the required device after applying the quantization (line 170 before) we now handle this in the quantization? Wow

src/pruna/config/compatibility_checks.py
```diff
@@ -0,0 +1,156 @@
...
+    ----------
+    model : Any
+        The model to check for device consistency.
+    smash_config : SmashConfig
+        The smash config to check for device consistency.
+    """
+    model_device = get_device(model)
+
+    if smash_config.device in ["cpu", "cuda", "mps"]:
+        if model_device == smash_config.device:
```
begumcig: THE BIGGEST SLAY OF ALL ETERNITY

src/pruna/config/compatibility_checks.py
```diff
@@ -0,0 +1,156 @@
...
+            pruna_logger.debug("Device consistency check passed.")
+        else:
+            if model_device != "accelerate":
```
begumcig: why do we not check the accelerate condition above (line 39)? Because smashing on multiple devices is not something we support yet completely?

src/pruna/config/compatibility_checks.py
```diff
@@ -0,0 +1,156 @@
...
+                pruna_logger.warning(
+                    (
+                        f"SmashConfig specifies {smash_config.device} but model is distributed. "
+                        f"Casting model to {smash_config.device}. If this is not desired, please use "
+                        f"SmashConfig(device='accelerate') to continue with distributed model."
+                    )
+                )
+            move_to_device(model, smash_config.device)
+
+    elif smash_config.device == "accelerate":
```
begumcig: sorry the check was here! forget my above comment!

src/pruna/config/compatibility_checks.py
```diff
@@ -0,0 +1,156 @@
...
+            else:
+                smash_config.device_map = hf_device_map
+        else:
+            pruna_logger.warning(
+                (
+                    f"SmashConfig specifies 'accelerate' but model is not distributed and is on device {model_device}. "
+                    f"Updating SmashConfig to device='{model_device}'."
+                )
+            )
+            smash_config.device = model_device
```
begumcig: So above,  if the model is already distributed across multiple devices, but the `smash_config.device` is not set to "accelerate", then we move the model to the device specified in `smash_config.device` (this happens at line 60).

However, here in the opposite case: when the model is not distributed `but smash_config.device` is "accelerate", we update `smash_config.device` to match the model's actual device.

Do you think this inconsistency could lead to issues?
In particular, in the first scenario, if the model is too large to fit on a single device, forcing it onto smash_config.device might fail. Would it make more sense to also update smash_config.device to "accelerate" in that case, instead of moving the model?

begumcig (changes requested):
An absolute slay-job, really impressive and thoughtfully done! I mostly left comments out of admiration, just a few small notes to better understand certain parts and help with consistency. Seriously, fantastic job 💜💜💜

src/pruna/engine/utils.py
johannaSommer: yeah good catch, all pipelines I tested should but I turned the else into an elif just to make sure

tests/engine/test_device.py
johannaSommer: Yes you got it! I know what you mean, it isn't exactly obvious here but this already happens at the first `move_to_device`. In the distributed scenario the model comes in distributed (we need this in order to obtain the device map). So this is actually already part of the test!

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
johannaSommer: yessss 🫡 and I will follow up with more algorithm adjustments on this, this is just for proof of concept

src/pruna/config/compatibility_checks.py
johannaSommer: yeah good point - we definitely support it, it is more that I think we can't cast the model for the user. There are very many ways to distribute a model (e.g. 50/50 on both GPUs or 20% on CUDA:0 and 80% on CUDA:1) so in this case we can't really do that for the user

src/pruna/config/compatibility_checks.py
johannaSommer: Updated! Stephan had the same comment, I was slightly conflicted about it but I will adjust it after I looked into it more. In a nutshell I was afraid someone might unintentionally have their model distributed (because of device_map="auto" being the default) and if we maintain the accelerated state the user will have access to less algorithms (not all will be able to support distributed). But I double checked and "auto" fills up GPU 0 first if possible, so we should be good then, i.e., you don't really accidentally distribute the model unless you have to.

johannaSommer: i refactored the overall logic to make this a bit clearer, maybe you can look at it closely one more time to see if it makes sense.

tests/fixtures.py
johannaSommer: This is to test basically the full pipeline of getting a model in a distributed way and then checking if everything that follows passes... My current strategy is to always have the fixture in a distributed way and cast it to the requested device in the test itself. I am not 100% happy with it yet also :/ Should we load it on CPU always and then put it on the device in the test itself? Might make more sense?

src/pruna/engine/pruna_model.py
```diff
@@ -102,6 +102,22 @@ def run_inference(
         outputs = self.inference_handler.process_output(outputs)
         return outputs
 
+    def compare_model_isinstance(self, instance_type: type) -> bool:
```
johnrachwan123: can we rename to isinstance

tests/engine/test_device.py
johnrachwan123, line 1: can you exclude the distributed tests from the nightly tests ?

src/pruna/config/compatibility_checks.py
```diff
@@ -0,0 +1,152 @@
...
+        # in case of accelerate, we need to store the device map
+        if model_device == "accelerate":
+            pruna_logger.debug("Device consistency check passed.")
+            hf_device_map = get_device(model, return_device_map=True)
+            if not all(isinstance(v, int) for v in hf_device_map.values()):
+                raise ValueError("Device map indicates CPU offloading, this is not supported at this time.")
+            else:
+                smash_config.device_map = hf_device_map
+
+    if smash_config.device in ["cpu", "cuda", "mps"] and model_device in ["cpu", "cuda", "mps"]:
```
begumcig: I am probably not seeing something but how does this point to an incompatibility?

src/pruna/engine/utils.py
```diff
@@ -95,10 +98,36 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
...
         Whether to raise an error when the device movement fails.
+    device_map : dict[str, str] | None
+        The device map to use if the target device is "accelerate".
     """
-    if hasattr(model, "device") and check_model_already_on_device(model, device):
+    if isinstance(model, Pipeline):
+        move_to_device(model.model, device, raise_error, device_map)
+        # this is a workaround for a flaw in the transformers pipeline handling
+        # specifically for a pipeline, the model is not expected to have a hf_device_map attribute
+        if device != "accelerate" and hasattr(model.model, "hf_device_map"):
```
begumcig, line 129: clean up slay

src/pruna/engine/utils.py
```diff
@@ -110,32 +139,109 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
...
+    ):
+        # transformers models are all torch.nn.Module, which is what the hook removal expects
+        remove_hook_from_module(model, recurse=True)
+    elif hasattr(model, "components"):
+        # diffusers pipelines e.g. are not torch modules, so we need to find all attributes that are modules
+        # we only do this at the first level, recurse will take care of the rest
+        for attr in model.components:
+            if isinstance(getattr(model, attr), torch.nn.Module):
+                remove_hook_from_module(getattr(model, attr), recurse=True)
+
```
begumcig, line 202: sorry for being super annoying should we add an else here to log some info that we couldn't remove the hooks 🥺🥺🥺

tests/engine/test_device.py
```diff
@@ -39,3 +44,123 @@ def test_device_available(device: str | torch.device, expected: str) -> None:
...
+
+
+@pytest.mark.distributed
+@pytest.mark.parametrize("target_device", ["cuda", "cpu"])
+@pytest.mark.parametrize("model_fixture", ["sd_tiny_random"], indirect=True)
+def test_accelerate_diffusers_casting(target_device: str | torch.device, model_fixture: Any) -> None:
+    """Test that a diffusers pipeline can be cast to the target device."""
+    model, _ = model_fixture
+    device_map = construct_device_map_manually(model)
+    move_and_verify(model, target_device, device_map)
```
begumcig, line 77: Just marking this based on our convo, we need to move to accelerate here :)

johannaSommer: added and ran the test again - still passing 🫡

johnrachwan123 (approved):
Let's gooooo 🚀

PR #139, opened 2025-05-20
fix: failing docs tests

davidberenstein1957:
## Description

This PR fixes the failing tests related to the typos in the code blocks.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/docs/user_manual/evaluate.rst
+++ b/docs/user_manual/evaluate.rst
@@ -74,6 +74,11 @@ Let's see what that looks like in code.
     # Create and configure EvaluationAgent
     eval_agent = EvaluationAgent(task)
 
+    # Optional: tweak model generation parameters for benchmarking
+    optimized_model.inference_handler.model_args.update(
+        {"max_new_tokens": 100}
+    )
+
     # Evaluate the model
     results = eval_agent.evaluate(optimized_model)
 
@@ -353,8 +358,6 @@ Let's see how this works in code.
 
         .. code-block:: python
 
-            import copy
-
             from diffusers import StableDiffusionPipeline
 
             from pruna import SmashConfig, smash
@@ -372,8 +375,7 @@ Let's see how this works in code.
             pipe = StableDiffusionPipeline.from_pretrained(model_path)
 
             # Smash the model
-            copy_pipe = copy.deepcopy(pipe)
-            smashed_pipe = smash(copy_pipe, smash_config)
+            smashed_pipe = smash(pipe, smash_config)
 
             # Define the task and the evaluation agent
             metrics = [CMMD()]
@@ -381,6 +383,11 @@ Let's see how this works in code.
             datamodule.limit_datasets(5)
             task = Task(metrics, datamodule=datamodule)
             eval_agent = EvaluationAgent(task)
+
+            # Optional: tweak model generation parameters for benchmarking
+            smashed_pipe.inference_handler.model_args.update(
+                {"num_inference_steps": 1, "guidance_scale": 0.0}
+            )
 
             # Evaluate base model, all models need to be wrapped in a PrunaModel before passing them to the EvaluationAgent
             first_results = eval_agent.evaluate(pipe)
@@ -417,6 +424,14 @@ Let's see how this works in code.
             datamodule.limit_datasets(5)
             task = Task(metrics, datamodule=datamodule)
             eval_agent = EvaluationAgent(task)
+
+            # wrap the model in a PrunaModel to use the EvaluationAgent
+            wrapped_pipe = PrunaModel(pipe, None)
+
+            # Optional: tweak model generation parameters for benchmarking
+            inference_arguments = {"num_inference_steps": 1, "guidance_scale": 0.0}
+            wrapped_pipe.inference_handler.model_args.update(inference_arguments)
+            wrapped_pipe.inference_handler.update_model(wrapped_pipe)
 
             # Evaluate base model, all models need to be wrapped in a PrunaModel before passing them to the EvaluationAgent
             first_results = eval_agent.evaluate(pipe)
--- a/docs/user_manual/evaluate.rst
+++ b/docs/user_manual/evaluate.rst
@@ -68,13 +68,14 @@ Let's see what that looks like in code.
         dataset_name="WikiText",
         tokenizer=optimized_model.smash_config.tokenizer,
     )
+    datamodule.limit_datasets(10)
     task = Task(request=["perplexity"], datamodule=datamodule, device="cpu")
 
     # Create and configure EvaluationAgent
     eval_agent = EvaluationAgent(task)
 
     # Evaluate the model
-    eval_agent.evaluate(optimized_model)
+    results = eval_agent.evaluate(optimized_model)
 
 Evaluation Components
 ---------------------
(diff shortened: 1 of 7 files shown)


docs/user_manual/save_load.rst
```diff
@@ -150,10 +150,11 @@ To load a previously saved ``PrunaModel``, use the ``PrunaModel.from_pretrained(
     .. tab:: Hugging Face Hub Loading
 
         .. code-block:: python
+            :class: noextract
```
johannaSommer: just a quick question, why do we do no-extract here? the from hub should work independently as far as i can see

johannaSommer (changes requested):
LGTM, thanks a lot for taking care of this! I think we should also include the API reference snippets to check, maybe we can create a seperate PR for this? I also think there are some tutorial tests that are outdated, we can probably also parametrize the `test_tutorials` in the same way.

docs/user_manual/smash.rst
```diff
@@ -29,36 +29,34 @@ Let's see what that looks like in code.
...
+    task = Task(metrics, datamodule=PrunaDataModule.from_string("LAION256"))
     eval_agent = EvaluationAgent(task)
     eval_agent.evaluate(optimized_model)
 
     # Run inference
     optimized_model.set_progress_bar_config(disable=True)
-    optimized_model.inference_handler.model_args.update(
-        {"num_inference_steps": 1, "guidance_scale": 0.0}
-    )
+    optimized_model.inference_handler.model_args.update({"num_inference_steps": 1, "guidance_scale": 0.0})
```
begumcig: I am a bit confused here, we update the inference handler args but then we don't use it to run inference. I doubt the num_inference_steps and the guidance_scale will be applied

docs/user_manual/smash.rst
```diff
@@ -118,8 +127,16 @@ To evaluate the optimized model, we can use the same interface as the original m
...
+
+    # Define metrics
     metrics = ['clip_score', 'psnr']
+
+    # Define task
     task = Task(metrics, datamodule=PrunaDataModule.from_string('LAION256'))
+
+    # Evaluate the model
     eval_agent = EvaluationAgent(task)
     eval_agent.evaluate(optimized_model)
```
begumcig: since we have MetricResult now maybe also makes sense to assign the results to a variable and print them?

docs/user_manual/smash.rst
```diff
@@ -132,11 +149,17 @@ To run inference with the optimized model, we can use the same interface as the
...
+    # Load the optimized model
+    optimized_model = PrunaModel.from_hub("PrunaAI/segmind-tiny-sd-smashed")
+
     optimized_model.set_progress_bar_config(disable=True)
     optimized_model.inference_handler.model_args.update(
         {"num_inference_steps": 1, "guidance_scale": 0.0}
     )
-    optimized_model("A serene landscape with mountains").images[0]
+    prompt = "A serene landscape with mountains"
+    optimized_model(prompt).images[0].save("output.png")
```
begumcig, line 160: Same question with above. We set inference_handler parameters but we don't use it.

docs/user_manual/smash.rst
```diff
@@ -149,59 +172,64 @@ Example 1: Diffusion Model Optimization
...
+    smash_config["compiler"] = "torch_compile"
+    smash_config["quantizer"] = "hqq_diffusers"
 
     # Optimize the model
     optimized_model = smash(model=model, smash_config=smash_config)
 
     # Generate an image
-    optimized_model("A serene landscape with mountains").images[0]
+    prompt = "A serene landscape with mountains"
+    optimized_model(prompt).images[0].save("cat.png")
```
begumcig: Same comment. Just marking these so you don't have to search for them 🥹

docs/user_manual/smash.rst
```diff
@@ -149,59 +172,64 @@ Example 1: Diffusion Model Optimization
...
     # Create and configure SmashConfig
     smash_config = SmashConfig()
-    smash_config["quantizer"] = "gptq"  # Apply GPTQ quantization
+    smash_config["compiler"] = "torch_compile"
+    smash_config["quantizer"] = "hqq"
 
     # Optimize the model
-    optimized_model = smash(model=model, smash_config=smash_config)
+    optimized_model = smash(model=pipe.model, smash_config=smash_config)
+    pipe.model = optimized_model
```
begumcig: I am a bit confused about this and I am sure there's a valid reason, can you explain why are we patching the PrunaModel on the LLama model

docs/user_manual/evaluate.rst
```diff
@@ -403,16 +413,16 @@ Let's see how this works in code.
...
             metrics = [CMMD(call_type="pairwise")]
-            task = Task(metrics, datamodule=PrunaDataModule.from_string('LAION256'))
+            datamodule = PrunaDataModule.from_string("LAION256")
+            datamodule.limit_datasets(5)
+            task = Task(metrics, datamodule=datamodule)
             eval_agent = EvaluationAgent(task)
 
             # Evaluate base model, all models need to be wrapped in a PrunaModel before passing them to the EvaluationAgent
             first_results = eval_agent.evaluate(pipe)
-            print(first_results)
```
begumcig, line 411: I am definitely fine by this, but I have noticed for the algorithms docs we have always also showed the results so I wondered, why did we opt for not printing the evaluation results?

begumcig (changes requested):
Looks really really good already, love that the models in our docs got an update! Left some smol comments :))

docs/user_manual/save_load.rst
davidberenstein1957: No reason, I removed it.

docs/user_manual/evaluate.rst
davidberenstein1957: I just decided to remove them as they don't have a direct value IMO, so I decided to remove the statements and the outcome.

docs/user_manual/smash.rst
davidberenstein1957: I am not sure when to use this exactly then. In what scenario would we use the inference handler? Is this only used internally within the EvaluationAgent

docs/user_manual/smash.rst
davidberenstein1957: I did not realise the model was update in place.

tests/documentation/test_code_blocks.py
```diff
@@ -12,13 +12,9 @@
...
     [
-        pytest.param(path.stem, marks=pytest.mark.cuda)
+        pytest.param(f"user_manual/{path.stem}", marks=pytest.mark.cuda)
         for path in TUTORIAL_PATH.glob("user_manual/*.rst")
-        if path.stem not in ["dataset"]
-    ]
-    + [
-        pytest.param(path.stem, marks=pytest.mark.cuda)
-        for path in TUTORIAL_PATH.glob("contributions/*.rst")
+        if "customize_" not in path.stem
```
johannaSommer: could we also include the customization tutorials? I remember testing them (and having them pass) for the algorithms, I think @begumcig also did the same for the metrics. For the algorithms it would help since the tests would fail if we don't implement the ABC correctly (e.g. the Pruna Base class changed and we forgot it here)

johannaSommer (approved):
Thank you!!

davidberenstein1957:
@sharpenb could you review this PR, I think it should be ready to be merged before the next release.

docs/user_manual/smash.rst
begumcig: we would use the inference handler when we call `run_inference` in PrunaModel :)

PR #143, opened 2025-05-21
reafactor: remove deprecated awq algorithm

johnrachwan123:
## Description
- remove `AWQQuantizer` implementation and references
- drop awq save/load helpers
- update handler utils
- clean up SmashConfig examples
- remove autoawq dependency
- drop awq test case

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
None.

## Checklist

## Additional Notes
There will be another PR that adds AWQ back through llm-compressor.


llcnt (commented):
Look already good to me :)
I have a general question: don't you think we should add some deprecation messages somewhere for the user(s) using awq with pruna ?
Also, what about models on hf hub that are already quantized with awq, do you think we will be able to still re-load them with `pruna` thanks to this [PR](https://github.com/PrunaAI/pruna/pull/144)?

johnrachwan123:
> Look already good to me :) I have a general question: don't you think we should add some deprecation messages somewhere for the user(s) using awq with pruna ? Also, what about models on hf hub that are already quantized with awq, do you think we will be able to still re-load them with `pruna` thanks to this [PR](https://github.com/PrunaAI/pruna/pull/144)?

I discussed the first point with Johanna and we will just name this new method awq so that the user doesn't really feel a difference :) 

Regarding the second point, I guess in our HF models, we specify what version of auto-awq you need to load the model so we don't really need to worry about that imo.

llcnt (approved):
Looks good to me:)

PR #145, opened 2025-05-21
feat: add janus support for quantization+torch.compile combo(s)

llcnt:
## Description
This goal of this PR is to decrease the memory impact and the latency of the [janus(pro-7b) model](https://huggingface.co/deepseek-community/Janus-Pro-7B).
This model are based on [Llamagen](https://arxiv.org/abs/2406.06525) and compute tokens in a latent space with an autoregressive fashion, thanks to an attribute (here called `language_model`) defined as a llama model.
There is currently no standardization regarding llamagen AR models, so this PR is exclusively dedicated to janus models (that are compatible with the `transformers` package). But we expect in the near future (see this [thread](https://github.com/huggingface/transformers/issues/33905)) that llmgenAR models will have similar `.generate()` functions.

The idea of the code change is:
- create a `JanusGenerator` (similar to the TransformersGenerator we already had, and that is renamed CausalLMGenerator);
- adapt the context manager to be able to deal with Janus models;
- adapt HQQ to be able to work on Janus;
- adapt hqq save function (distinct savings for the lm model and the rest: similar to what is done for `diffusers` pipeline, but more tricky as there is no (yet) pipeline for janus);
- adapt hqq load function (also distinct loads for the LM model and the rest).
The above points can be extended to other llm quantizer. However, testing the code and adapting all save/load functions is time consuming. I leave this work for a future PR.

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
When quantized with hqq4bits and combined with torch.compile, we can obtain a ~x3 speedup.
I have added 2 unit tests, and provided (below) a notebook for reproducing the results.
Edit: The notebook works well for torch==2.5.1. For torch==2.7, major changes has been introduced in torch dynamo, we have to slightly adapt the smash_config: smash_config['torch_compile_fullgraph'] = False
smash_config['torch_compile_mode'] = 'default'
smash_config['torch_compile_backend'] = 'inductor'.
Otherwise the compilation step takes very long, and is reapplied at each step with torch==2.7

## Checklist

## Additional Notes
- A notebook for reproducing results is provided [here](https://drive.google.com/file/d/1zweLifIM2Hn_9ODak_AYdh-B6Ds8eKlQ/view?usp=sharing);
- The current Janus models lags behind the current sota flow matching models. It is more comparable to the first versions of midjourney and dalee's, as you can see in [this blog](https://www.the-next-tech.com/review/janus-pro-7b-text-to-image/).

### DIFF
--- a/src/pruna/algorithms/compilation/utils.py
+++ b/src/pruna/algorithms/compilation/utils.py
@@ -643,9 +643,48 @@ class JanusGenerator:
         generation_config.validate()
         self.model._validate_model_kwargs(model_kwargs.copy())
 
+    def prepare_logits_processor(self, generation_config, input_ids, device, logits_processor):
+        """
+        Prepare (and merge) the logits processor.
+
+        Parameters
+        ----------
+        generation_config : GenerationConfig
+            The generation config.
+        input_ids : torch.Tensor
+            The input token ids that serve as the prompt.
+        device : torch.device
+            The device to use for the input tokens.
+        logits_processor : LogitsProcessorList | None
+            The logits processor for the input tokens.
+
+        Returns
+        -------
+        LogitsProcessorList
+            The logits processor.
+        """
+        # Initialize logit processors
+        logits_processor = logits_processor if logits_processor is not None else LogitsProcessorList()
+
+        # Add CFG processor along with user passed logit processor.
+        if generation_config.guidance_scale and generation_config.guidance_scale > 1:
+            logits_processor.append(ClassifierFreeGuidanceLogitsProcessor(generation_config.guidance_scale))
+            generation_config.guidance_scale = None  # Reset to prevent processor duplication.
+
+        # Prepare and merge logits processor
+        logits_processor = self.model._get_logits_processor(
+            generation_config=generation_config,
+            input_ids_seq_length=input_ids.shape[1],
+            encoder_input_ids=input_ids,
+            prefix_allowed_tokens_fn=None,
+            logits_processor=logits_processor,
+            device=device,
+        )
+        return logits_processor
+
     def prepare_inputs_tokens(self, inputs, generation_config, model_kwargs, attention_mask):
         """
-        Prepare the input tokens and model kwargs.
+        Check inputs shapes, and setup special tokens and model kwargs.
 
         Parameters
         ----------
@@ -677,6 +716,14 @@ class JanusGenerator:
         # Prepare special tokens which will be used generate internally.
         kwargs_has_attention_mask = attention_mask is not None
         self.model._prepare_special_tokens(generation_config, kwargs_has_attention_mask, device=input_ids.device)
+
+        # Expand inputs for multiple image generations per prompt.
+        input_ids, model_kwargs = self.model._expand_inputs_for_generation(
+            input_ids=input_ids,
+            attention_mask=attention_mask,
+            expand_size=generation_config.num_return_sequences,
+            **model_kwargs,
+        )
 
         return input_ids, model_kwargs, dtype, device
 
@@ -724,7 +771,10 @@ class JanusGenerator:
 
     def prepare_input_and_cache(self, input_ids, model_kwargs, attention_mask, generation_config, device):
         """
-        Prepare the input tokens, inputs embeddings, model kwargs, batch size, and number of image tokens.
+        Setup input tokens, mask and cache.
+
+        Prepare the input tokens, inputs embeddings, model kwargs, batch size, the number of image tokens,
+        and setup the KV cache.
 
         Parameters
         ----------
@@ -938,9 +988,6 @@ class JanusGenerator:
         # Validate the configuration and model kwargs
         self.validate_config_and_model_kwargs(generation_config, model_kwargs)
 
-        # 2. Initialize logit processors
-        logits_processor = logits_processor if logits_processor is not None else LogitsProcessorList()
-
         # Set `use_cache=True` as we will be using input embeds for generation.
         model_kwargs["use_cache"] = True
 
@@ -950,35 +997,15 @@ class JanusGenerator:
             generation_config.guidance_scale = 5
         model_kwargs["guidance_scale"] = generation_config.guidance_scale
 
-        # 3. Prepare model inputs
+        # 2. Prepare model inputs shapes, and check special tokens.
         input_ids, model_kwargs, dtype, device = self.prepare_inputs_tokens(
             inputs, generation_config, model_kwargs, attention_mask
         )
 
-        # 4. Add CFG processor along with user passed logit processor.
-        if generation_config.guidance_scale and generation_config.guidance_scale > 1:
-            logits_processor.append(ClassifierFreeGuidanceLogitsProcessor(generation_config.guidance_scale))
-            generation_config.guidance_scale = None  # Reset to prevent processor duplication.
+        # 3. Prepare logits processor
+        logits_processor = self.prepare_logits_processor(generation_config, input_ids, device, logits_processor)
 
-        # 5. Prepare logits processor
-        logits_processor = self.model._get_logits_processor(
-            generation_config=generation_config,
-            input_ids_seq_length=input_ids.shape[1],
-            encoder_input_ids=input_ids,
-            prefix_allowed_tokens_fn=None,
-            logits_processor=logits_processor,
-            device=device,
-        )
-
-        # 6. Expand inputs for multiple image generations per prompt.
-        input_ids, model_kwargs = self.model._expand_inputs_for_generation(
-            input_ids=input_ids,
-            attention_mask=attention_mask,
-            expand_size=generation_config.num_return_sequences,
-            **model_kwargs,
-        )
-
-        # 7. Prepare input and model caches
+        # 4. Prepare input and model caches
         input_tokens, inputs_embeds, model_kwargs, batch_size, num_image_tokens = self.prepare_input_and_cache(
             input_ids,
             model_kwargs,
@@ -990,7 +1017,7 @@ class JanusGenerator:
         # Placeholder for generated tokens.
         generated_tokens = torch.zeros((batch_size, num_image_tokens), dtype=dtype, device=device)
 
-        # 8. init attention / hidden states / scores tuples
+        # 5. init attention / hidden states / scores tuples
         output_attentions = generation_config.output_attentions
         output_hidden_states = generation_config.output_hidden_states
         output_scores = generation_config.output_scores
@@ -1002,7 +1029,7 @@ class JanusGenerator:
         decoder_hidden_states = () if (return_dict_in_generate and output_hidden_states) else None
         decoder_attentions = () if (return_dict_in_generate and output_attentions) else None
 
-        # 9. Loop over the latent tokens.
+        # 6. Loop over the latent tokens.
         scores, hidden_state, outputs = self.loop_over_latent_tokens(
             input_tokens,
             input_ids,
@@ -1016,7 +1043,7 @@ class JanusGenerator:
             generation_config,
         )
 
-        # 10. Return the results.
+        # 7. Return the results.
         if return_dict_in_generate:
             if output_scores:
                 raw_scores = tuple(raw_scores) + (scores,) if raw_scores is not None else (scores,)


src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -420,6 +425,48 @@ def causal_lm_logic(model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+        temperature = model.generation_config.temperature if hasattr(model.generation_config, "temperature") else 1.0
+    else:
+        pruna_logger.warning("No generation config found, using default values for top_k and temperature.")
+        # https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig.top_k
+        top_k = 50
+        # https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig.temperature
+        temperature = 1.0
+
+    # We use a generator as in https://github.com/mobiusml/hqq/blob/1f052eb5a0aab0572d380d48b708ae1c74936d23/hqq/utils/generation_hf.py
+    gen = JanusGenerator(
```
sharpenb: This is the single line which is different from causal lm logic. We could factorize it. In general, the function could have the Generator has argument.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+        self.temperature = kwargs.pop("temperature", self.temperature)
+        self.top_k = kwargs.pop("top_k", self.top_k)
+
+        # 1. Handle generation config and model kwargs
+        generation_config = kwargs.pop("generation_config", self.model.generation_config)
+        generation_config = copy.deepcopy(generation_config)
+
+        # Default to "text" generation if mode isn't provided
+        generation_mode = kwargs.pop("generation_mode", "text")
+        if generation_mode == "text":
```
sharpenb, line 969: IN which case would we be in this generation mode?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+            The logits processor for the input tokens.
+        **kwargs : dict
+            Keyword arguments dictionary.
+
+        Returns
+        -------
+        torch.Tensor | GenerateDecoderOnlyOutput
+            The generated latent tokens.
+        """
+        # Extract parameters from kwargs with defaults from instance variables
```
sharpenb, line 959: Is this function inspired from some other code? If yes, it would make sense to link the source code as well. RIght now it involves many steps which are all done in a single function. It would be nice to structure/clean it for code readability. First steps focus on validating, initing variables

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+        logits_processor = self.model._get_logits_processor(
+            generation_config=generation_config,
+            input_ids_seq_length=input_ids.shape[1],
+            encoder_input_ids=input_ids,
+            prefix_allowed_tokens_fn=None,
+            logits_processor=logits_processor,
+            device=device,
+        )
+
+        # 6. Expand inputs for multiple image generations per prompt.
```
sharpenb: Why do we want mulitple image generation per prompt?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+
+        inputs_embeds = self.model.get_input_embeddings()(input_tokens)
+
+        model_kwargs = self.model._get_initial_cache_position(input_ids, model_kwargs)
+
+        if model_kwargs.get("past_key_values", None) is None:
+            # Prepare cache if not provided.
+            model_kwargs["past_key_values"] = self.model._get_cache(
+                cache_implementation=generation_config.cache_implementation or "static",
+                # batch_size should account for both conditional/unconditional input; hence multiplied by 2.
```
sharpenb, line 819: This is unclear why we need batch size = 2

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+        output_scores = generation_config.output_scores
+        output_logits = generation_config.output_logits
+        return_dict_in_generate = generation_config.return_dict_in_generate
+
+        raw_scores = () if (return_dict_in_generate and output_scores) else None
+        raw_logits = () if (return_dict_in_generate and output_logits) else None
+        decoder_hidden_states = () if (return_dict_in_generate and output_hidden_states) else None
+        decoder_attentions = () if (return_dict_in_generate and output_attentions) else None
+
+        for i in range(num_image_tokens):
```
sharpenb, line 873: It would be nice to have relevant independent subfunctions to avoid having everything in a single large function.

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,317 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+            # Pad attention mask to max length to avoid dynamic shapes error during compilation.
+            max_length = model_inputs["past_key_values"].get_max_cache_shape()
+            current_length = model_inputs["attention_mask"].shape[1]
+            if current_length < max_length:
+                padding = torch.zeros((model_inputs["attention_mask"].shape[0], max_length - current_length),
+                                   dtype=model_inputs["attention_mask"].dtype,
+                                   device=model_inputs["attention_mask"].device)
+                model_inputs["attention_mask"] = torch.cat([model_inputs["attention_mask"], padding], dim=1)
+
+            # no compilation for the prefill.
```
sharpenb, line 892: For a future PR, it would be nice also to compile the prefill.

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
     try:  # Try to use pipeline for HF specific HQQ quantization
-        model = HQQModelForCausalLM.from_quantized(
-            model_path, **filter_load_kwargs(HQQModelForCausalLM.from_quantized, kwargs)
+        qmodel = HQQModelForCausalLM.from_quantized(
+            q_path, **filter_load_kwargs(HQQModelForCausalLM.from_quantized, kwargs)
         )
     except Exception as e:  # Default to generic HQQ pipeline if it fails
         pruna_logger.error(f"Error loading model using HQQ: {e}")
-        model = AutoHQQHFModel.from_quantized(model_path, **filter_load_kwargs(AutoHQQHFModel.from_quantized, kwargs))
+        smash_config = SmashConfig()
```
sharpenb: It is unclear why we do these two smash config lines

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
+        smash_config.load_from_json(model_path)
+        compute_dtype = torch.float16 if smash_config["hqq_compute_dtype"] == "torch.float16" else torch.bfloat16
+        qmodel = AutoHQQHFModel.from_quantized(
+            q_path,
+            compute_dtype=compute_dtype,
+            **filter_load_kwargs(AutoHQQHFModel.from_quantized, kwargs)
+        )
 
-    return model
+    original_config = load_json_config(model_path, "config.json")
```
sharpenb, line 393: This is only for Janu because we quantize only the llm part?

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
+        weights["lm_head"] = torch.nn.Linear(1024, 1024).state_dict()
+        # save the weights to the q_path under the format expected by hqq.
+        torch.save(weights, os.path.join(q_path, "qmodel.pt"))
+    else:
+        q_path = model_path
 
     try:  # Try to use pipeline for HF specific HQQ quantization
-        model = HQQModelForCausalLM.from_quantized(
-            model_path, **filter_load_kwargs(HQQModelForCausalLM.from_quantized, kwargs)
+        qmodel = HQQModelForCausalLM.from_quantized(
```
sharpenb: It would be nice to have more explicit/transparent variable names e.g. `quantized_model`

src/pruna/engine/model_checks.py
```diff
@@ -569,3 +570,37 @@ def is_opt_model(model: Any) -> bool:
...
+    ----------
+    model : Any
+        The model to check.
+
+    Returns
+    -------
+    bool
+        True if the model is a Janus LlamaGen AR model, False otherwise.
+    """
+    return is_image_text_to_text_model(model) and model.__class__.__name__ == "JanusForConditionalGeneration"
```
sharpenb: is the second oncidtion not enough in this case?

src/pruna/engine/save.py
```diff
@@ -307,10 +307,32 @@ def save_model_hqq(model: Any, model_path: str, smash_config: SmashConfig) -> No
     from hqq.engine.hf import HQQModelForCausalLM
     from hqq.models.hf.base import AutoHQQHFModel
 
-    if isinstance(model, HQQModelForCausalLM):
-        model.save_quantized(model_path)
+    # we need to create a separate path for the quantized model
+    if hasattr(model, "model") and hasattr(model.model, "language_model"):
+        q_path = os.path.join(model_path, "hqq_language_model")
```
sharpenb: See load function for vairable names :)

src/pruna/engine/save.py
```diff
@@ -307,10 +307,32 @@ def save_model_hqq(model: Any, model_path: str, smash_config: SmashConfig) -> No
...
+    with ModelContext(model) as (pipeline, working_model, denoiser_type):
+        if isinstance(working_model, HQQModelForCausalLM):
+            working_model.save_quantized(q_path)
+        else:
+            AutoHQQHFModel.save_quantized(working_model, q_path)
+        pipeline.working_model = working_model
+
+    # save the rest of the model, if it is a janus like model,
+    # and add a config file to the quantized model path.
+    if hasattr(model, "model") and hasattr(model.model, "language_model"):
```
sharpenb, line 340: Not for this PR but we really need to handle properly target module

sharpenb: Actually, it would be nice to have one save function attached to each target module. Here we have to manually combine hqq save and standard saving.

tests/fixtures.py
```diff
@@ -166,4 +173,7 @@ def get_torchvision_model(name: str) -> tuple[Any, SmashConfig]:
     "llama_3_1_8b": partial(get_automodel_transformers, "NousResearch/Hermes-3-Llama-3.1-8B"),
     "llama_3_tiny_random": partial(get_automodel_transformers, "llamafactory/tiny-random-Llama-3"),
     "dummy_lambda": dummy_model,
+
+    # image generation AR models
+    "tiny_janus_pro": partial(get_janus_model, "loulou2/tiny_janus"),
```
sharpenb, line 160: Is there no official place for tiny janus? Otherwise we can also upload on our org.

sharpenb (changes requested):

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
             " --extra-index-url https://prunaai.pythonanywhere.com/`."
         )
         raise
+    if os.path.exists(os.path.join(model_path, "hqq_language_model")):
```
gsprochette, line 357:
```suggestion
    
    if os.path.exists(os.path.join(model_path, "hqq_language_model")):
```

gsprochette: Could you add a comment explaining in which case this `if` is triggered? Since this replaces the previous behaviour I assume that it triggers when loading quantized Janus but a comment would be nice for future reference

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
+        q_path = os.path.join(model_path, "hqq_language_model")
+        # load the weight on cpu to rename attr -> model.attr,
+        # and also artifically add a lm_head to the weights.
+        weights = torch.load(os.path.join(q_path, "qmodel.pt"), map_location="cpu", weights_only=True)
+        weights = {f"model.{k}" if not k.startswith("model.") else k: v for k, v in weights.items()}
+        weights["lm_head"] = torch.nn.Linear(1024, 1024).state_dict()
+        # save the weights to the q_path under the format expected by hqq.
+        torch.save(weights, os.path.join(q_path, "qmodel.pt"))
```
gsprochette: This `join(...)` is done twice, giving it a clear name (like `expected_qmodel_path` or something) would simplify the code and can be clearer than having the comment above

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
     except Exception as e:  # Default to generic HQQ pipeline if it fails
         pruna_logger.error(f"Error loading model using HQQ: {e}")
-        model = AutoHQQHFModel.from_quantized(model_path, **filter_load_kwargs(AutoHQQHFModel.from_quantized, kwargs))
+        smash_config = SmashConfig()
+        smash_config.load_from_json(model_path)
+        compute_dtype = torch.float16 if smash_config["hqq_compute_dtype"] == "torch.float16" else torch.bfloat16
+        qmodel = AutoHQQHFModel.from_quantized(
+            q_path,
+            compute_dtype=compute_dtype,
+            **filter_load_kwargs(AutoHQQHFModel.from_quantized, kwargs)
```
gsprochette: I think this will result in an error if the user provided a `compute_dtype` in `kwargs`. `compute_dtype` should probably be added in kwargs before this, and only if it is not already there

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
+        )
 
-    return model
+    original_config = load_json_config(model_path, "config.json")
+    if original_config["architectures"][0] == "JanusForConditionalGeneration":
+        cls = getattr(transformers, "JanusForConditionalGeneration")
```
gsprochette, line 395: Why `getattr(transformers, "JanusForConditionalGeneration")` and not `transformers.JanusForConditionalGeneration`? Did you intend to give a `None` value if unavailable and send an explicit error?

src/pruna/engine/load.py
```diff
@@ -340,16 +340,41 @@ def load_hqq(model_path: str, **kwargs) -> Any:
...
+        model = cls.from_pretrained(model_path, **kwargs)
```
gsprochette, line 396: Do we need to pull the `torch_dtype=compute_dtype` from the qmodel or is it redundant?

src/pruna/algorithms/quantization/hqq.py
```diff
@@ -92,7 +92,7 @@ def model_check_fn(self, model: Any) -> bool:
         bool
             True if the model is a causal language model, False otherwise.
         """
-        return is_causal_lm(model)
+        return is_causal_lm(model) or is_janus_llamagen_ar(model)
```
gsprochette, line 97: Need to update the docstring

src/pruna/engine/utils.py
```diff
@@ -345,6 +348,8 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
             self.pipeline.transformer = self.pipeline.working_model
         elif hasattr(self.pipeline, "unet"):
             self.pipeline.unet = self.pipeline.working_model
+        elif hasattr(self.pipeline, "model") and hasattr(self.pipeline.model, "language_model"):
+            self.pipeline.model.language_model = self.pipeline.working_model
```
gsprochette, line 518: Unrelated to this PR but I think it's weird this context manager expects the user to create a working_model attribute, and also it doesn't appear in the doc. I love the context manager I think it's a beautiful way to handle these different cases, so two things to make it work:

For now, the doc of this context should be edited and we should add an explicit error in `__exit__` if `pipeline.working_model` doesn't exist.

In a future PR we should either
- create the self.pipeline.working_model in the context and don't pass it when entering, so the user can access a consistent attribute and can edit it since we'll recover it as an attribute when exiting, or
- provide a way for the user to edit the working_model from outside (it could be a function, or we could pass the working_model in a list so we only provide a pointer to the working model, or other idea)

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -202,6 +203,10 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
 
         if is_causal_lm(model):
             return causal_lm_logic(model, smash_config)
+
+        if is_janus_llamagen_ar(model):
+            return compile_janus_logic(model, smash_config)
```
gsprochette: shouldn't it just be `janus_logic`? Or `causal_lm_logic` should be renamed to `compile_causal_lm_logic` and same for the rest. Either way is fine but it should be consistent

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -202,6 +203,10 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
 
         if is_causal_lm(model):
             return causal_lm_logic(model, smash_config)
+
+        if is_janus_llamagen_ar(model):
```
gsprochette: Why are these `if` and not `elif`? It would be slightly cleaner I think, and the last return could be `else`.

gsprochette (changes requested):
Awesome work, can't wait to have this working for Janus in Pruna! I also hope that the HF PR for adding generate functions to LlamaGen models will be added soon so we can remove the Generator code. For the rest of the code I left a few comments, nothing breaking only trying to improve already good code and make some double checks here and there :)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: Good idea! I have unified it :)

src/pruna/algorithms/compilation/utils.py
llcnt: Yes, I have specified it at the beginning of the class: https://github.com/PrunaAI/pruna/blob/1a6c0a5383405eeaa1cf16da4b0432ddddd2b30a/src/pruna/algorithms/compilation/utils.py#L563

Most of the code is a copy/paste from the above link. However, our changes take place in the middle of the code. And this code is written as a long single function. So we can not extract some sub-functions, and wrap them around our modification.
Indeed we can create ourself these tiny sub-functions:
- main advantage=easier to read;
- main drawback=tricky to maintain on the long term when the file from transformers is modified.

What do you think ? :)

src/pruna/algorithms/compilation/utils.py
llcnt: We need this to enable CFG (one element of the batch will be conditionned to the prompt, and the other one will be uncondtionned) !

src/pruna/algorithms/compilation/utils.py
llcnt: I am note sure to understand what is the idea behind compiling the prefill. As it is run once, the computation is not the bottleneck, no?

src/pruna/algorithms/compilation/utils.py
llcnt: see the discussion above

src/pruna/algorithms/compilation/utils.py
llcnt: Interesting remark ;) 
In fact Janus is not only a Image gen model. It can also do multimodal understanding. So you do Image+Text -> Text

src/pruna/algorithms/compilation/utils.py
llcnt: It is funny that you ask because there was actually the same question raised during the PR of janus in transformers. ;)
The idea is to enable the users to generate several different images for the same prompt (so it will internally change your batch from batch_size -> batch_size*num_return_sequences). This is something initially implemented in the deepseek/janus codebase and that was kept. IMO it is a nice feature that we can also keep :) (btw this is also a feature that exist in transformers for musicgen model)

src/pruna/engine/load.py
llcnt: In case the compute_dtype on which the quantized model was saved is different from the one used by a user after loading, the inference will bug. So we need to load the saved smash_config and get the initial compute_dtype.
As mentioned by gaspar, this could introduce a bug if the user also specify the compute_dtype as part of the kwargs: so I slightly modified the code to handle this :)

src/pruna/engine/load.py
llcnt: Yes :)

src/pruna/engine/load.py
llcnt: Good point :) I have modified the code

src/pruna/engine/model_checks.py
llcnt: You are absolutely right :)

src/pruna/engine/save.py
llcnt: done ;)

src/pruna/engine/save.py
llcnt: I think I dont get your second comment ahah

src/pruna/engine/load.py
llcnt: Good remark, I have added a comment :)

src/pruna/engine/load.py
llcnt: done :)

src/pruna/engine/load.py
llcnt: Good catch ! I fix it (see my comment above on bertrands comment)

src/pruna/engine/load.py
llcnt: Nop, no need here (as the quantized model lm is already loaded correctly in the line above, we only load the rest of the model, no need to specify the dtype here again) ;)

src/pruna/engine/load.py
llcnt: I dont have a specific point of view on this one. I just get inspiration from the rest of pruna codebase, and I can see that we use `getattr(transformers, "STR")`, so I followed the same pattern :)

src/pruna/algorithms/quantization/hqq.py
llcnt: Nice catch :) it is upd now

src/pruna/engine/utils.py
llcnt: I completely agree :)

src/pruna/algorithms/compilation/torch_compile.py
llcnt: Good point! I have fused the 2 logic into a single function `causal_lm_or_janus_logic`

src/pruna/algorithms/compilation/torch_compile.py
llcnt: see comment above :)

src/pruna/engine/load.py
```diff
@@ -352,21 +352,51 @@ def load_hqq(model_path: str, smash_config: SmashConfig, **kwargs) -> Any:
 
     algorithm_packages = HQQQuantizer().import_algorithm_packages()
 
+    # if the model is a janus like model, we need to load the quantized model from the hqq_language_model directory
+    if os.path.exists(os.path.join(model_path, "hqq_language_model")):
+        quantized_path = os.path.join(model_path, "hqq_language_model")
```
sharpenb: These lines are a bit custom for Janus.
- Cna we really expect this fodler structure?
- Why do we have to resave the model at loading time? It sounds like some silent effects.

src/pruna/engine/model_checks.py
```diff
@@ -570,3 +571,37 @@ def is_opt_model(model: Any) -> bool:
...
+    ----------
+    model : Any
+        The model to check.
+
+    Returns
+    -------
+    bool
+        True if the model is an image text to text model, False otherwise.
+    """
+    return isinstance(model, tuple(MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING.values()))
```
sharpenb: WHy do we mean by image text to text model? Do we mean image+text->text? :)

src/pruna/engine/load.py
gsprochette: I meant reusing this name also when loading the weights 7 lines above :sweat_smile:

src/pruna/engine/save.py
sharpenb: Sorry, that was unclear. I udpated my comment :)

src/pruna/algorithms/compilation/utils.py
sharpenb: Makes sense. If you have any idea to make the code more readable I feel that it would be a big plus. We cna go through it together if needed.

gsprochette:
Looks good to me, there's just this `expected_quantized_model_path` that could be re-used. If Bertrand is satisfied with your answers this is ready to go for me, thanks for the updates :)

gsprochette (approved):
Looks good to me, there's just this expected_quantized_model_path that could be re-used to make the code a bit cleaner, see comment. If Bertrand is satisfied with your answers this is ready to go for me, thanks for the updates :)

src/pruna/engine/load.py
llcnt: Ah oki, I see :)

src/pruna/engine/model_checks.py
llcnt: Yep :) (even if the model is not strictly outputing only text... but this is a long debate lol)
I made the comment more clear to follow the [comment from hf](https://github.com/huggingface/transformers/pull/36053/files#r1978961386)
Edit: in fact we don't need it anymore because we directly use the `is_janus_llamagen_ar` function now. So I just removed it :)

llcnt:
I have edited the main comment, and I am fixing the code (will appear in the next push), for enabling speedups with torch==2.7

src/pruna/algorithms/compilation/utils.py
llcnt: I have added 5 functions: `validate_config_and_model_kwargs, prepare_inputs_tokens, get_initial_cache_position, prepare_input_and_cache, loop_over_latent_tokens` to ease the reading of the `generate` function.

src/pruna/engine/load.py
llcnt: Good remark!
- We have this folder structure because this is how we have defined it in the `save.py` file (I took inspiration on how pipelines are saved in `diffusers`). We could have done something different (for eg we could provide 2 distinct paths 'backbone_model_path' and 'janus_structure_path' and load the hqq-quantized model from the first one, and load the rest of the janus model (without its language_model) from the later), but I feel it is less natural. We can change it if you have a better idea about how to save/load this 2 components :)
- yes I agree this is painful. We could move this inside the save function, but we can not avoid the 'resave'. Here's why: the from_quantized function in hqq expects a model with a strict structure: the model weights should be wrapped into a model class that contains 'lm_head' and the 'layers' separately. But the language model we have inside janus does not follow this structure (even though it is a llama). So once the hqq-quantized-model is saved, we have to manually load it (on cpu) to artificially rename its layer_0.q_proj into model.layer_0.q_proj and add a random lm_head, so that the .from_quantized is tricked ;)

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,495 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+            generation_config=generation_config,
+            input_ids_seq_length=input_ids.shape[1],
+            encoder_input_ids=input_ids,
+            prefix_allowed_tokens_fn=None,
+            logits_processor=logits_processor,
+            device=device,
+        )
+
+        # 6. Expand inputs for multiple image generations per prompt.
+        input_ids, model_kwargs = self.model._expand_inputs_for_generation(
```
sharpenb: Step 3. and Step 6. prepare inputs in some way. Could we factorize these in one step?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,495 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+        input_ids, model_kwargs, dtype, device = self.prepare_inputs_tokens(
+            inputs, generation_config, model_kwargs, attention_mask
+        )
+
+        # 4. Add CFG processor along with user passed logit processor.
+        if generation_config.guidance_scale and generation_config.guidance_scale > 1:
+            logits_processor.append(ClassifierFreeGuidanceLogitsProcessor(generation_config.guidance_scale))
+            generation_config.guidance_scale = None  # Reset to prevent processor duplication.
+
+        # 5. Prepare logits processor
```
sharpenb: It is unlcear why step 5. prepare logit processors while step 4. alreayd used the suer passed processor. Could we also merge all processors preparation step?

src/pruna/algorithms/compilation/utils.py
```diff
@@ -550,3 +554,495 @@ def generate(self, *args, **kwargs) -> torch.Tensor:
...
+                past_length = cache.get_seq_length()
+
+            cache_position = cache_position[past_length:]
+
+        model_kwargs["cache_position"] = cache_position
+        return model_kwargs
+
+    def prepare_input_and_cache(self, input_ids, model_kwargs, attention_mask, generation_config, device):
+        """
+        Prepare the input tokens, inputs embeddings, model kwargs, batch size, and number of image tokens.
```
sharpenb: THe name of the function suggest that cache is also prepared but the comment not.

src/pruna/algorithms/compilation/utils.py
sharpenb: Thanks! I think that it reads already nicer. Each of this function has still some level of complexity which might be fine for now.

src/pruna/engine/load.py
sharpenb: After discussion, parts of these will be tackled in some future "target_module" PR.

sharpenb (approved):
Thanks for tackling the comments! I left a couple more but I think that it should be good to go then

src/pruna/algorithms/compilation/utils.py
llcnt: Nice catch! Indeed step6 only depends on the previous step3, I merged it and put setp6 into the sub-function "self.prepare_inputs_tokens" :)

src/pruna/algorithms/compilation/utils.py
llcnt: You are definitely right ;) I merged all process related to logits_processor into a single sub-function

(ps: it is maybe unclear in the function name self.model._get_logits_processor, but this function will use [the function from transformers](https://github.com/huggingface/transformers/blob/1580f6465347df4e8d7c12cd4d3dc603b8a689ad/src/transformers/generation/utils.py#L1371) that merge the user-defined logits_processor together with the logits_processor defined into the generation_config. I have added a small comment in the function to highlight this)

src/pruna/algorithms/compilation/utils.py
llcnt: I have modified the docstring accordingly :)

PR #135, opened 2025-05-19
feat: extend device compatibility of algorithms

johannaSommer:
## Description
Following up on #128 this PR extends the algorithms in `pruna` with specifications on whether distributed base models are supported or not. As a consequence of this the following adjustments have been made:
- updating the Pruna base class
- updating the documentation generation
- additionally to the device consistency checks (model == smashconfig) we now also give guidance on whether an algorithm is applicable to a model in a certain device state
- extend the tests with checks on the model device state.

## Related Issue
Closes #147 

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?
Ran all algorithm tests locally, only known failures occured.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Additional Notes
Additionally to this PR and #128 there will be one more PR tackling the saving and loading for models in a distributed state.

PR #148, opened 2025-05-22
feat: improve overall device placement handling

### LINKED ISSUE #108: [FEATURE] Improve overall device placement handling

I noticed this broader habbit in #97 and partially covered there, but I realised there is a wider habit of enforcing `cuda` usage, although people might still be interested in using `cpu` or `mps`, and @sharpenb mentioned we wanted to try supporting it.

I propose re-using the logic in #97 and doing some other things:
- Always check and warm for device placement, where we prefer `mps`/`cuda` when no preference is given. 
- Warn when using `mps` about limited support while nudging users to open an issue in case of weird behaviour. (torch..compile implements internal MPS warning, for example)

davidberenstein1957:
## Description

Before, running the following code on CPU would fail because it was being forced on CUDA.
```python
from pruna.evaluation.metrics import CMMD

CMMD()
```

## Related Issue

Fixes #108

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -266,19 +266,19 @@ def check_device_compatibility(device: str | torch.device | None) -> str:
 
     if device is None:
         device = "cuda" if torch.cuda.is_available() else "mps" if torch.backends.mps.is_available() else "cpu"
-        pruna_logger.info(f"No device specified. Using best available device: '{device}'")
+        pruna_logger.info(f"Using best available device: '{device}'")
         return device
 
     if device == "cpu":
         return "cpu"
 
     if device == "cuda" and not torch.cuda.is_available():
-        pruna_logger.warning("'cuda' requested but not available. Falling back to 'cpu'")
-        return "cpu"
+        pruna_logger.warning("'cuda' requested but not available.")
+        return check_device_compatibility(device=None)
 
     if device == "mps" and not torch.backends.mps.is_available():
-        pruna_logger.warning("'mps' requested but not available. Falling back to 'cpu'")
-        return "cpu"
+        pruna_logger.warning("'mps' requested but not available.")
+        return check_device_compatibility(device=None)
 
     return device
 
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -244,9 +244,9 @@ def determine_dtype(pipeline: Any) -> torch.dtype:
     return torch.float32
 
 
-def check_device_compatibility(device: str | torch.device | None) -> str:
+def set_to_best_available_device(device: str | torch.device | None) -> str:
     """
-    Validate if the specified device is available on the current system.
+    Set the device to the best available device.
 
     Supports 'cuda', 'mps', 'cpu' and other PyTorch devices.
     If device is None, the best available device will be returned.
@@ -274,11 +274,11 @@ def check_device_compatibility(device: str | torch.device | None) -> str:
 
     if device == "cuda" and not torch.cuda.is_available():
         pruna_logger.warning("'cuda' requested but not available.")
-        return check_device_compatibility(device=None)
+        return set_to_best_available_device(device=None)
 
     if device == "mps" and not torch.backends.mps.is_available():
         pruna_logger.warning("'mps' requested but not available.")
-        return check_device_compatibility(device=None)
+        return set_to_best_available_device(device=None)
 
     return device
 
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -272,6 +272,9 @@ def set_to_best_available_device(device: str | torch.device | None) -> str:
     if device == "cpu":
         return "cpu"
 
+    if device == "accelerate" and torch.cuda.is_available():
+        return "accelerate"
+
     if device == "cuda" and not torch.cuda.is_available():
         pruna_logger.warning("'cuda' requested but not available.")
         return set_to_best_available_device(device=None)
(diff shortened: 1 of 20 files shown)


src/pruna/evaluation/metrics/metric_cmmd.py
```diff
@@ -41,8 +42,9 @@ class CMMD(StatefulMetric):
     ----------
     *args : Any
         Additional arguments to pass to the StatefulMetric constructor.
-    device : str | torch.device
-        The device to run the CLIP model on to calculate the embeddings for the metric.
+    device : str | torch.device | None, optional
+        The device to be used for smashing, e.g., 'cuda' or 'cpu'. Default is None.
```
guennemann: Some minor point: It says "the device used for smashing". But we are in a metric here. So this should actually be "the device used for inference". In principle, the device for smashing and (later) doing the inference/metric could be different. Might be good to streamline this across all mentions.

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -54,8 +55,9 @@ class InferenceTimeStats(BaseMetric):
         The number of batches to evaluate the model.
     n_warmup_iterations : int, default=10
         The number of warmup batches to evaluate the model.
-    device : str | torch.device, default="cuda"
-        The device to evaluate the model on.
+    device : str | torch.device | None, optional
+        The device to be used for smashing, e.g., 'cuda' or 'cpu'. Default is None.
```
guennemann: See before: Please replace smashing with inference/evaluation

guennemann (changes requested):
Very nice. See some minor request inline.

guennemann (approved):
LGTM! Thanks.

johannaSommer (approved):
LGTM from my side w.r.t. device casting, etc, but I really feel @begumcig should review this.

src/pruna/evaluation/metrics/metric_memory.py
```diff
@@ -328,7 +328,7 @@ def _load_and_prepare_model(self, model_path: str, model_cls: Type[PrunaModel])
         model = model_cls.from_pretrained(
             model_path,
         )
-        model.move_to_device("cuda")
```
begumcig, line 331: The memory_metric is designed to measure the GPU memory, it should ideally give an error and exit if the device is CPU :)

src/pruna/evaluation/metrics/registry.py
```diff
@@ -123,4 +124,4 @@ def get_metrics(cls, names: List[str], **kwargs) -> List[BaseMetric | StatefulMe
         -------
             A list of metric instances.
         """
-        return [cls.get_metric(name, **kwargs) for name in names]
+        return [cls.get_metric(name, **filter_load_kwargs(cls.get_metric, **kwargs)) for name in names]
```
begumcig: As far as I understood, filter_load_kwargs checks the function 'get_metric' here and filters the keyword arguments to match to the arguments of the get_metric. However here we could also pass along some parameters to the metric itself, so I don't think we should filter it this way

src/pruna/engine/utils.py
```diff
@@ -266,21 +266,30 @@ def check_device_compatibility(device: str | torch.device | None) -> str:
...
+        pruna_logger.info(f"Using best available device: '{device}'")
         return device
 
     if device == "cpu":
         return "cpu"
 
-    if device == "cuda" and not torch.cuda.is_available():
-        pruna_logger.warning("'cuda' requested but not available. Falling back to 'cpu'")
-        return "cpu"
+    if device == "accelerate":
```
begumcig, line 275: I understand the intention here, but we are basically allowing the user to set the device to accelerate here. It's not currently supported for the evaluation agent, do you think this could give off a wrong impression?

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, task: Task) -> None:
         self.task = task
         self.first_model_results: List[MetricResult] = []
         self.subsequent_model_results: List[MetricResult] = []
-        self.device = self.task.device
+        self.device = set_to_best_available_device(self.task.device)
```
begumcig, line 44: slay!

src/pruna/evaluation/task.py
```diff
@@ -139,14 +147,14 @@ def get_metrics(request: str | List[str | BaseMetric | StatefulMetric]) -> List[
     """
     if isinstance(request, List):
         if all(isinstance(item, BaseMetric | StatefulMetric) for item in request):
-            return _process_metric_instances(cast(List[BaseMetric | StatefulMetric], request))
+            return _process_metric_instances(request=cast(List[BaseMetric | StatefulMetric], request))
```
begumcig, line 150: I love that now whenever we instantiate metrics from "strings" the Task device and the devices of the metrics will align! This was something that was also bugging me for the longest time so thank you so much for doing this! Do you think it also makes sense to extend this to metric instances? Totally get it if out of scope for now, but it would complete the process of making sure that every part of the agent is on the same device :)

src/pruna/engine/utils.py
davidberenstein1957: Hi, there is a PR to add support for this, right? https://github.com/PrunaAI/pruna/pull/135

src/pruna/evaluation/metrics/registry.py
davidberenstein1957: Yes, I agree.

src/pruna/evaluation/task.py
davidberenstein1957: I don't think it is needed for the instances. They are either instantiated without passing a device (an therefore on the optimal one), or instantiated manually to run on other devices. We don't need them to be right?

PR #154, opened 2025-05-28
feat: modular pruning

begumcig:
## Description

Enables safe, module-level pruning by finding each target module’s interior, boundary, and exterior, pruning only the interior while auto-patching surrounding shapes. Works agnostically for sub modules.

## Type of Change

## Checklist

## Additional Notes
Here's an example code snippet:

```smash_config = SmashConfig()
smash_config['pruner'] = 'torch_structured'
# Add your target module
smash_config.add_target_module(model.model.token_rep_layer.bert_layer) 
# Now finds the interior   of the target module
model = smash(model, config, experimental=True)

### CODE REVIEW

File: src/pruna/config/smash_config.py [hunk 1]
```diff
@@ -445,6 +445,19 @@ def add_processor(self, processor: str | ProcessorMixin) -> None:
...
+
+        Parameters
+        ----------
+        target_module : Any
+            The target module to prune.
+        """
+        if self["pruner"] != "torch_structured":
+            pruna_logger.error("Target module is only supported for torch_structured pruner")
+        self._target_module = target_module
+
```
guennemann: A more general discussion (which we don't need to resolve for the current PR but fine later): Add_target_module is something very general and goes beyond pruners. Here you only check the compatability w.r.t. a specific "pruner" though.

A small point: You could improve the error message by distinguishing between "no pruner has been provided in the smash config" and "a non-compatible pruner has been provided"

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -46,6 +45,10 @@ class TorchStructuredPruner(PrunaPruner):
 
     Structured pruning removes entire units like neurons, channels, or filters from a network, leading to a more compact
     and computationally efficient model while preserving a regular structure that standard hardware can easily optimize.
+
+    Note: If you would like a to prune a target module,
```
guennemann: typo. remove "a"

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -169,28 +173,50 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                    batch[k] = v.to(device)
+            example_input = batch
+        else:
+            # PrunaDataModule always returns a tuple of two tensors, the first is the input.
+            example_input = batch[0][:1, :].to(device)  # type: ignore[arg-type]
 
-        ch_groups, num_heads, ignored_layers = get_prunable_layers(model, smash_config, imported_modules)
+        # Get the target module to prune, and it's boundary and exterior
+
+        target_module = get_target_module(model, imported, smash_config)
```
guennemann, line 186: Just for a future release: It would be nice to allow handle multiple target modules.

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
             import timm
             import torch_pruning as tp
             import torchvision
+            from gliner import GLiNER
             from timm.models.mvitv2 import MultiScaleAttention
             from timm.models.mvitv2 import MultiScaleVit as MViT
+            from transformers.models.llama.modeling_llama import LlamaForCausalLM as Llama
+            from transformers.models.opt.modeling_opt import OPTForCausalLM as Opt
+            from transformers.models.vit.modeling_vit import ViTForImageClassification as ViT
```
guennemann: For a future release: Why importing all of these if only a specifc model is used? Would be good to handle this differently.

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -37,7 +30,13 @@
 from pruna.engine.save import SAVE_FUNCTIONS
 from pruna.logging.logger import pruna_logger
 
-is_gradient_based = ["TaylorImportance", "HessianImportance"]
+_QKV_PAT = re.compile(r"(q|k|v).*proj|query|key|value", re.I)
```
sharpenb: This woudl be helpful to add some comments to make clear what is the purpose of these general variables (since this is a important trick) :)

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -128,18 +129,21 @@ def model_check_fn(self, model: Any) -> bool:
         Returns
         -------
         bool
-            True if the model is a torch.nn.Module, False otherwise.
+            True if the model is supported, False otherwise.
         """
         # Torch structured pruning is currently broken for LLMs, to be fixed
         if is_causal_lm(model):
```
sharpenb: WOuld Opt and Llama not be CausalLM? It is actually not clear why we need l.135 since all causal lm that are not menion below should return false.

pyproject.toml
```diff
@@ -87,6 +87,7 @@ whisper-s2t = "==1.3.0"
 hqq = "*"
 torchao = "*"
 llmcompressor = "*"
+gliner = "0.2.19"
```
sharpenb: We only aim to support a single version of Gliner?

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -169,28 +173,50 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
             model.float()
 
         # Retrieve the importance function or class from the mapping based on the pruning type
-        importance_function = getattr(imported_modules["tp"].importance, smash_config["type"])
+        importance_function = getattr(imported["tp"].importance, smash_config["type"])
+
+        # Get the example input and move to device correctly if it's a dict
+        batch = next(iter(smash_config.train_dataloader()))
+        if isinstance(batch, dict):
```
sharpenb, line 175: Coudl we make clearer why we need if/else here. IF sounds like the batch format should be handled by the Pruna data module making unclear why we need to have the if/else.

src/pruna/algorithms/pruning/torch_structured.py
sharpenb: 100% agree. This is planned for future release :)

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -169,28 +173,50 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+            example_input = batch
+        else:
+            # PrunaDataModule always returns a tuple of two tensors, the first is the input.
+            example_input = batch[0][:1, :].to(device)  # type: ignore[arg-type]
 
-        ch_groups, num_heads, ignored_layers = get_prunable_layers(model, smash_config, imported_modules)
+        # Get the target module to prune, and it's boundary and exterior
+
+        target_module = get_target_module(model, imported, smash_config)
+        dg = (
```
sharpenb: For transparency reasons, I would recommend to rename "dg" -> "dependency_graph"

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -169,28 +173,50 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+            .DependencyGraph()
+            .build_dependency(model, example_input, output_transform=safe_output_transform)
+        )
+        boundary, exterior = get_boundary_and_exterior(target_module, model, dg)
+
+        # Get the ignored layers
+        ignored_layers = get_ignored_layers(boundary, exterior, model, imported)
+
+        # Get the number of heads
+        num_heads = find_attention_blocks(target_module)
```
sharpenb: It ousnds like this function return the number of head inside some target module but it is not really reflect in the funciton name. Could we rename the function or `num_heads` for sthg more transparent? :)

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+    target_module : nn.Module
+        The target module to prune.
     model : nn.Module
-        The model object to perform tensor pruning on.
-    smash_config : SmashConfigPrefixWrapper
-        A dictionary containing the pruning parameters.
-    imported_modules : Dict
-        Dictionary containing the imported modules.
+        The full model.
+    dg : tp.DependencyGraph
```
sharpenb: I would recommend `dependency_graph` for transparency of the variable name

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+    enc_inside = enc_inside_prms | enc_inside_modules  # We want both the parameters and the modules
+
+    interior = real_named & enc_inside  # interior is intersection of entire model and target module
+    exterior = real_named - interior  # exterior is the rest of the modules
+
+    def touches_exterior(node):
+        stack, seen = node.inputs + node.outputs, set()
+        while stack:
+            n = stack.pop()
+            if n in seen:
```
sharpenb: We could do the inverse condition and avoid having the `continue` line

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+                continue
+            seen.add(n)
+            mod = n.module
+            # We don't want the auxiliary nodes like autograd nodes
+            # So we only want the modules that have actual names
+            if mod in real_named:
+                return mod in exterior  # If the input or output is in the exterior, then the module touches the exterior
```
sharpenb: Why making a proper function (not inside another fucntion) out of `touches_interior`. There is now a side effect with the variable `exterior` that could be handle simply by having exterior as input variable.

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+            stack.extend(n.inputs)
+            stack.extend(n.outputs)
+        return False
+
+    boundary = {m for m in interior if touches_exterior(real_nodes[m])}
+
+    return boundary, exterior
+
+
+def _get_q_head_attr_value(m: nn.Module) -> Optional[int]:
```
sharpenb: The `_get_*_attr_value` fucntions are very similar up to the `*_ATTRS` usage. We could factorize these functions

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+
+    Returns
+    -------
+    Dict[nn.Linear, int]
+        The projection to head_count map.
+    """
+    mapping: Dict[nn.Linear, int] = {}
+    for m in model.modules():
+        # Is m an attention block? Keep going if not.
+        if (heads := _get_q_head_attr_value(m)) is None:
```
sharpenb: Inverse conditions could avoid having `continue` line

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+            continue
+        # Do we have a separate KV head count?
+        kv_heads = _get_kv_head_attr_value(m)
```
sharpenb: `kv_heads` is actually a head count right? I would recommend to name it accordingly. Same for `heads` above? :)

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+    # Get the modules and parameters that are inside the target module
+    enc_inside_prms = set(target_module.parameters())
+    enc_inside_modules = set(target_module.modules())
+    enc_inside = enc_inside_prms | enc_inside_modules  # We want both the parameters and the modules
+
+    interior = real_named & enc_inside  # interior is intersection of entire model and target module
+    exterior = real_named - interior  # exterior is the rest of the modules
+
+    def touches_exterior(node):
+        stack, seen = node.inputs + node.outputs, set()
```
sharpenb, line 317: Some docstring would be nice. It sounds like a node search in a graph to find the exterior

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
-        if isinstance(m, LlamaAttention):
-            num_heads[m.q_proj] = m.num_heads
-            num_heads[m.k_proj] = m.num_key_value_heads
-            num_heads[m.v_proj] = m.num_key_value_heads
+        if (_get_q_head_attr_value(m)) is None:
+            continue
+
+        q_proj = (
+            getattr(m, "q_proj", None) or getattr(m, "query", None) or getattr(m, "query_proj", None)
+        )  # Maybe this should be the QKV pattern instead.
```
sharpenb: The comment suggests that sthg is still unclear here.

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+        if q_proj not in new_heads:
+            continue  # This block was ignored
+
+        new_h = new_heads[q_proj]
```
sharpenb: Could we make the name of all variables more consistent and transparent? In particular it owuld help to understand what are counts or not :)

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+        if len(qkv) < 3:
+            continue
+        # Map every projection to the original head count
+        for name, proj in qkv:
+            is_kv = name.lower().startswith(("k", "v"))
+            mapping[proj] = kv_heads if is_kv and kv_heads is not None else heads
+    return mapping
+
+
+def patch_heads(model: nn.Module, new_heads: Dict[nn.Linear, int]):
```
sharpenb: `update_heads_attribute` could maybe be a more trasnparent name here?

sharpenb (changes requested):
Very cool update! I left some comment to help for code quality. Some of these will indeed be updated in future "target_modules" PRs ;)

src/pruna/algorithms/pruning/torch_structured.py
begumcig: Agreed. This was in the original implementation, so I respected it but I also don't understand the purpose. Removing!

src/pruna/algorithms/pruning/torch_structured.py
begumcig: Yes but one other way to create a PrunaDataModule is by giving it 3 sets of datasets for train, validation and test. So Someone can create a PrunaDataModule wrapper around any dataset :). (I for one also sometimes do it!). So even though the datasets we provide never returns dicts, we cannot be sure this will be the general use case.

src/pruna/algorithms/pruning/torch_structured.py
begumcig: Agreed. I think this is how we do our imports in for every algorithm implementation. Maybe a general refactoring of this to be more dynamic would make sense in the future

src/pruna/algorithms/pruning/torch_structured.py
begumcig: I made it a nested function since it's a helper that's only going to be used by the boundary function. We are using both "real_named" and the "exterior" we would have to pass as arguments to the function.

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+    return boundary, exterior
+
+
+def _get_q_head_attr_value(m: nn.Module) -> Optional[int]:
+    """
+    Get the head count attribute value from the module.
+
+    Parameters
+    ----------
+    m : nn.Module
```
GreatBahram: Just passing by here:
I see `m` and `mod` are used for module; I wonder why not using `module` the whole word; it's more readable and yet short enough, wdyt?

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+            return getattr(m.config, attr)
+    return None
 
-    elif isinstance(model, ViT):
-        ch_groups, num_heads, ignored_layers = get_vit_params(model, smash_config)
 
-    elif isinstance(model, imported_modules["timm"].models.convnext.ConvNeXt):
-        ignored_layers = [model.stem, model.head]
-        ch_groups, num_heads = dict(), dict()
+def _get_kv_head_attr_value(m: nn.Module) -> Optional[int]:
```
GreatBahram: `_get_kv_head_attr_value` and `_get_q_head_attr_value` are almost identical I wonder if we can generalise the function, wdyt?

something like, not tested though :upside_down_face: 
```
def _get_attr_value(module: nn.Module, attr_names: list[str]):
    for attr in attr_names:
        if hasattr(module, attr):
            return getattr(module, attr)
        if hasattr(module, "config") and hasattr(module.config, attr):
            return getattr(module.config, attr)
    return None


def _get_q_head_attr_value(module: nn.Module):
    return _get_attr_value(module, _Q_HEAD_ATTRS)

def _get_kv_head_attr_value(module: nn.Module):
    return _get_attr_value(module, _KV_HEAD_ATTRS)
```

src/pruna/algorithms/pruning/torch_structured.py
GreatBahram: This approach is a bit old, see here:
https://docs.python.org/3/library/typing.html#typing.Optional
```suggestion
def _get_q_head_attr_value(m: nn.Module) -> int | None:
```

src/pruna/algorithms/pruning/torch_structured.py
```diff
@@ -224,154 +250,302 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
-        if isinstance(m, OPTAttention):
-            num_heads[m.q_proj] = m.num_heads
-            num_heads[m.k_proj] = m.num_heads
-            num_heads[m.v_proj] = m.num_heads
+    return [(name, sub) for name, sub in m.named_children() if isinstance(sub, nn.Linear) and _QKV_PAT.fullmatch(name)]
 
-    ignored_layers = [model.lm_head]
-    return dict(), num_heads, ignored_layers
 
+def find_attention_blocks(model: nn.Module) -> Dict[nn.Linear, int]:
```
GreatBahram: NIT: `module` is called model here :shrug:

GreatBahram: This approach is deprecated, see here:
https://docs.python.org/3/library/typing.html#typing.Dict
```
suggestion
def find_attention_blocks(model: nn.Module) -> dict[nn.Linear, int]:
```

pyproject.toml
begumcig: Because I have only tested with that version, so I was being cautious. Changed!

src/pruna/config/smash_config.py
begumcig: Yes, I think there's a more general discussion around target modules. Probably for different algorithms target_modules will also be in the algorithm's hyperparameters in the future. For now, this use is limited to the structured pruner. :)

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -116,7 +116,10 @@ def _time_inference(self, model: PrunaModel, x: Any) -> float:
         """
         if self.timing_type == "async" or self.device == "cpu":
             startevent_time = time.time()
-            _ = model(x, **model.inference_handler.model_args)
+            if isinstance(x, dict):
```
sharpenb: Can it be managed by a (potentially new) inference handler?

sharpenb (approved):
Looks very good to me! I left a comment in case we can extend the inference handlers to make some things smoother. I feel that it is ready to merged though

PR #165, opened 2025-05-30
 feat: switch pruna from poetry to uv

GreatBahram:
## Description
- [x] remove poetry as a dependency
- [x] use uv everywhere including github files

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


pyproject.toml
```diff
@@ -28,10 +53,8 @@ ignore_missing_imports = true
 name = "pruna"
 version = "0.2.5"
 description = "Smash your AI models"
-authors = [
-    {name = "Pruna AI", email = "hello@pruna.ai"}
-]
-license = {text = "All Rights Reserved"}
+authors = [{ name = "Pruna AI", email = "hello@pruna.ai" }]
+license = { file = "LICENSE" }
```
GreatBahram: I wonder if this makes sense to you?

.github/workflows/documentation.yaml
```diff
@@ -18,32 +18,26 @@ jobs:
...
 
-    
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v5
```
GreatBahram, line 26: > [!NOTE]
> These were a bit old, so I updated them. It would be nice to use `renovatebot` or `dependabot`; probably after this PR; we can work on that

pyproject.toml
johnrachwan123: makes sense :)

pyproject.toml
johnrachwan123, line 1: I was wondering why there is a lot of formatting changes here?

GreatBahram: I use this extension which auto-formats toml files [[link](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml)] there is also this tool [[link](https://taplo.tamasfe.dev/cli/introduction.html)]; in case it's not good; I can undo it

GreatBahram:
It seems to me that we don't need to update any documentation files, as they are quite general and not tied specifically to poetry. What do you think?  @johnrachwan123

pyproject.toml
davidberenstein1957: @GreatBahram would you be able to undo them for clarity in the PR?

davidberenstein1957:
Hi @GreatBahram, the PR looks good. Could you change it in such a way that changes in the github files also trigger tests? Or could you create a small commit to "fake" trigger the tests, which we could revert after.

pyproject.toml
GreatBahram: Did it here
https://github.com/PrunaAI/pruna/pull/165/commits/60a8d7a4965b5fc8344fb2f024996f3c268882b3

GreatBahram:
> Hi @GreatBahram, the PR looks good. Could you change it in such a way that changes in the github files also trigger tests? Or could you create a small commit to "fake" trigger the tests, which we could revert after.

I created a placeholder file in [this commit](https://github.com/PrunaAI/pruna/pull/165/commits/2aed34e6ff55da6da0a3cd8febde3752a359c9c4); let's see if that helps. If not, we can add `workflow_dispatch:` to the workflows and trigger them manually. Let me know if you need anything else. 🌻

**Update**: It seems most of workflows run only if they're target main branch

davidberenstein1957:
Hi @GreatBahram could you resolve the merge conflicts in the pyproject.toml?

GreatBahram:
> Hi @GreatBahram could you resolve the merge conflicts in the pyproject.toml?

Working on it and before merging it let's ensure we remove this file `src/pruna/alaki.py`

GreatBahram:
Would it be possible to allow the pipeline to run without requiring manual approval? Right now, it blocks the workflow unnecessarily wdyt? I believe it would be more effective for the reviewer to see that everything works, rather than having to approve it only for the pipeline to fail later. what do you think?

@davidberenstein1957

src/pruna/alaki.py
GreatBahram, line 1: :+1:

pyproject.toml
```diff
@@ -18,121 +18,100 @@ convention = "numpy"
 [tool.ruff.lint.flake8-copyright]
 author = "- Pruna AI GmbH"
 
+[tool.uv]
+extra-index-url = ["https://prunaai.pythonanywhere.com/"]
+
```
davidberenstein1957: Why are we adding this?

pyproject.toml
```diff
@@ -18,121 +18,100 @@ convention = "numpy"
...
 stable-fast = [
-    "xformers",
-    "stable-fast-pruna",
+    "xformers==0.0.30",
+    "stable-fast-pruna==1.0.7",
 ]
-# Note: You must first install the base package with ``pip install pruna`` before installing the GPTQ extension with ``pip install pruna[gptq]``
 gptq = [
-    "gptqmodel",
+    "gptqmodel==4.0.0.dev0+cu126torch2.7",
```
davidberenstein1957: I think this need to be more general

pyproject.toml
```diff
@@ -18,121 +18,100 @@ convention = "numpy"
...
+    "llmcompressor",
+]
 
-[tool.poetry.extras]
+[project.optional-dependencies]
 stable-fast = [
-    "xformers",
-    "stable-fast-pruna",
+    "xformers==0.0.30",
+    "stable-fast-pruna==1.0.7",
```
davidberenstein1957: we don't want to pin these versions

pyproject.toml
```diff
@@ -18,121 +18,100 @@ convention = "numpy"
...
     "tokenicer",
     "threadpoolctl",
     "device-smi",
     "random-word",
 ]
 full = [
-    "xformers",
-    "stable-fast-pruna",
+    "xformers==0.0.30",
+    "stable-fast-pruna==1.0.7",
```
davidberenstein1957: same here, no version pins please

pyproject.toml
```diff
@@ -145,7 +124,7 @@ dev = [
     "numpydoc-validation",
 ]
 tests = [
-    "pytest",
+    "pytest==7.4.4",
```
davidberenstein1957: version pin

pyproject.toml
```diff
@@ -18,121 +18,100 @@ convention = "numpy"
...
-numpydoc-validation = { version = "*", optional = true }
-mypy = { version = "*", optional = true }
-types-PyYAML = { version = "*", optional = true }
-gptqmodel = { version = "==4.0.0.dev0+cu126torch2.7", optional = true , markers = "extra == 'gptq'"}
-logbar = { version = "*", optional = true, markers = "extra == 'gptq'" }
-tokenicer = { version = "*", optional = true, markers = "extra == 'gptq'" }
-threadpoolctl = { version = "*", optional = true, markers = "extra == 'gptq'" }
-device-smi = { version = "*", optional = true, markers = "extra == 'gptq'" }
-random-word = { version = "*", optional = true, markers = "extra == 'gptq'" }
+dependencies = [
```
davidberenstein1957: can you check alignment of version pins with the original here too?

davidberenstein1957:
> Would it be possible to allow the pipeline to run without requiring manual approval? Right now, it blocks the workflow unnecessarily wdyt? I believe it would be more effective for the reviewer to see that everything works, rather than having to approve it only for the pipeline to fail later. what do you think?
> 
> @davidberenstein1957

@GreatBahram, I had also asked about this when I just joined, and I agree it can be perceived as blocking, but we do this to ensure we don't use any unnecessary resources.

GreatBahram:
> > Would it be possible to allow the pipeline to run without requiring manual approval? Right now, it blocks the workflow unnecessarily wdyt? I believe it would be more effective for the reviewer to see that everything works, rather than having to approve it only for the pipeline to fail later. what do you think?
> > @davidberenstein1957
> 
> @GreatBahram, I had also asked about this when I just joined, and I agree it can be perceived as blocking, but we do this to ensure we don't use any unnecessary resources.

I appreciate your input! Personally, I don’t see this as a best practice when the pipeline cost is low. In our case, I think running it earlier would’ve reduced the back-and-forth. Unfortunately, this approach creates a long feedback loop — even after resolving conflicts and rebasing, I’m not sure that’s enough. That said, it’s your call in the end :sunflower:

davidberenstein1957:
I agree with you. Just looping in @sharpenb and @johannaSommer  to see what they think.

pyproject.toml
GreatBahram: Hi David,

It's coming from here:
https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L40-L43

It configures an additional Python package source (repository) for the project

pyproject.toml
GreatBahram: I feel there was an issue and then @johnrachwan123 pinned this to resolve the issue, but it's coming from main branch

https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L115

pyproject.toml
GreatBahram: I guess this is the difference between `poetry` and `uv`. In poetry you can define a dependency once and re-use it multiple times in different extras, in `uv` on the other hand, you gotta be specific and define it each time.

Let me explain it further; in the `poetry` configuration we have it like this:
https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L109

and then we are using it two times, in both extra cases when ask; we install  a specific version which is specified above ^:
https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L124-L128
https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L138-L141

uv distincts the main dependencies from the extras which means you need to specify the version each time you need:
https://github.com/PrunaAI/pruna/blob/60a8d7a4965b5fc8344fb2f024996f3c268882b3/pyproject.toml#L100-L103
https://github.com/PrunaAI/pruna/blob/60a8d7a4965b5fc8344fb2f024996f3c268882b3/pyproject.toml#L112-L115

pyproject.toml
GreatBahram: Same as [above](https://github.com/PrunaAI/pruna/pull/165/files#r2128162602):

https://github.com/PrunaAI/pruna/blob/01a57b2ada6b56893fda0c232584955b7b8912ed/pyproject.toml#L106

pyproject.toml
GreatBahram: save as [here](https://github.com/PrunaAI/pruna/pull/165#discussion_r2135294860)

pyproject.toml
```diff
@@ -102,7 +102,8 @@ stable-fast = [
     "stable-fast-pruna==1.0.7",
 ]
 gptq = [
-    "gptqmodel==4.0.0.dev0+cu126torch2.7",
+    "gptqmodel==4.0.0.dev0+cu126torch2.7; sys_platform != 'darwin' or platform_machine != 'arm64'",
```
GreatBahram, line 110: @davidberenstein1957 could you please double check to ensure this is correct :pray:

pyproject.toml
```diff
@@ -18,126 +18,101 @@ convention = "numpy"
...
-    "xformers",
-    "stable-fast-pruna",
+    "xformers==0.0.30",
+    "stable-fast-pruna==1.0.7",
 ]
-# Note: You must first install the base package with ``pip install pruna`` before installing the GPTQ extension with ``pip install pruna[gptq]``
 gptq = [
-    "gptqmodel",
+    "gptqmodel==4.0.0.dev0+cu126torch2.7; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "gptqmodel; sys_platform == 'darwin' and platform_machine == 'arm64'",
```
davidberenstein1957, line 111:
```suggestion
    "gptqmodel>=4.0.0; sys_platform == 'darwin' and platform_machine == 'arm64'",
```

pyproject.toml
davidberenstein1957: I belive this should do it :)

pyproject.toml
GreatBahram: Sorry I needed to undo this, as it didn't work on machine; I push the latest changes and let's what happens on the pipeline and from there we can make a better decision

GreatBahram: Here the pipeline failed [[link](https://github.com/PrunaAI/pruna/actions/runs/15566956517/job/43833250829)]

johnrachwan123:
@GreatBahram currently the linting test still fails. Any idea why ?

.github/workflows/linting.yaml
```diff
@@ -32,8 +32,8 @@ jobs:
         run: |
           uv sync --extra tests
 
-      - name: Remove .so files
-        run: find . -name "*.so" -delete
+      # - name: Remove .so files
```
GreatBahram: @johnrachwan123 Not sure if it's related to this or not :shrug:

GreatBahram: It seems [that](https://github.com/PrunaAI/pruna/pull/165/commits/1245fcd6e89f745090ce373be1f0b6705bbe4bec) was it; it's in your hands now @johnrachwan123

PR #167, opened 2025-06-02
fix: add size check when limiting the dataset

begumcig:
## Description

`limit_datasets` in PrunaDataModule doesn't check the size of the datasets before limiting the size, resulting in IndexErrors: 

```
E           IndexError: Index 9 out of range for dataset of size 1. 
```
## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/data/pruna_datamodule.py
+++ b/src/pruna/data/pruna_datamodule.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 
 import inspect
 from functools import partial
-from typing import Callable, List, Tuple, Union
+from typing import Callable, List, Tuple, Union, cast
 
 from datasets import Dataset, IterableDataset
 from pytorch_lightning import LightningDataModule
@@ -189,16 +189,22 @@ class PrunaDataModule(LightningDataModule):
             train_limit, val_limit, test_limit = limit
 
         if isinstance(self.train_dataset, Dataset):
+            self.train_dataset = cast(Dataset, self.train_dataset)  # for mypy
+            self.val_dataset = cast(Dataset, self.val_dataset)
+            self.test_dataset = cast(Dataset, self.test_dataset)
+
             self.train_dataset = self.train_dataset.select(range(min(len(self.train_dataset), train_limit)))
             self.val_dataset = self.val_dataset.select(range(min(len(self.val_dataset), val_limit)))  # type: ignore[union-attr]
             self.test_dataset = self.test_dataset.select(range(min(len(self.test_dataset), test_limit)))  # type: ignore[union-attr]
         elif isinstance(self.train_dataset, IterableDataset):
-            train_length = sum(1 for _ in self.train_dataset)
-            val_length = sum(1 for _ in self.val_dataset)
-            test_length = sum(1 for _ in self.test_dataset)
-            self.train_dataset = self.train_dataset[: min(train_length, train_limit)]
-            self.val_dataset = self.val_dataset[: min(val_length, val_limit)]
-            self.test_dataset = self.test_dataset[: min(test_length, test_limit)]
+            self.train_dataset = cast(IterableDataset, self.train_dataset)  # for mypy
+            self.val_dataset = cast(IterableDataset, self.val_dataset)
+            self.test_dataset = cast(IterableDataset, self.test_dataset)
+
+            self.train_dataset = self.train_dataset.take(train_limit)
+            self.val_dataset = self.val_dataset.take(val_limit)
+            self.test_dataset = self.test_dataset.take(test_limit)
+
         elif isinstance(self.train_dataset, TorchDataset):
             train_indices = list(range(min(len(self.train_dataset), train_limit)))  # type: ignore[arg-type]
             val_indices = list(range(min(len(self.val_dataset), val_limit)))  # type: ignore[arg-type]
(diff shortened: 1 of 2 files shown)


src/pruna/data/pruna_datamodule.py
```diff
@@ -189,13 +189,16 @@ def limit_datasets(self, limit: int | list[int] | tuple[int, int, int]) -> None:
...
+            self.train_dataset = self.train_dataset.select(range(min(len(self.train_dataset), train_limit)))
+            self.val_dataset = self.val_dataset.select(range(min(len(self.val_dataset), val_limit)))  # type: ignore[union-attr]
+            self.test_dataset = self.test_dataset.select(range(min(len(self.test_dataset), test_limit)))  # type: ignore[union-attr]
         elif isinstance(self.train_dataset, IterableDataset):
-            self.train_dataset = self.train_dataset[:train_limit]
-            self.val_dataset = self.val_dataset[:val_limit]
-            self.test_dataset = self.test_dataset[:test_limit]
+            train_length = sum(1 for _ in self.train_dataset)
+            val_length = sum(1 for _ in self.val_dataset)
+            test_length = sum(1 for _ in self.test_dataset)
```
GreatBahram: I wonder why not using `len` :thinking: 
```suggestion
            train_length = len(self.train_dataset)
            val_length = len(self.val_dataset)
            test_length = len(self.test_dataset)
```

GreatBahram: If use this approach, then they're small enough to put them directly into next lines

davidberenstein1957: I believe we need this because it is an iterable dataset.

src/pruna/data/pruna_datamodule.py
```diff
@@ -189,13 +189,25 @@ def limit_datasets(self, limit: int | list[int] | tuple[int, int, int]) -> None:
...
+
+            self.train_dataset = self.train_dataset.select(range(min(len(self.train_dataset), train_limit)))
+            self.val_dataset = self.val_dataset.select(range(min(len(self.val_dataset), val_limit)))  # type: ignore[union-attr]
+            self.test_dataset = self.test_dataset.select(range(min(len(self.test_dataset), test_limit)))  # type: ignore[union-attr]
         elif isinstance(self.train_dataset, IterableDataset):
-            self.train_dataset = self.train_dataset[:train_limit]
-            self.val_dataset = self.val_dataset[:val_limit]
-            self.test_dataset = self.test_dataset[:test_limit]
+            self.train_dataset = cast(IterableDataset, self.train_dataset)  # for mypy
+            self.val_dataset = cast(IterableDataset, self.val_dataset)
```
gsprochette, line 201: Doing this avoids relying on the len method which is not implemented for IterableDatasets, and also avoids generating all the samples which the user is probably trying to avoid if using an IterableDataset. 
```suggestion
            self.train_dataset = self.train_dataset.take(train_limit)
            self.val_dataset = self.val_dataset.take(val_limit)
            self.test_dataset = self.test_dataset.take(test_limit)
```

begumcig: That's amazing Gaspar!! Thank you so much, changing what I have :)

tests/data/test_datamodule.py
```diff
@@ -61,3 +64,45 @@ def test_dm_from_dataset(setup_fn: Callable, collate_fn: Callable, collate_fn_ar
...
+        raise ValueError("Unsupported dataset type")
+
+    def dummy_collate(batch):
+        return batch
+
+    datamodule = PrunaDataModule(
+        train_ds=train_ds,
+        val_ds=val_ds,
+        test_ds=test_ds,
+        collate_fn=dummy_collate,
```
GreatBahram, line 91: `collate_fn=lambda batch: batch`

tests/data/test_datamodule.py
```diff
@@ -61,3 +64,45 @@ def test_dm_from_dataset(setup_fn: Callable, collate_fn: Callable, collate_fn_ar
...
+
+    # Create small datasets
+    if ds_type == "hf":
+        data = {"text": ["a", "b", "c", "d", "e"]}
+        train_ds = val_ds = test_ds = Dataset.from_dict(data)
+    elif ds_type == "torch":
+        x = torch.arange(5).float().unsqueeze(1)
+        y = torch.arange(5)
+        train_ds = val_ds = test_ds = TensorDataset(x, y)
+    else:
```
GreatBahram, line 81: I guess we can remove this `else` as we specify the `ds_type` :thinking: , wdyt?

gsprochette: I think the current code is fine since this is in the tests, we may want to add more dataset types later on and this will make sure that the test fails if we add a dataset in the test parameters but forget to add logic for it. It's not the cleanest option but it's the clear and safe one, right?

GreatBahram: Yes, totally agree :rocket:

tests/data/test_datamodule.py
```diff
@@ -61,3 +64,45 @@ def test_dm_from_dataset(setup_fn: Callable, collate_fn: Callable, collate_fn_ar
 
     # iterate through the dataloaders
     iterate_dataloaders(datamodule)
+@pytest.mark.cpu
+@pytest.mark.parametrize("limit_len", [2, (2, 3, 1), (10, 10, 10)])
+@pytest.mark.parametrize("ds_type", ["hf", "torch"])
```
gsprochette, line 69: Could we have a test for the IterableDataset case?

PR #169, opened 2025-06-03
fix: 168 bug device placement does not work with torchmetric

### LINKED ISSUE #168: [BUG] device placement does not work with `TorchMetric`
## Describe the bug

Because you cannot initialize a torchmetric using a device attribute, you can only set it using .to()

## What I did

```python
task = Task(request=['perplexity'], datamodule=datamodule)
```

## Environment
 - pruna version: 
 - python version: 
 - Operating System:

davidberenstein1957:
## Description

There was a tiny error with pro-active device assignment in case of `TorchMetric` classes.

## Related Issue

Fixes #168 

## Type of Change

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/pruna/engine/load.py
+++ b/src/pruna/engine/load.py
@@ -475,6 +475,8 @@ def filter_load_kwargs(func: Callable, kwargs: dict) -> dict:
     """
     Filter out keyword arguments that cannot be passed to the given function.
 
+    Only filters if the function does not accept arbitrary keyword arguments.
+
     Parameters
     ----------
     func : Callable
@@ -489,9 +491,15 @@ def filter_load_kwargs(func: Callable, kwargs: dict) -> dict:
     """
     # Get the function's signature
     signature = inspect.signature(func)
-    valid_params = set(signature.parameters.keys())
 
-    # Filter valid and invalid kwargs
+    # Check if function accepts arbitrary kwargs
+    has_kwargs = any(param.kind == inspect.Parameter.VAR_KEYWORD for param in signature.parameters.values())
+
+    if has_kwargs:
+        return kwargs
+
+    # Only filter if function doesn't accept arbitrary kwargs
+    valid_params = set(signature.parameters.keys())
     valid_kwargs = {k: v for k, v in kwargs.items() if k in valid_params}
     invalid_kwargs = {k: v for k, v in kwargs.items() if k not in valid_params}
 
--- a/src/pruna/evaluation/metrics/metric_torch.py
+++ b/src/pruna/evaluation/metrics/metric_torch.py
@@ -219,6 +219,7 @@ class TorchMetricWrapper(StatefulMetric):
         if metric_name == "clip_score" and call_type.startswith(PAIRWISE):
             from pruna.evaluation.metrics.metric_pairwise_clip import PairwiseClipScore
 
+            kwargs.pop("device", None)
             return PairwiseClipScore(**kwargs)
         return super().__new__(cls)
 
@@ -233,8 +234,9 @@ class TorchMetricWrapper(StatefulMetric):
         """
         super().__init__()
         try:
+            device = kwargs.pop("device", None)
             if metric_name == "perplexity":
-                device = set_to_best_available_device(device=kwargs.get("device"))
+                device = set_to_best_available_device(device=device)
                 self.metric = TorchMetrics[metric_name](**kwargs).to(device)
             else:
                 self.metric = TorchMetrics[metric_name](**kwargs)
--- a/src/pruna/evaluation/metrics/metric_torch.py
+++ b/src/pruna/evaluation/metrics/metric_torch.py
@@ -14,6 +14,7 @@
 
 from __future__ import annotations
 
+from contextlib import suppress
 from enum import Enum
 from functools import partial
 from typing import Any, Callable, List, Optional, Union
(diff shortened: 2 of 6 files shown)


src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -233,8 +234,9 @@ def __init__(self, metric_name: str, call_type: str = "", **kwargs) -> None:
         """
         super().__init__()
         try:
+            device = kwargs.pop("device", None)
             if metric_name == "perplexity":
-                device = set_to_best_available_device(device=kwargs.get("device"))
+                device = set_to_best_available_device(device=device)
                 self.metric = TorchMetrics[metric_name](**kwargs).to(device)
```
begumcig: Do you think it makes sense to check if other metrics also have a device attribute? Something that checks if `self.metric` has a device attribute, then tries to set this device and if that fails tries to use` .to()` to move to the device?

src/pruna/evaluation/metrics/registry.py
```diff
@@ -105,7 +106,12 @@ def get_metric(cls, name: str, **kwargs) -> BaseMetric | StatefulMetric:
         """
         if name not in cls._registry:
             raise ValueError(f"Metric '{name}' is not registered.")
-        return cls._registry[name](**kwargs)
+
+        metric_cls = cls._registry[name]
+        if isinstance(metric_cls, partial) and "TorchMetricWrapper" in metric_cls.func.__name__:
+            return metric_cls(**kwargs)
+        else:
+            return metric_cls(**filter_load_kwargs(metric_cls, kwargs))
```
begumcig, line 114: Why do we need the filtering here? As we have other metrics that are wrappers and will probably have more in the future, I don't think we should use the filtering function unless we need it :)

tests/evaluation/test_task.py
```diff
@@ -0,0 +1,33 @@
...
+from pruna.data import base_datasets
+from pruna.evaluation.metrics.metric_torch import TorchMetrics
+from torchmetrics.classification import Accuracy, Precision, Recall
+from functools import partial
+
+@pytest.fixture(autouse=True)
+def _mock_torch_metrics():
+    """Mock TorchMetrics enum values for accuracy, perplexity and recall only for this test."""
+    def make_mock_metric(metric_class):
+        metric = partial(metric_class, task="binary")
```
begumcig, line 15: Thank you so much for adding a test for the Task!!

begumcig (changes requested):
Thank you so much for fixing this super fast!! Suggested a change to possibly make the device casting more agnostic for all torchmetrics, looks really good overall! Love the Task test!

src/pruna/evaluation/metrics/registry.py
davidberenstein1957: I think it is a nice safety measure to ensure nothing breaks. We only filter the kwargs and warn if a wrong one has be passed without breaking the flow, right?

src/pruna/evaluation/metrics/metric_torch.py
davidberenstein1957: @begumcig you mean other torch metrics?

src/pruna/evaluation/metrics/registry.py
begumcig: Yes but we have other metrics that are wrappers that might have additional arguments that need to be passed down.  For instance I can see this filtering breaking pairwise_clip_score and any future wrapper metric :)

src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -233,11 +235,12 @@ def __init__(self, metric_name: str, call_type: str = "", **kwargs) -> None:
         """
         super().__init__()
         try:
-            if metric_name == "perplexity":
-                device = set_to_best_available_device(device=kwargs.get("device"))
-                self.metric = TorchMetrics[metric_name](**kwargs).to(device)
-            else:
-                self.metric = TorchMetrics[metric_name](**kwargs)
+            device = kwargs.pop("device", None)
```
begumcig, line 238: Thanks!

begumcig (changes requested):
One tiny change on filtering :) Almost there!

src/pruna/evaluation/metrics/registry.py
davidberenstein1957: Ahh yes, I see what you mean. I misundertood what the function was doing. Anyhow, I changed the underlying function of `filter_load_kwargs` to make sure **kwargs in the signature allows passing all of the kwargs passed through it. Let me know what you think.

johannaSommer (approved):

src/pruna/engine/load.py
```diff
@@ -493,9 +495,15 @@ def filter_load_kwargs(func: Callable, kwargs: dict) -> dict:
     """
     # Get the function's signature
     signature = inspect.signature(func)
-    valid_params = set(signature.parameters.keys())
 
-    # Filter valid and invalid kwargs
+    # Check if function accepts arbitrary kwargs
+    has_kwargs = any(param.kind == inspect.Parameter.VAR_KEYWORD for param in signature.parameters.values())
```
begumcig, line 496: Should we also add a check for if the function has *args with VAR_POSITIONAL for when we need to pass positional arguments?

begumcig (approved):

PR #171, opened 2025-06-03
fix: enhance device handling in set_to_best_available_device function

davidberenstein1957:
- Improved support for CUDA and MPS devices by allowing device indices (e.g., "cuda:0").
- Added validation for CUDA device indices to ensure they are within available range, with appropriate warnings for invalid requests.
- Updated MPS handling to always return "mps:0" since it only supports one device.
- Introduced a new test to verify device handling with indices, ensuring fallback behavior works as expected.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -373,6 +373,37 @@ def determine_dtype(pipeline: Any) -> torch.dtype:
     return torch.float32
 
 
+def _resolve_cuda_device(device: str) -> str:
+    """
+    Resolve CUDA device string to a valid CUDA device.
+
+    Parameters
+    ----------
+    device : str
+        CUDA device string (e.g. "cuda", "cuda:0", "cuda:1")
+
+    Returns
+    -------
+    str
+        Valid CUDA device string
+    """
+    if not torch.cuda.is_available():
+        pruna_logger.warning("'cuda' requested but not available.")
+        return set_to_best_available_device(device=None)
+
+    device_idx = device.split(":")[1] if ":" in device else "0"
+    try:
+        idx = int(device_idx)
+        if idx >= torch.cuda.device_count():
+            pruna_logger.warning(f"CUDA device {idx} not available, using device 0")
+            return "cuda:0"
+        torch.cuda.get_device_properties(idx)
+        return f"cuda:{idx}"
+    except (ValueError, AssertionError):
+        pruna_logger.warning(f"Invalid CUDA device index: {device_idx}")
+        return "cuda:0"
+
+
 def set_to_best_available_device(device: str | torch.device | None) -> str:
     """
     Set the device to the best available device.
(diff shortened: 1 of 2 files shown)


src/pruna/engine/utils.py
```diff
@@ -277,17 +277,32 @@ def set_to_best_available_device(device: str | torch.device | None) -> str:
             raise ValueError("'accelerate' requested but neither CUDA nor MPS is available.")
         return "accelerate"
 
-    if device == "cuda":
+    # Handle cuda devices (e.g. "cuda", "cuda:0", "cuda:1")
+    if device.startswith("cuda"):
```
GreatBahram: Nice, just a small NIT: I wonder if it makes sense to put the logic for cuda into a small function like `_resolve_cuda_device` and simply return it here?

Ideally, I guess this function can be further improved using a match-case, something like:

```
def set_to_best_available_device(device: str | torch.device | None) -> str:
    if isinstance(device, torch.device):
        device = str(device)

    match device:
        case None:
            ...
        case "cpu":
            ...
        case "accelerate":
            ...
        case d if d.startswith("cuda"):
            ...
        case d if d.startswith("mps"):
            ...
        case _:
            raise ValueError(f"Device not supported: '{device}'")
```

begumcig (approved):
LGTM!!!! Let's gooo!

src/pruna/engine/utils.py
davidberenstein1957: We rely oin Python 3.9 so we cannot do this yet.

GreatBahram: Oh, good to know, David, hopefully after merging the uv PR, we can tackle this :crossed_fingers: 

But I guess you can still introduce the `_resolve_cuda_device`  function if you want; then later on we get back and use the match-case to improve the code

tests/engine/test_device.py
```diff
@@ -39,3 +39,15 @@ def test_device_available(device: str | torch.device, expected: str) -> None:
     smash_config = SmashConfig(device=device)
     assert smash_config.device == expected
 
+@pytest.mark.cuda
```
GreatBahram: this is cool; I just wonder what would happen if both are not available :thinking:  the test might fail or pass misleadingly.

If you want to improve it; probably we can mock the underlying layers, something like this (not test though):

```
from unittest.mock import patch
import pytest
import torch

@pytest.mark.parametrize(
    "device, expected, cuda_available, mps_available",
    [
        ("mps:0", "mps:0", False, True),
        ("cuda:0", "cuda:0", True, False),
    ],
)
def test_device_with_index(device, expected, cuda_available, mps_available):
    with patch("torch.cuda.is_available", return_value=cuda_available), \
         patch("torch.backends.mps.is_available", return_value=mps_available), \
         patch("torch.cuda.device_count", return_value=1), \
         patch("torch.cuda.get_device_properties"):
        
        assert set_to_best_available_device(device) == expected
```

GreatBahram (approved):
LGTM :rocket:

PR #181, opened 2025-06-06
test: connect inference/eval tests to algorithms

begumcig:
## Description

Adding simple evaluation runs to the algorithm tests to check if model is capable of running inference, and the inference outputs are compatible with the evaluation agent.
Small bug fixes here and there to get the tests up and running

## Type of Change

## Checklist


src/pruna/engine/handler/handler_utils.py
```diff
@@ -73,6 +74,16 @@ def scan_for_exceptions(model: Any) -> InferenceHandler | None:
     # this avoids directly importing external packages
     for handler, model_classes in HANDLER_EXCEPTIONS.items():
         for model_class in model_classes:
-            if model_class in model.__class__.__name__:
+            if model.__class__.__name__ == "OptimizedModule":  # torch_compile abstracts over the model class.
+                if hasattr(model, "_orig_mod"):
+                    name = model._orig_mod.__class__.__name__
+                else:
+                    pruna_logger.warning(
```
johannaSommer: can we even continue in this case? it seems the name is not defined and it will fail down the line?

johannaSommer: if you feel like it, we could change this to just call the function recursively on the orig model:
```python
def get_handler(model, args):
 if model.__class__.__name__ == "OptimizedModule":
    get_handler(model._orig_model, args)
 # continue logic
```
I feel like this sometimes can make it easier to understand that we are doing the exact same thing just on some other attribute

src/pruna/engine/utils.py
```diff
@@ -111,7 +111,10 @@ def move_to_device(model: Any, device: str | torch.device, raise_error: bool = F
             else:
                 pruna_logger.warning(f"Could not move model to device: {str(e)}")
     elif hasattr(model, "task") and getattr(model, "task") == "automatic-speech-recognition":
-        model.model.to(device)
+        try:
```
johannaSommer: i think i fixed this in the multi gpu PR so i think we can discard this when rebasing (just as a heads up)

src/pruna/evaluation/metrics/metric_pairwise_clip.py
```diff
@@ -51,7 +51,13 @@ def __init__(self, **kwargs: Any) -> None:
         if "call_type" in kwargs:
             pruna_logger.error(f"Call type is not supported for {self.metric_name}. Using default call type {PAIRWISE}")
             kwargs.pop("call_type")
-        super().__init__(**kwargs)
+        if "device" in kwargs:
```
johannaSommer, line 54: can you put a bit of commentary what's happening here? couldn't quite parse it on the first take

tests/algorithms/testers/compilation.py
```diff
@@ -13,6 +13,8 @@ class TestTorchCompile(AlgorithmTesterBase):
     reject_models = []
     allow_pickle_files = False
     algorithm_class = TorchCompileCompiler
+    metrics = ["latency"]
+    hyperparameters = {"torch_compile_fullgraph": False}
```
johannaSommer: @nifleisch @johnrachwan123 if we require the fullgraph this often, should we consider making this the new default?

tests/common.py
```diff
@@ -70,6 +70,7 @@ def run_full_integration(algorithm_tester: Any, device: str, model_fixture: tupl
         load_kwargs = algorithm_tester.check_loading_dtype(model)
         model = algorithm_tester.cast_to_device(model, device=smash_config["device"])
         smashed_model = algorithm_tester.execute_smash(model, smash_config)
+        algorithm_tester.execute_evaluation(smashed_model, smash_config.data, smash_config["device"])
```
johannaSommer: really just picking your brain here, does it make sense that we execute evaluation again after the loading? I can see some things going wrong when reloading a model so it would be cool to check it, however, i can also see this might be overkill

johannaSommer (changes requested):
I cant believe you did this and its almost flawless on the first take, i really appreciate the effort you put into this, it's so key! just a bit of comments and logic changes then we're already GTG

src/pruna/engine/handler/handler_utils.py
begumcig: If the name is not defined it will default to the standard handler, which might still work if the model doesn't require special inference handling. That's why I thought giving a warning and continuing might be better, what do you think?

src/pruna/evaluation/metrics/metric_pairwise_clip.py
begumcig: Yes, I can see how this is super confusing. With the accelerate PR device casting for metrics are hopefully will be more understandable, but also added commentary here. Lmk if it's clearer now!

johannaSommer (approved):
🚔

tests/algorithms/testers/compilation.py
nifleisch: Agreed. `fullgraph=False` is also the default torch.compile uses so it would be more convenient to match this default. Especially since `fullgraph=True` can lead to graph breaks for certain models and is less stable.

nifleisch (approved):
I like the PR, approved! 🙌

johannaSommer (commented):
LGTM! 🫡

PR #183, opened 2025-06-06
feat: add arniqa

begumcig:
## Description

This PR adds support for the arniqa image quality metric. Since ARNIQA uses an internal dataset, it now runs with the new call_type="y" (output-only).

## Type of Change

## Checklist

### DIFF
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -148,8 +148,9 @@ Understanding Call Types
 | `pairwise_y_gt`    | Base model's output first, then subsequent model's output   |
 +--------------------+-------------------------------------------------------------+
 | `pairwise_gt_y`    | Subsequent model's output first, then base model's output   |
-+--------------------+-------------------------------------------------------------+
-
++--------------------+-------------------------------------------------------------+ 
+| `y`                | Only the output is used, the metric has an internal dataset |
++--------------------+-------------------------------------------------------------+ 
 
 You need to decide on the default ``call_type`` based on the metric you are implementing.
 
@@ -158,6 +159,8 @@ For example, if you are implementing a metric that compares two models, you shou
 If you are implementing an alignment metric comparing model's output with the input, you should use the ``x_gt`` or ``gt_x`` call type. Examples from |pruna| include ``clip_score``.
 
 If you are implementing a metric that compares the model's output with the ground truth, you should use the ``y_gt`` or ``gt_y`` call type. Examples from |pruna| include ``fid``, ``cmmd``, ``accuracy``, ``recall``, ``precision``.
+
+If you are wrapping an Image Quality Assessment (IQA) metric, that has an internal dataset, you should use the ``y`` call type. Examples from |pruna| include ``arniqa``.
 
 You may want to switch the mode of the metric despite your default ``call_type``. For instance you may want to use ``fid`` in pairwise mode to get a single comparison score for two models.
 
--- a/docs/user_manual/evaluate.rst
+++ b/docs/user_manual/evaluate.rst
@@ -216,7 +216,8 @@ The ``EvaluationAgent`` accepts ``Metrics`` in three ways:
 Metric Call Types
 ~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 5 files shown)


src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -350,6 +367,9 @@ def get_call_type(call_type: str, metric_name: str) -> str:
         # If the call type and default call type match, we use the default call type.
         if TorchMetrics[metric_name].call_type.startswith(PAIRWISE):
             return TorchMetrics[metric_name].call_type
+        elif TorchMetrics[metric_name].call_type == "y":
+            pruna_logger.warning("IQA metrics cannot be used with pairwise call type. Using default call type.")
```
davidberenstein1957: Should we mention the default call type in the log?

tests/evaluation/test_torch_metrics.py
```diff
@@ -61,3 +61,11 @@ def test_torch_metrics(dataloader_fixture: Any, metric: str) -> None:
     _, gt = next(iter(dataloader_fixture))
     metric.update(gt, gt, gt)
     assert metric.compute().result == 1.0
+
+@pytest.mark.cpu
+@pytest.mark.parametrize("dataloader_fixture", ["LAION256"], indirect=True)
+def test_arniqa(dataloader_fixture: Any) -> None:
+    """Test arniqa."""
+    metric = TorchMetricWrapper("arniqa")
```
davidberenstein1957, line 70: should we add test for the default calltype error too?

docs/user_manual/customize_metric.rst
```diff
@@ -149,7 +149,8 @@ Understanding Call Types
 +--------------------+-------------------------------------------------------------+
 | `pairwise_gt_y`    | Subsequent model's output first, then base model's output   |
 +--------------------+-------------------------------------------------------------+ 
-
+| `y`                | Only the output is used, the metric has an internal dataset |
++--------------------+-------------------------------------------------------------+
```
davidberenstein1957, line 153: @begumcig I think the behaviour with `single` vs `pairwise` might need to be updated too here?

https://docs.pruna.ai/en/stable/docs_pruna/user_manual/evaluate.html#internal-call-types

begumcig: I have added the call type y under there, was this review before I added this or do you mean another addition :)

docs/user_manual/evaluate.rst
```diff
@@ -216,9 +216,9 @@ The ``EvaluationAgent`` accepts ``Metrics`` in three ways:
 Metric Call Types
 ~~~~~~~~~~~~~~~~~
 
-|pruna| metrics can operate in both single-model and pairwise modes.
+All |pruna| stateful metrics besides Image Quality Assessment (IQA) metrics can operate in both single-model and pairwise modes.
```
sharpenb: Could we say "Stateful metrics can generally be evaluated in single-model and pariwise modes. Single-mode allows to compare a mode against ground-truth data, while pairwise mode allows to compare the fidelity of model against another model. "

src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -353,6 +370,9 @@ def get_call_type(call_type: str, metric_name: str) -> str:
         # If the call type and default call type match, we use the default call type.
         if TorchMetrics[metric_name].call_type.startswith(PAIRWISE):
             return TorchMetrics[metric_name].call_type
+        elif TorchMetrics[metric_name].call_type == "y":
+            pruna_logger.warning("IQA metrics cannot be used with pairwise call type. Using default call type y.")
+            return TorchMetrics[metric_name].call_type
```
sharpenb: My suggestion would be to return an error here instead of silently fixing the problem :)

sharpenb (approved):
That looks very nice! If you can just fix the two comments before merging that would be perfect :)

src/pruna/evaluation/metrics/metric_torch.py
begumcig: Makes so much sense! I also moved throwing this error to the general utils file where i check call types for all metrics, so we would throw this error not just for tochmetrics but for all of our metrics!

PR #205, opened 2025-06-13
fix: device placement with indexed devices

davidberenstein1957:
## Description

Fixes error in nightly tests https://github.com/PrunaAI/prunatree/actions/runs/15632633387/job/44040260688
These errors started after my PR, which raised an error within the move_to_device method because it was not able to deal with devices that also contained indexes. I wanted to fix the test by adding this functionality.

## Related Issue

Fixes NA

## Type of Change

## Checklist


johnrachwan123:
bugbot run

src/pruna/engine/utils.py
```diff
@@ -135,9 +136,16 @@ def move_to_device(
...
 
-    # sanity check for expected device types
-    if device not in ["cpu", "cuda", "mps", "accelerate"]:
-        raise ValueError("Device must be a string in [cpu, cuda, mps, accelerate].")
+    # Convert string device to torch.device for consistent handling
+    if isinstance(device, str):
+        if device == "accelerate":
+            pass  # Handle accelerate separately
+        elif any(device.startswith(d) for d in ["cpu", "cuda", "mps"]):
+            device = torch.device(device)  # This handles "cuda:0" etc correctly
```
begumcig: is this to handle what `_resolve_cuda_device` is returning? I believe currently we return cuda:idx, where int is the index of the GPU.

src/pruna/engine/utils.py
```diff
@@ -396,17 +403,12 @@ def _resolve_cuda_device(device: str) -> str:
...
 
-    device_idx = device.split(":")[1] if ":" in device else "0"
     try:
-        idx = int(device_idx)
-        if idx >= torch.cuda.device_count():
-            pruna_logger.warning(f"CUDA device {idx} not available, using device 0")
-            return "cuda:0"
-        torch.cuda.get_device_properties(idx)
-        return f"cuda:{idx}"
+        torch.cuda.get_device_properties(device)
```
begumcig: i am a bit confused by this line, we are not assigning it to anything, what are we doing with it?

src/pruna/engine/utils.py
```diff
@@ -396,17 +403,12 @@ def _resolve_cuda_device(device: str) -> str:
...
+        return device
```
begumcig: Just to double check, this device string can still be cuda:0 cuda:1 right? We are not stripping it down to just "cuda".

src/pruna/engine/utils.py
```diff
@@ -107,6 +107,25 @@ def safe_is_instance(model: Any, instance_type: type) -> bool:
...
+        The normalized device type as a string.
+    """
+    if isinstance(device, torch.device):
+        return device.type
+    return device
+
+
 def move_to_device(
     model: Any,
     device: str | torch.device,
```
GreatBahram, line 113: Just wanted to mention here you can use Literal:

https://adamj.eu/tech/2021/07/09/python-type-hints-how-to-use-typing-literal/

src/pruna/engine/utils.py
davidberenstein1957: Yes, that is correct.

src/pruna/engine/utils.py
```diff
@@ -120,40 +141,44 @@ def move_to_device(
...
     # do not cast if the model is already on the correct device
-    if get_device(model) == device:
+    if _normalize_device_type(get_device(model)) == device_str:
         return
 
-    if device == "accelerate":
+    if device_str == "accelerate":
         if device_map is None:
-            raise ValueError("Device map is required when moving to accelerate.")
+            device_map = infer_auto_device_map(model)
```
johannaSommer: I would keep the Value Error here - the `move_to_device` function is something we only use internally, this isnt actually something the user should get in contact with. For accelerate-casted models in pruna, we assumed that they are already casted with accelerate when they are passed to `smash`, and we save the input device map.

src/pruna/engine/utils.py
```diff
@@ -120,40 +141,44 @@ def move_to_device(
...
+                "'accelerate' requested, but no device map provided. Automatically inferring device map..."
+            )
+            pruna_logger.info(f"Device map: {device_map}")
         cast_model_to_accelerate_device_map(model, device_map)
-
     else:
         if get_device(model) == "accelerate":
             remove_all_accelerate_hooks(model)
             # transformers model maintain single-device models with a None map, diffusers does not
-            model.hf_device_map = {"": "cpu" if device == "cpu" else 0}
```
johannaSommer, line 151: afaik "cuda" is not valid here, it needs to be the device index (0 / 1)

src/pruna/engine/utils.py
```diff
@@ -107,6 +107,27 @@ def safe_is_instance(model: Any, instance_type: type) -> bool:
     return isinstance(model, instance_type)
 
 
+def _normalize_device_type(device: str | torch.device | dict[str, str] | None) -> str | dict[str, str]:
```
johannaSommer: In order to actually enable the user to specify a device as torch.device("cuda:1") in the smash config, this would also have to be adjusted:
https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/utils.py#L429

johannaSommer: as far as i can see we would otherwise just "normalize" this to cuda:0

johannaSommer: I think this comparison also needs to be adjusted with the normalization:
https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/compatibility_checks.py#L106

could you please double check whether this one needs to be adjusted too? i think so
https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/compatibility_checks.py#L40

johannaSommer (changes requested):

src/pruna/engine/utils.py
```diff
@@ -391,35 +426,31 @@ def set_to_best_available_device(device: str | torch.device | None) -> str:
...
         return "accelerate"
-
-    if device == "cuda":
-        if not torch.cuda.is_available():
-            pruna_logger.warning("'cuda' requested but not available.")
-            return set_to_best_available_device(device=None)
-        return "cuda"
-
-    if device == "mps":
+    elif device_str.startswith("cuda"):
```
johannaSommer, line 490: I like the check you added for "mps not available", should we also ensure here that cuda is available and return the best device with `None` otherwise?

johannaSommer: nevermind its in the resolve function, for clarity i like it a bit better here but also okay with where it is placed now

src/pruna/engine/utils.py
```diff
@@ -374,6 +374,41 @@ def determine_dtype(pipeline: Any) -> torch.dtype:
...
+    str
+        Valid CUDA device string
+    """
+    if not torch.cuda.is_available():
+        pruna_logger.warning("'cuda' requested but not available.")
+        return set_to_best_available_device(device=None)
+
+    # If just "cuda", return as is
+    if device == "cuda":
+        return device
```
johannaSommer: I would almost vote for, if no index is specified, always returning "cuda:0", what do you think? I know for example that some other libraries like accelerate expect there to be an index present which is what i fixed in https://github.com/PrunaAI/pruna/pull/226

johannaSommer: same for the return statements below + if the user inputs "cuda" into the `set_best_available_device` I would also append a ":0". This way it is consistent that if a cuda device is specified, we also always give the index

src/pruna/engine/utils.py
```diff
@@ -108,48 +108,51 @@ def safe_is_instance(model: Any, instance_type: type) -> bool:
...
             raise ValueError("Device map is required when moving to accelerate.")
         cast_model_to_accelerate_device_map(model, device_map)
-
     else:
         if get_device(model) == "accelerate":
             remove_all_accelerate_hooks(model)
             # transformers model maintain single-device models with a None map, diffusers does not
-            model.hf_device_map = {"": "cpu" if device == "cpu" else 0}
-
+            model.hf_device_map = {"": "cpu" if device_str == "cpu" else 0}
```
johannaSommer: here the 0 would refer to the device index, we should probably parse this from the device (for me, would be one more argument to by default always have "cuda" + device index

johannaSommer: when you rebase you will also find this change, where the 0-index can now also be adjusted
https://github.com/PrunaAI/pruna/commit/41a1ad16212e65b9f5eb6f78cea6f6c3b5d8eaf5#diff-6b1a3c4dd9435714e775e110ecf61a293fd9a9db2f26f03b2228290f83a83560R246-R293

johannaSommer (changes requested):
Looking really good already, maybe we can just think about having a device index by default but then its already really clean and actually easier than I thought, love it! Thanks David 🚀

src/pruna/engine/utils.py
davidberenstein1957: thanks, I updated this

davidberenstein1957:
> Almost! As far as I can tell the change i merged in the meantime assumes a device index that we should also tackle in this PR, can you double-check the `get_device_map` function in `pruna.engine.utils`? It currently assumes any cuda model is on "cuda:0" but we probably want to adjust this, what do you think?

@johannaSommer , thanks I misssed this. Should be good to go now.

johannaSommer (approved):

uv.lock
GreatBahram, line 1: Hey! Just passing by to mention that the `uv.lock` file is actually quite important to keep alongside the `pyproject.toml`. It contains the exact hash of each package version, which the version numbers in pyproject.toml alone can't fully guarantee. This helps ensure the environment is reliably reproducible every time. Probably I'm also missing the context :sunflower: 

If the lockfile was just outdated, you can regenerate it anytime with `uv sync`.

davidberenstein1957: Thanks. I could not reproduce the error locally, and then I did not change anything in the code and got a mypy linting error, so I was trying to figure out what was happening.

GreatBahram: I apologize for my quick reaction! 😄 It seems it was a false alarm :handshake:

johnrachwan123:
bugbot run

johnrachwan123:
@davidberenstein1957 I think it looks good but can we handle the bugbot comments before merging ?

src/pruna/config/compatibility_checks.py
```diff
@@ -47,8 +48,10 @@ def ensure_device_consistency(model, smash_config):
                 raise ValueError("Device map indicates CPU offloading, this is not supported at this time.")
             else:
                 smash_config.device_map = hf_device_map
-
-    elif smash_config.device in ["cpu", "cuda", "mps"] and model_device in ["cpu", "cuda", "mps"]:
+    # test if deve or device index index like cuda:0, cpu:1, mps:0, match options
```
guennemann: super minor. but the comment seems to be broken (phrasing)

src/pruna/config/compatibility_checks.py
guennemann, line 1: Question on line 40/41: Can it actually happen that one has an index and the other not? E.g. cuda vs. cuda:0 ? Since we are checking for exact equality.

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -123,7 +124,7 @@ def evaluate(self, model: Any) -> List[MetricResult]:
 
     def prepare_model(self, model: Any) -> PrunaModel:
         """
-        Prepare the model for evaluation by wrapping it in a PrunaModel if it is not already.
+        Prepare the model for evaluation by wrapping it in a PrunaModel and moving it to the cpu if it is not already.
```
guennemann, line 127: Probably I am missing a bit context here, but why do we move it to CPU? Usually we execute prepare_model before doing some inference. Wouldn't it be better to then use cuda if available?

guennemann (commented):
Some comments/questions inline.

src/pruna/evaluation/evaluation_agent.py
davidberenstein1957: We use move it to CPU as the `Task` definition will determine where on which device the model needs to run, so it is just additional assurance we start from a clean sheet.

src/pruna/config/compatibility_checks.py
davidberenstein1957: rephrased a bit

src/pruna/config/compatibility_checks.py
davidberenstein1957: In theory, we can initialise the model with something like "cuda" and the smash_config with "cuda:1," which means we would need to align them. This is a niche scenario, but one we decided to cover.

PR #209, opened 2025-06-23
build: unpin torch and torchvision version except for stable-fast

gsprochette:
## Description
Change the pruna dependencies on torch and torchvision to increase compatibility
- when installing with `pip install pruna[stable-fast]` or `pip install pruna[full]`, torch version remains pinned to 2.7.0
- in all other cases, torch version can now be >=2.5.1

## Type of Change

## How Has This Been Tested?
- Installed pruna with fixed versions of torch starting at 2.5.1 and fixing the version of torchvision accordingly, checking if install is done correctly
- installed pruna[stable-fast] with the dependency declared in this new way in optional-dependencies and making sure the code from the "quick start" section of the readme (using stable-fast) runs correctly.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,8 +43,8 @@ classifiers = [
 ]
 dependencies = [
     "wget",
-    "torch==2.7.0",
-    "torchvision==0.22.0",
+    "torch>=2.5.1",
+    "torchvision>=0.20.1",
     "torchmetrics[image]",
     "requests>=2.31.0",
     "transformers",
@@ -94,6 +94,7 @@ dependencies = [
 
 [project.optional-dependencies]
 stable-fast = [
+    "torch==2.7.0",
     "xformers==0.0.30",
     "stable-fast-pruna==1.0.7",
 ]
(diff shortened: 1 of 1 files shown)


johnrachwan123 (approved):
LGTM

pyproject.toml
```diff
@@ -43,8 +43,8 @@ classifiers = [
 ]
 dependencies = [
     "wget",
-    "torch==2.7.0",
-    "torchvision==0.22.0",
+    "torch>=2.5.1",
```
johannaSommer, line 46: I am a bit worried about this torch version change because we do rely on the latest version of torch for some features like e.g. the torch portable compilation, which was only introduces in 2.7. What do you think about this? For me, for now, it would be fine if we keep your change but add a check in torch compile for the `make_portable` hyperparameter and only allow it to be true if the torch version matches. If we see in the future that it is too difficult to maintain old versions, we can still just pin it again.

johannaSommer (changes requested):

PR #211, opened 2025-06-23
feature: Refactor CI to use shared setup-uv-project action 

GreatBahram:
## Description
This PR introduces a reusable composite GitHub Action to standardize `uv` setup across workflows.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #212, opened 2025-06-23
🧹 CI Cleanup

GreatBahram:
## Description
If agree with these changes, we can merge this after [this PR](https://github.com/PrunaAI/pruna/pull/211), here my goal was to simplify the workflow files which means fewer workflows to maintain, changes include:
- Rename: `package_build.yaml` → `build.yaml`
- Rename: `documentation.yaml` → `docs.yaml`
- Merged `linting.yaml` & `tests.yaml `into one → `tests.yaml `(now includes both linting + test jobs they should run on their runner and in parallel)
- Deleted: `installation.yaml` — since we have this functionality elsewhere

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


.github/workflows/installation.yaml
johannaSommer, line 1: I think we definitely have to keep this file - I understand your comment in the description, we do installation in other places, but we use this workflow to check specifically other python versions and other OS types. Some packages like e.g. gptqmodel are not compatible with all OS

.github/workflows/build.yaml
johannaSommer, line 1: the badges at the top of the README depend on this naming, can we also update them in this PR?

johannaSommer (changes requested):
Cool changes, we love a good cleanup! 🫧 Just a few things to adjust and let's also rebase once PRs this depends on are merged, then we are GTG!

.github/workflows/installation.yaml
GreatBahram: Oh, good to know that Johanna! Then I'll bring it back after we merge [this PR](https://github.com/PrunaAI/pruna/pull/211)

.github/workflows/build.yaml
GreatBahram: sure thing :saluting_face:

GreatBahram: Fixed here:
https://github.com/PrunaAI/pruna/pull/212/commits/ce573729228574b8f64e6d3bf97793bee7f23f48

.github/workflows/installation.yaml
GreatBahram: I restored the `installation.yaml` file :pray:

.github/workflows/installation.yaml
```diff
@@ -33,3 +33,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - uses: ./.github/actions/setup-uv-project
+
+      - name: Install dependencies
```
johannaSommer: One quick question here, is the pip install -e . necessary? In my understanding this is taken care of by the uv setup action

johannaSommer (commented):
Just one comment then we are already there I think!

.github/workflows/installation.yaml
GreatBahram: I checked this, you're totally right we don't need this :pray:

GreatBahram: Okay if this is the case, then theoretically if we have [python strategy matrix](https://github.com/PrunaAI/pruna/blob/02af8c28a775e64d38ecec18fd52d3b79dbaca33/.github/workflows/installation.yaml#L22-L24) in other workflows, we can potentially remove this one :thinking:

johannaSommer: ah good point, to clarify, you're suggesting this because uv anyway resolves the packages for all specified python version from the pyproject.toml?

GreatBahram: yeah, exactly, since we have the setup-uv-project almost in all workflows

.github/workflows/tests.yaml
```diff
@@ -48,3 +51,24 @@ jobs:
       - name: Set lint check output
         id: lint-check
         run: echo "success=true" >> $GITHUB_OUTPUT
+  test:
```
GreatBahram, line 54: > [!NOTE]
> Here we have the `cpu_tests.yaml` file content. It still run on the configured runner and in parallel to the linting job.

.github/workflows/installation.yaml
johannaSommer: 💯 thats a great idea, then we could basically keep the installation tests but only with a matrix over macos/windows, agree? i love it, let's tackle it in a separate PR?

PR #226, opened 2025-07-01
fix: `device_map` specification for `accelerate`-compatible quantizers

johannaSommer:
## Description
This PR fixes a small bug in the device map specification in `pruna`, namely, the device map should be specified as `
{"":"cuda:0"}` instead of `{"":"cuda"}`.  Otherwise, the resulting model will work at inference time but will no longer be compatible with other `accelerate` functionality.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Reran all algorithms tests in pruna (due to change in the load-function for transformers models), including the accelerate tests for both quantizers. 

## Checklist

## Additional Notes
At the moment we assume throughout the pruna repository that the device index is 0, hence the changes in this PR. We will update this soon and be compatible with specifying the device index when using only "cuda".

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -230,7 +230,7 @@ def cast_model_to_accelerate_device_map(model, device_map):
     model.hf_device_map = device_map.copy()
 
 
-def get_device(model: Any) -> str | dict[str, str]:
+def get_device(model: Any) -> str:
     """
     Get the device of the model.
 
@@ -241,7 +241,7 @@ def get_device(model: Any) -> str | dict[str, str]:
 
     Returns
     -------
-    str | dict[str, str]
+    str
         The device or device map of the model.
     """
     if isinstance(model, Pipeline):


gsprochette (commented):
Looks great ! Thanks for the improvement and for future-proofing pruna :)
While we're on this subject I think the `get_device` function in `pruna.engine.utils` could be improved:
-  most important: the `return_device_map` is only handled in the case of the `hf_device_map` check at the end, but the `model_device.type` does not seem to be a valid device_map as suggested by the argument name and docstring. This will probably cause issues in the future.
- readability: the first if statement would be slightly cleaner with the else case first
- readability: the initial `model_device` is overwritten at the end, there should maybe be a `if..else` to make the different cases more explicit
We can make this into a separate PR but the first point seems related.

simlang (commented):
Changes so far look good to me, thanks for fixing it so fast! :+1:
Will approve, depending on how we handle @gsprochette comment :slightly_smiling_face:

src/pruna/engine/utils.py
```diff
@@ -230,44 +230,71 @@ def cast_model_to_accelerate_device_map(model, device_map):
...
+        The device map of the model.
+    """
+    model_device = get_device(model)
+    if model_device == "accelerate":
+        if subset_key is None:
+            return model.hf_device_map
+        else:
+            return model.hf_device_map[subset_key]
+    else:
+        device = "cuda:0" if model_device == "cuda" else "cpu"
```
gsprochette: There should probably be mps case here

src/pruna/engine/load.py
```diff
@@ -249,7 +249,8 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
         architecture = config["architectures"][0]
         cls = getattr(transformers, architecture)
         # transformers discards kwargs automatically, no need for filtering
-        device_map = smash_config.device_map if smash_config.device == "accelerate" else smash_config.device
+        device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
```
simlang, line 252: Can the device in smash_config be a torch.device?
and if so, would torch.device("type": "cuda") == "cuda" be true?

simlang (approved):
Looks good! Very clean now 🫧🧼

gsprochette (approved):
Looks absolutely perfect, thanks for all the "just this last update" iterations it's so clean now :)

PR #227, opened 2025-07-02
Reduce Core Dependencies in pyproject.toml 

### LINKED ISSUE #225: Reduce the amount of core dependencies in pyproject.toml
I was exploring the `pyproject.toml` for the project and I noticed some dependencies that can be safely removed in the [project] section without any major breaking changes.

Firstly, `setuptools` and `packaging` is not required to be in the dependencies as hatchling is a [PEP517](https://peps.python.org/pep-0517/) compliant and it implements the same functionality without the usage of `setuptools` and `packaging`.

https://github.com/PrunaAI/pruna/blob/034d474e2c5b8da8b2eca95a19877a4e6a246271/pyproject.toml#L61-L62

Using the `grep` and `find` command, I found that some dependencies are not explicitly used in the codebase, we could make these dependencies optional to reduce the size of the library and make it efficient as well.

The jupyter notebook dependency, along with `ipykernel` can be kept in optional dependencies as well. The same can be done for the `wget` library as it does not seem to be explicitly used in the codebase.

https://github.com/PrunaAI/pruna/blob/034d474e2c5b8da8b2eca95a19877a4e6a246271/pyproject.toml#L49

https://github.com/PrunaAI/pruna/blob/034d474e2c5b8da8b2eca95a19877a4e6a246271/pyproject.toml#L75

Also, I think some dependencies from the `pruna-pro` package are present which are not used in this codebase, could we make them optional as well?

https://github.com/PrunaAI/pruna/blob/034d474e2c5b8da8b2eca95a19877a4e6a246271/pyproject.toml#L69

https://github.com/PrunaAI/pruna/blob/034d474e2c5b8da8b2eca95a19877a4e6a246271/pyproject.toml#L79-L81

Could you please let me know?

cc: @davidberenstein1957

ParagEkbote:
## Description
As described in the issue, I have reduced the number of core dependencies required for the package. All the dependencies which are not explicitly imported are added in the optional dependencies. Could you please review?

## Related Issue

Fixes #225

## Type of Change

## Checklist


pyproject.toml
```diff
@@ -116,10 +101,24 @@ full = [
...
     "notebook",
+    "peft",
+    "optuna",
+    "trl",
+    "platformdirs",
+    "jsonschema",
+    "blobfile",
+    "pillow>=9.5.0",
     "pre-commit",
-    "build",
```
ParagEkbote, line 122: Since the `hatchling` build backend is used, the build package is not required to be added explicitly :)

ParagEkbote (commented):
Some initial comments on the PR:

pyproject.toml
```diff
@@ -46,39 +46,24 @@ classifiers = [
...
     "torch_pruning",
-    "python-dotenv",
     "ConfigSpace>=1.2.1",
-    "cython",
-    "lago-python-client",
     "sentencepiece",
-    "blobfile",
     "librosa",
     "openvino",
     "soundfile",
```
ParagEkbote, line 64: Are `librosa`, `openvino` and `soundfile` imported explicitly, or used in other important utility. Could they be made optional?

pyproject.toml
```diff
@@ -116,10 +101,24 @@ full = [
...
     "jupyterlab",
+    "ipywidgets>=8.1.5",
+    "cython",
+    "protobuf",
+    "lago-python-client",
+    "ipykernel",
     "notebook",
+    "peft",
+    "optuna",
+    "trl",
```
johannaSommer: all of those can be safely remove completely afaik 😊

pyproject.toml
```diff
@@ -116,10 +101,24 @@ full = [
...
 ]
 dev = [
+    "wget",
+    "python-dotenv",
+    "jsonschema",
     "jupyterlab",
+    "ipywidgets>=8.1.5",
+    "cython",
+    "protobuf",
+    "lago-python-client",
```
johannaSommer: can also be removed

pyproject.toml
```diff
@@ -116,10 +101,24 @@ full = [
     "stable-fast-pruna==1.0.7",
 ]
 dev = [
+    "wget",
+    "python-dotenv",
+    "jsonschema",
     "jupyterlab",
+    "ipywidgets>=8.1.5",
+    "cython",
```
johannaSommer: can also be removed

johannaSommer (changes requested):
Thanks a lot for taking the time for this, a 20% reduction in core dependencies is AMAZING! I've identified a few dependencies that are no longer used overall and can be removed and after @johnrachwan123 has had a final look we're already good to go! 🚀

pyproject.toml
ParagEkbote: Done 👍

pyproject.toml
ParagEkbote: Done 👍

pyproject.toml
johnrachwan123: No these ones should be kept as they are used implicitely

johnrachwan123 (approved):
Thanks a lot for the nice PR! LGTM!

PR #233, opened 2025-07-08
docs: create end to end video tutorial

### LINKED ISSUE #186: [DOC] Add and end to end tutorial for `text-to-video`
## What part of the documentation do you want to update?

It would be great to showcase a clean end2end tutorial for video / `text-to-video` like the ones we did earlier for other key modalities.

https://github.com/PrunaAI/pruna/issues/120
https://github.com/PrunaAI/pruna/issues/121

sdiazlor:
## Related Issue

Fixes #186 

## Type of Change

## Checklist

### DIFF
--- a/docs/tutorials/index.rst
+++ b/docs/tutorials/index.rst
@@ -19,6 +19,12 @@ This tutorial will guide you through the process of using |pruna| to optimize yo
 
       Compress with ``hqq`` quantization and ``torch_compile`` compilation and evaluate with ``elapsed_time`` and ``perplexity``.
 
+   .. grid-item-card:: Compress and Evaluate Video Generation Models
+      :text-align: center
+      :link: ./video_generation.ipynb
+
+      Compress with a ``pab``, and evaluate with ``total time`` and ``latency``.
+
    .. grid-item-card:: Transcribe 2 hour of audio in 2 minutes with Whisper
       :text-align: center
       :link: ./asr_tutorial.ipynb
--- a/docs/tutorials/video_generation.ipynb
+++ b/docs/tutorials/video_generation.ipynb
@@ -1,0 +1,389 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Compress and Evaluate Video Generation Models\n",
+    "\n",
+    "<a target=\"_blank\" href=\"https://colab.research.google.com/github/PrunaAI/pruna/blob/v|version|/docs/tutorials/video_generation.ipynb\">\n",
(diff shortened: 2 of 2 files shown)


docs/tutorials/index.rst
```diff
@@ -19,6 +19,12 @@ This tutorial will guide you through the process of using |pruna| to optimize yo
 
       Compress with ``hqq`` quantization and ``torch_compile`` compilation and evaluate with ``elapsed_time`` and ``perplexity``.
 
+   .. grid-item-card:: Compress and Evaluate Video Generation Models
+      :text-align: center
+      :link: ./video_generation.ipynb
+
+      Compress with a ``pab``, and evaluate with ``total time`` and ``latency``.
```
davidberenstein1957: aren't there multiple techniques we can apply? Like compilation? I believe latency = 1/total time. Perhaps we can add things like energy consumption too?

docs/tutorials/video_generation.ipynb
davidberenstein1957, line 1: Perhaps we can also add the device within the overview. Could you double check if the notebook button works. It seemed to have broken for the other tutorials. 
Could you add some print statements to the notebook too?

docs/tutorials/index.rst
sdiazlor: In open source, the best output for Wan was using only `pab` . I'll add the energy consumption metric.

sharpenb: We should be able also to apply some compilation technique. Let's also add througput, co2 and nergy for the metrics :)

sharpenb (changes requested):
Looks great. Here are some comment regarding the notebook:
- We should link to the installation page for the pruna installation in addition to  recalling the cli. BTW, [here](https://docs.pruna.ai/en/stable/docs_pruna/tutorials/image_generation.html) is missing the cli and the link ot the isntallation page.
- It mentions "Text-to-image model" but should be "text-to-video" I think
- We do not see the video in the notebook. Could we show an example of a base and compressed video?
- Let's use a more involved smash config
- The optimization we use here are (almost) deterministic here. The seed might not be fixed though.
- Let's remove the statement that we will add video metrics in the future. While it is true, it might be depricated very soon.
- The result table is not shown on this notebok
- We do reduce the memory in this tutorial yet but it could be interesting to have memory metrics and some quantization algos.

davidberenstein1957:
cc @sdiazlor

johannaSommer:
Hey! Wanted to quickly let you know that I am merging #287 soon which applies to video diffusion :) Feel free to integrate into this tutorial if you think it makes sense!

PR #236, opened 2025-07-08
feat: evaluation agent on multiGPU

begumcig:
## Description

This PR adds support for evaluating time metrics on multi-GPU setups using HuggingFace Accelerate and introduces a consistent, centralized system for device resolution across the model, Task, EvaluationAgent, and metrics.

Evaluation Pipeline

- EvaluationAgent now records the model's device and device map, and restores the model to its original setup after evaluation.

- Introduced `ensure_device_consistency()` to verify that Task and model devices match, or to sync them if not explicitly set.

Task Improvements

- Task now has a `low_memory=True` mode to offload stateful metrics to CPU by default.

- Separates inference-time device (task.device) from stateful-metric device (task.stateful_metric_device) to support mixed configurations.

Improved fallback logic:

- If the user doesn’t explicitly set a device on the Task, the system defaults to using the **model’s device**. Metrics inherit this unless overridden. 

- If `low_memory=True`, only stateful metrics are automatically offloaded to CPU to reduce GPU memory pressure.

- If the user does specify a device and it doesn’t match the model, we raise a clear error; no silent casting. Metrics themselves also raise if their .device is incompatible with their .runs_on spec.

Inference Flow Update

- `PrunaModel.run_inference()` now automatically detects model device and applies input placement logic accordingly.

- `move_inputs_to_device()` now supports Accelerate device maps and works consistently across CPU, CUDA, and Accelerate setups.

## Type of Change

## Checklist

### DIFF
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -40,9 +40,13 @@ Create a new class that inherits from ``BaseMetric`` and implements the ``comput
 
 Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
 
-Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
+Every metric has a ``runs_on`` attribute that specifies the device(s) it supports.
 
-Please update the ``runs_on`` if you want to change the default devices. For instance, you can also include ``accelerate`` if your metric logic works on multiple devices. Inference with ``accelerate`` should work out of the box, and some of |pruna|'s metrics already use it!
+For custom metrics, you should explicitly set this list. If not specified, the default is ``["cuda", "cpu", "mps"]``.
+
+You can extend this list as needed. For example, include ``accelerate`` if your metric logic is compatible with multiple devices.
+
+Inference with accelerate works out of the box, and some of |pruna|’s built-in metrics already make use of it.
 
 ``compute()`` takes two parameters: ``model`` and ``dataloader``.
 
@@ -81,9 +85,13 @@ To implement a ``StatefulMetric``, create a class that inherits from ``StatefulM
 
 Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
 
-Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
+Every metric has a ``runs_on`` attribute that specifies the device(s) it supports.
 
-Please update the ``runs_on`` if you want to change the default devices.
+For custom metrics, you should explicitly set this list. If not specified, the default is ``["cuda", "cpu", "mps"]``.
+
+You can extend this list as needed. For example, include ``accelerate`` if your metric logic is compatible with multiple devices.
+
+Inference with accelerate works out of the box, and some of |pruna|’s built-in metrics already make use of it.
 
 Use ``add_state()`` method to define internal state variables that will accumulate data across batches. For example, you might track totals and counts to compute an average.
 
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -42,7 +42,7 @@ Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. 
 
 Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
 
-Please update the ``runs_on`` if you want to change the default devices.
+Please update the ``runs_on`` if you want to change the default devices. For instance, you can also include ``accelerate`` if your metric logic works on multiple devices. Inference with ``accelerate`` should work out of the box, and some of |pruna|'s metrics already use it!
 
 ``compute()`` takes two parameters: ``model`` and ``dataloader``.
 
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -13,11 +13,9 @@ If anything is unclear or you want to discuss your contribution before opening a
 ------------------------------------
 
 |pruna|'s evaluation system supports two types of metrics, located under ``pruna/evaluation/metrics``: ``BaseMetric`` and ``StatefulMetric``.
-|pruna|'s evaluation system supports two types of metrics, located under ``pruna/evaluation/metrics``: ``BaseMetric`` and ``StatefulMetric``.
 
 These two types are designed to accommodate different use cases.
 
-- **BaseMetric**: Inherit from ``BaseMetric`` and compute values directly without maintaining state.
 - **BaseMetric**: Inherit from ``BaseMetric`` and compute values directly without maintaining state.
     - Used when isolated inference is required (e.g., ``latency``, ``disk_memory``, etc.)
 - **StatefulMetric**: Inherit from ``StatefulMetric`` and accumulate state across multiple batches.
@@ -32,7 +30,6 @@ These two types are designed to accommodate different use cases.
 Create a new file in ``pruna/evaluation/metrics`` with a descriptive name for your metric. (e.g, ``your_new_metric.py``)
 
 We use snake_case for the file names (e.g., ``your_new_metric.py``), PascalCase for the class names (e.g, ``YourNewMetric``) and NumPy style docstrings for documentation.
-We use snake_case for the file names (e.g., ``your_new_metric.py``), PascalCase for the class names (e.g, ``YourNewMetric``) and NumPy style docstrings for documentation.
 
 Both  ``BaseMetric`` and ``StatefulMetric`` return a ``MetricResult`` object, which contains the metric name, result value and other metadata.
 
@@ -41,11 +38,14 @@ Implementing a ``BaseMetric``
 
 Create a new class that inherits from ``BaseMetric`` and implements the ``compute()`` method.
 
-Your metric should have a ``metric_name`` attribute and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+
+Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
+
+Please update the ``runs_on`` if you want to change the default devices.
 
 ``compute()`` takes two parameters: ``model`` and ``dataloader``.
 
-Inside ``compute()``, you are responsible for running inference manually.
 Inside ``compute()``, you are responsible for running inference manually.
 
 Your method should return a ``MetricResult`` object with the metric name, result value and other metadata. The result value should be a float or int.
@@ -54,13 +54,13 @@ Your method should return a ``MetricResult`` object with the metric name, result
 
     from pruna.evaluation.metrics.metric_base import BaseMetric
     from pruna.evaluation.metrics.result import MetricResult
-    from pruna.evaluation.metrics.result import MetricResult
 
     class YourNewMetric(BaseMetric):
         '''Your metric description'''
 
         metric_name = "your_metric_name"
         higher_is_better = True # or False
+        runs_on = ["cuda", "cpu"] # This is optional. Only change if your metric runs on different devices than the default ones.
 
         def __init__(self):
             super().__init__()
@@ -79,13 +79,19 @@ Implementing a ``StatefulMetric``
 
 To implement a ``StatefulMetric``, create a class that inherits from ``StatefulMetric``. These metrics are designed to accumulate state across multiple batches and can share inference with other metrics.
 
-Your metric should have a ``metric_name`` attribute and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+
+Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
+
+Please update the ``runs_on`` if you want to change the default devices.
 
 Use ``add_state()`` method to define internal state variables that will accumulate data across batches. For example, you might track totals and counts to compute an average.
 
 The ``update()`` method processes each batch of data, updating the state variables based on the current batch. It takes three parameters: ``inputs``, ``ground_truths`` and ``predictions``.
 
 The ``compute()`` method is called after all batches are processed and returns a ``MetricResult`` object, which contains the final metric value calculated from the accumulated state.
+
+The ``move_to_device()`` method is used to move the metric and necessary attributes and variables of the metric to the specified device.
 
 Metrics can operate in both single-model and pairwise modes, determined by the ``call_type`` parameter. Common ``call_types`` include ``y_gt``, ``gt_y``, ``x_gt``, ``gt_x``, ``pairwise_y_gt``, and ``pairwise_gt_y``. For more details, see the :ref:`Understanding Call Types <understanding-call-types>` section.
 
@@ -106,7 +112,7 @@ Here's a complete example implementing a ``StatefulMetric`` with a single ``call
         default_call_type = "y_gt"
         metric_name = "your_metric_name"
         higher_is_better = True # or False
-
+        runs_on = ["cuda", "cpu"]
         def __init__(self, param1='default1', param2='default2', call_type=SINGLE): # Since we picked a single call_type for default, we can use it as a default value
             super().__init__()
             self.param1 = param1
@@ -132,6 +138,12 @@ Here's a complete example implementing a ``StatefulMetric`` with a single ``call
                 return 0
             return MetricResult(self.metric_name, self.__dict__.copy(), self.total / self.count)
 
+        def move_to_device(self, device: str | torch.device):
+            # Move the metric and necessary attributes and variables of the metric to the specified device
+            self.device = device
+            self.total = self.total.to(device)
+            self.count = self.count.to(device)
+
 .. _understanding-call-types:
 
 Understanding Call Types
@@ -156,9 +168,10 @@ Understanding Call Types
 | `pairwise_y_gt`    | Base model's output first, then subsequent model's output   |
 +--------------------+-------------------------------------------------------------+
 | `pairwise_gt_y`    | Subsequent model's output first, then base model's output   |
-+--------------------+-------------------------------------------------------------+
++--------------------+-------------------------------------------------------------+ 
 | `y`                | Only the output is used, the metric has an internal dataset |
 +--------------------+-------------------------------------------------------------+
+
 
 You need to decide on the default ``call_type`` based on the metric you are implementing.
 
@@ -198,7 +211,6 @@ In this case, you can pass ``pairwise`` to the ``call_type`` parameter of the ``
 
 
     # Initialize your metric from the instance
-    YourNewStatefulMetric(param1='value1', param2='value2', call_type="pairwise")
     YourNewStatefulMetric(param1='value1', param2='value2', call_type="pairwise")
 
 If you have implemented your metric using the correct ``get_call_type_for_metric`` function and ``metric_data_processor`` function, this will work as expected.
@@ -251,7 +263,6 @@ Thanks to this registry system, everyone using |pruna| can now refer to your met
     from pruna.evaluation.metrics.your_metric_file import YourNewMetric
 
     # Classic way: Initialize your metric from the instance
-    YourNewMetric(param1='value1', param2='value2')
     YourNewMetric(param1='value1', param2='value2')
 
 .. code-block:: python
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -1,5 +1,5 @@
 Customize a Metric
-===============================
+==================
 
 This guide will walk you through the process of adding a new metric to Pruna's evaluation system.
 
@@ -168,7 +168,7 @@ Understanding Call Types
 | `pairwise_y_gt`    | Base model's output first, then subsequent model's output   |
 +--------------------+-------------------------------------------------------------+
 | `pairwise_gt_y`    | Subsequent model's output first, then base model's output   |
-+--------------------+-------------------------------------------------------------+ 
++--------------------+-------------------------------------------------------------+
 | `y`                | Only the output is used, the metric has an internal dataset |
 +--------------------+-------------------------------------------------------------+
 
--- a/docs/user_manual/customize_metric.rst
+++ b/docs/user_manual/customize_metric.rst
@@ -13,9 +13,11 @@ If anything is unclear or you want to discuss your contribution before opening a
 ------------------------------------
 
 |pruna|'s evaluation system supports two types of metrics, located under ``pruna/evaluation/metrics``: ``BaseMetric`` and ``StatefulMetric``.
+|pruna|'s evaluation system supports two types of metrics, located under ``pruna/evaluation/metrics``: ``BaseMetric`` and ``StatefulMetric``.
 
 These two types are designed to accommodate different use cases.
 
+- **BaseMetric**: Inherit from ``BaseMetric`` and compute values directly without maintaining state.
 - **BaseMetric**: Inherit from ``BaseMetric`` and compute values directly without maintaining state.
     - Used when isolated inference is required (e.g., ``latency``, ``disk_memory``, etc.)
 - **StatefulMetric**: Inherit from ``StatefulMetric`` and accumulate state across multiple batches.
@@ -30,6 +32,7 @@ These two types are designed to accommodate different use cases.
 Create a new file in ``pruna/evaluation/metrics`` with a descriptive name for your metric. (e.g, ``your_new_metric.py``)
 
 We use snake_case for the file names (e.g., ``your_new_metric.py``), PascalCase for the class names (e.g, ``YourNewMetric``) and NumPy style docstrings for documentation.
+We use snake_case for the file names (e.g., ``your_new_metric.py``), PascalCase for the class names (e.g, ``YourNewMetric``) and NumPy style docstrings for documentation.
 
 Both  ``BaseMetric`` and ``StatefulMetric`` return a ``MetricResult`` object, which contains the metric name, result value and other metadata.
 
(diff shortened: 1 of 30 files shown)


src/pruna/config/compatibility_checks.py
```diff
@@ -36,19 +36,23 @@ def ensure_device_consistency(model, smash_config):
     """
     model_device = get_device(model)
 
+    # to handle the device cases like "cuda:0 and cuda, cuda:1"
+    model_device_kind, model_device_index = split_device(model_device)
+    smash_config_device_kind, smash_config_device_index = split_device(smash_config.device)
+
     # model and smash config devices match
-    if model_device == smash_config.device:
+    if (model_device_kind == smash_config_device_kind) and (model_device_index == smash_config_device_index):
```
johannaSommer: this is just me being petty but if you want you can consider sticking to the torch.device naming for this, which I think will be "device_type" instead of "device_kind"

src/pruna/engine/handler/handler_inference.py
```diff
@@ -80,15 +83,17 @@ def move_inputs_to_device(
...
         List[str] | torch.Tensor
             The prepared inputs.
         """
-        if isinstance(inputs, torch.Tensor):
-            return inputs.to(device)
-        elif isinstance(inputs, tuple):
-            return tuple(self.move_inputs_to_device(item, device) for item in inputs)  # type: ignore
-        else:
-            return inputs
+        if device == "accelerate":
```
johannaSommer, line 92: we discussed this already, *technically* this is not required, we can also cast the inputs to cpu/cuda and accelerate will take care of this... I don't really mind keeping this but to slim down the functionality overall we could consider removing this logic?

johannaSommer: if you decide to do this, we could use best_available_device to put it either on cuda:0 or cpu if unavailable

src/pruna/engine/pruna_model.py
```diff
@@ -148,6 +147,40 @@ def __delattr__(self, attr: str) -> None:
         """
         delattr(self.model, attr)
 
+    def get_device(self, **kwargs: Any) -> str:
```
johannaSommer: i am slightly lost why we also make this a pruna function here instead of using the utility directly, could you help me out?

src/pruna/engine/pruna_model.py
```diff
@@ -163,16 +196,21 @@ def set_to_eval(self) -> None:
         """Set the model to evaluation mode."""
         set_to_eval(self.model)
 
-    def move_to_device(self, device: str) -> None:
+    def move_to_device(self, device: str, device_map: dict[str, str] | None = None, **kwargs: Any) -> None:
```
johannaSommer: just an idea/question conceptually, the Pruna Model is tied to a _specific_ model & smash config, which already contains the device map... do we really require to pass it here or cant we just always get it from self.smash_config.device_map where this is anyway saved?

src/pruna/engine/utils.py
```diff
@@ -131,9 +131,17 @@ def move_to_device(model: Any, device: str, raise_error: bool = False, device_ma
             delattr(model.model, "hf_device_map")
         return
 
+    device = device_to_string(device)
```
johannaSommer, line 141: just a quick heads up that I think this is also tackled in davids PR about the device indexes, depending on who merges first I am just prepping you for a conflict

src/pruna/engine/utils.py
```diff
@@ -412,41 +471,93 @@ def set_to_best_available_device(device: str | torch.device | None) -> str:
...
+    elif kind == "mps":
         if not torch.backends.mps.is_available():
             pruna_logger.warning("'mps' requested but not available.")
-            return set_to_best_available_device(device=None)
         return "mps"
+    else:
+        raise ValueError(f"Device not supported: '{device}'")
+
+
+def device_to_string(device: str | torch.device) -> str:
```
johannaSommer, line 600: not 1000% sure but afaik you can just to this:

```python
device = torch.device("cuda:0")
print(str(device))  # Outputs: 'cuda:0'
```

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -106,7 +106,12 @@ def evaluate(self, model: Any) -> List[MetricResult]:
         pruna_logger.info("Evaluating isolated inference metrics.")
         results.extend(self.compute_stateless_metrics(model, stateless_metrics))
 
-        model.move_to_device("cpu")
+        # Move model back to the original device.
+        model.move_to_device(self.device, device_map=self.device_map)
+        pruna_logger.info(
+            f"Evaluation run has finished. Moved model to device {self.device} with device map {self.device_map}."
```
johannaSommer: I think this print might look weird in case of evaluation on CUDA/CPU, could that be? I guess its strange to mention a device map in this case

src/pruna/evaluation/metrics/metric_base.py
```diff
@@ -28,6 +30,33 @@ class BaseMetric(ABC):
...
+    @device.setter
+    def device(self, dvc: str | torch.device):
+        """Validate and store the device.
+
+        Parameters
+        ----------
+        value : str | torch.device
+            The device to set.
+        """
+        # To prevent the user from setting an unsupported device.
```
johannaSommer, line 49: you're so cool

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -91,13 +93,32 @@ def _measure(self, model: PrunaModel, dataloader: DataLoader, iterations: int, m
...
+        model : PrunaModel
+            The model to get the device map from.
+        """
+        device_type, device_idx = split_device(device_to_string(self.device))
+        if device_type == "cuda":
+            torch.cuda.synchronize(device_idx)
+        elif device_type == "accelerate":
+            device_map = model.get_device_map()
+            for device in device_map.values():
+                torch.cuda.synchronize(device)
```
johannaSommer, line 128: it's so cool that you're taking care of all of this, I am confident this is correct and I think I understand conceptually what is happening here, but since this is the secret sauce of the latency metrics, can we add some comments here? E.g. what exactly does the synchronize do and why is it important? would we have to de-sync?

src/pruna/evaluation/metrics/metric_pairwise_clip.py
```diff
@@ -48,12 +49,17 @@ class PairwiseClipScore(CLIPScore, StatefulMetric):
     metric_name: str = "pairwise_clip_score"
 
     def __init__(self, **kwargs: Any) -> None:
-        if "device" in kwargs:
-            kwargs.pop("device")
+        device = kwargs.pop("device", None)
+        device = set_to_best_available_device(device)
```
johannaSommer, line 53: at this point I feel like ive seen this code snippet a lot, would it be possible to perform this check in the superclass so that its called in the `super.__init__()`

src/pruna/evaluation/metrics/metric_pairwise_clip.py
```diff
@@ -96,6 +102,21 @@ def compute(self) -> Any:
         pairwise_score_item = pairwise_score.item() if isinstance(pairwise_score, Tensor) else pairwise_score
         return MetricResult(self.metric_name, self.__dict__.copy(), pairwise_score_item)
 
+    def move_to_device(self, device: str | torch.device) -> None:
```
johannaSommer, line 101: same comment as above for this

johannaSommer: i mean: can this also go into the superclass?

src/pruna/evaluation/metrics/utils.py
```diff
@@ -75,16 +94,34 @@ def metric_data_processor(
     >>> # Returns [ground_truth, model_outputs]
     """
     if call_type == "x_y":
+        if device is not None:
```
johannaSommer: i am for sure missing something, but could we - for conciseness - have a code snippet in the beginning where we move x, outputs, gt and then keep the logic as is?

johannaSommer (changes requested):
speechless

src/pruna/engine/handler/handler_inference.py
begumcig: Yes, I completely agree that we shouldn’t replicate logic that Accelerate already handles internally. That makes total sense.
 In cases where the model is on a specific CPU or GPU device, we can’t rely on `best_available_device(None) `though. Because we explicitly need to move the inputs to the model’s actual device, not just any available device.

Also `get_device()` currently returns "accelerate" as a placeholder, and replacing it with a concrete device (like the first CUDA device) might break other parts of the codebase that expect "accelerate".

So I think we can do two things: 

- Use `best_available_device(None) `only when the device is "accelerate", to resolve a concrete device for casting inputs.

- Skip moving inputs altogether if the device is "accelerate", and assume Accelerate will handle it.

I have done the first, lmk if that makes sense!

src/pruna/engine/pruna_model.py
begumcig: Yes, of course! Since the evaluation agent is used by both our open-source and pro models, we need to make sure the utility functions can be called from both sides. I don't think we could call the utility function from the evaluation agent agnostically for both oss/pro models in a safe manner.

src/pruna/engine/utils.py
```diff
@@ -131,9 +131,17 @@ def move_to_device(model: Any, device: str, raise_error: bool = False, device_ma
             delattr(model.model, "hf_device_map")
         return
 
+    device = device_to_string(device)
+    # To handle the device cases like "cuda:0, cuda:1"
+    kind, idx = split_device(device)
```
guennemann: I would prefer "type" here (I think Johanna had a similar comment somewhere else). Or is there some other reason why "type" is not optimal / has some conflict?

src/pruna/engine/utils.py
```diff
@@ -131,9 +131,17 @@ def move_to_device(model: Any, device: str, raise_error: bool = False, device_ma
...
+
     # sanity check for expected device types
-    if device not in ["cpu", "cuda", "mps", "accelerate"]:
-        raise ValueError("Device must be a string in [cpu, cuda, mps, accelerate].")
+    if device == "accelerate":
+        pass  # Handle accelerate separately
```
guennemann: Just some minor point. I find it easier to read to move this after the next statement.

src/pruna/engine/utils.py
```diff
@@ -401,7 +407,60 @@ def determine_dtype(pipeline: Any) -> torch.dtype:
...
+    dict[int, int]
+        The number of bytes free per GPU.
+    """
+    if torch.cuda.is_available():
+        return {i: torch.cuda.mem_get_info(i)[0] for i in range(torch.cuda.device_count())}
+    else:
+        return {}
+
+
+def set_to_best_available_device(
```
guennemann, line 554: I find this function a bit confusing. I would expect it to have no device argument since we want to place it on the best device (the case which you currently handle when device=None). But currently you also allow to provide a device (and check whether it is available). But this is then no longer "the best" device. Probably just renaming the function will work.

src/pruna/evaluation/evaluation_agent.py
guennemann: Agree. Probably just do a case distinction based on the types.

src/pruna/engine/utils.py
begumcig: omg perfect

src/pruna/evaluation/metrics/metric_pairwise_clip.py
begumcig: Yes you are absolutely correct!!!! I cannot do this for the PairwiseClip Score since it directly inherits from the torchmetrics CLIPScore which doesn't let us play with the device attribute. For all other StatefulMetrics i have moved the device setting to the base class!!

src/pruna/evaluation/metrics/utils.py
begumcig: Yes I think i went to overthinking mode here, I thought I was saving big memory by only moving two of the three tensors that's passed to this function 😂

src/pruna/evaluation/metrics/metric_elapsed_time.py
```diff
@@ -115,22 +136,16 @@ def _time_inference(self, model: PrunaModel, x: Any) -> float:
             The elapsed time in milliseconds.
         """
         if self.timing_type == "async" or self.device == "cpu":
```
guennemann, line 148: I do understand async and sync for training. But is async really useful (and makes a difference) for inference? Since the inference has to wait until every prediction/generation is done (even in the async setting) the times should be more or less the same?

Or is it because some previous jobs might still run on the GPUs?

src/pruna/engine/pruna_model.py
```diff
@@ -74,31 +74,30 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
...
         """
-        device = set_to_best_available_device(device)
-        batch = self.inference_handler.move_inputs_to_device(batch, device)  # type: ignore
+        if self.model is None:
+            raise ValueError("No more model available, this model is likely destroyed.")
+        # Rather than giving a device to the inference call,
+        # we should run the inference on the device of the model.
+        model_device = get_device(self.model)
+        device_map = get_device_map(self.model)
+        batch = self.inference_handler.move_inputs_to_device(batch, model_device, device_map)  # type: ignore
```
guennemann: What happens if there is not enough memory on the device?

src/pruna/engine/pruna_model.py
guennemann: I like this suggestion!

src/pruna/engine/pruna_model.py
guennemann: Seems that for accelerate you later check whether the model has been offloaded. But how is it handled in other cases? (And probably proactively to prevent offloading?).

src/pruna/evaluation/task.py
```diff
@@ -120,9 +128,46 @@ def is_pairwise_evaluation(self) -> bool:
         """
         return any(metric.is_pairwise() for metric in self.metrics if isinstance(metric, StatefulMetric))
 
+    def _set_stateful_metric_device_from_task_device(self) -> str:
+        """
+        Return the device for stateful metrics based on the task device.
```
guennemann: For a setter function, I would have expected to also set the device and not just return it.

src/pruna/evaluation/task.py
```diff
@@ -60,21 +60,29 @@ class Task:
...
+        low_memory: bool = False,
     ) -> None:
-        device = set_to_best_available_device(device)
-        self.metrics = get_metrics(request, device)
+        self.auto_device = device is None  # We set this for the compatibility checks later on.
+        self.low_memory = low_memory  # When we wish to be memory efficient on stateful metrics.
+        self.device = set_to_best_available_device(
+            device
+        )  # The inference device is set as the task device for evaluation agent and optimization agent.
+        self.stateful_metric_device = self._set_stateful_metric_device_from_task_device()
```
guennemann: see blow: For a setter function, I would have expected to also set the device and not just return it.

docs/user_manual/customize_metric.rst
```diff
@@ -36,7 +36,11 @@ Implementing a ``BaseMetric``
 
 Create a new class that inherits from ``BaseMetric`` and implements the ``compute()`` method.
 
-Your metric should have a ``metric_name`` attribute and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+
+Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
```
guennemann: Probably mention (if not already done) somewhere what the options are (i.e. these three plus accelerate)

guennemann (changes requested):
Thanks. Very nice. Some minor comments below.

src/pruna/engine/utils.py
begumcig: Yes type is reserved unfortunately, but I changed it with device_type for consistency also with the above updates! Thanks for pointing it out!

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: I think the timings will differ because the async branch doesn’t synchronize, so it mostly measures kernel enqueue time rather than the actual GPU work. The sync path waits for completion, so its number is the real latency. That’s why they’re not the same.

src/pruna/evaluation/metrics/metric_elapsed_time.py
begumcig: Yes I tried to be a bit more explicit in the docsring now! We don't need to de-sync at all, this call just makes sure that we wait until everything is synced!

src/pruna/evaluation/metrics/metric_pairwise_clip.py
begumcig: move to device can't go into the superclass, at least for now, because of the difference between the wrapper metrics and our own metrics.

src/pruna/engine/utils.py
begumcig: Totally open to improving this! Just to clarify, this function wasn’t introduced in this PR! it’s been around for a while and is used in quite a few places across the codebase. So renaming or refactoring it would probably be a broader task and might be better handled separately to keep this PR focused. We had previous discussions about this and I agree that it is a bit confusing, but we should probably have a broader conversation before changing it here :) !

src/pruna/engine/pruna_model.py
begumcig: Okay, after our discussion, I have completely removed the utility functions from the PrunaModel! They now only live in the utils :)

src/pruna/engine/pruna_model.py
begumcig: This is a great idea! I’ve moved all device casting logic out of PrunaModel into the utils file to keep it model-agnostic and reusable. In the updated logic, if the model has a smash_config and the requested device is "accelerate", it will use smash_config.device_map. This way, we don't need to manually pass the device_map from the outside, it gets inferred where appropriate.

I also added logic in the EvaluationAgent to explicitly set the device_map for when evaluating a base model. For any non-Pruna or non-smashed model, we normally wouldn't have a device_map, so I believe this might have caused issues when moving base models to "accelerate".

Lmk if I missed anything!

src/pruna/engine/pruna_model.py
begumcig: Oh, great catch! I am adding an OOM catch for batch movements! The offloading is something done on the accelerate side automatically, maybe we can investigate it further. As far as I know for OOM errors and offloading there's not much to do other than log them smartly, not even possible to continue even when you catch them.

johannaSommer (approved):
👸🏻

docs/user_manual/customize_metric.rst
```diff
@@ -36,7 +36,11 @@ Implementing a ``BaseMetric``
...
+
+Please update the ``runs_on`` if you want to change the default devices. For instance, you can also include ``accelerate`` if your metric logic works on multiple devices. Inference with ``accelerate`` should work out of the box, and some of |pruna|'s metrics already use it!
```
sharpenb: I am not sure about the phrasing here. It sounds like we want to change the run_on attribute of metrics that we already implemented in pruna. If people create their metrics, they should fill this attribute. If this is a pruna metric, we should have put alreadyt the proper attributes.

sharpenb: We can probably simply rephrase

docs/user_manual/customize_metric.rst
```diff
@@ -72,14 +77,20 @@ Implementing a ``StatefulMetric``
 
 To implement a ``StatefulMetric``, create a class that inherits from ``StatefulMetric``. These metrics are designed to accumulate state across multiple batches and can share inference with other metrics.
 
-Your metric should have a ``metric_name`` attribute and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+Your metric should have a ``metric_name`` and a ``higher_is_better`` attribute. Higher is better is a boolean value that indicates if a higher metric value is better.
+
+Every metric has a ``runs_on`` attribute that should be a list that contains the device(s) the metric can run on. For base metrics it is by default ``["cuda", "cpu", "mps"]``.
+
+Please update the ``runs_on`` if you want to change the default devices.
```
sharpenb: Maybe rephrase as well. I think the term "update" is confusing

src/pruna/config/compatibility_checks.py
```diff
@@ -36,19 +36,23 @@ def ensure_device_consistency(model, smash_config):
     """
     model_device = get_device(model)
 
+    # to handle the device cases like "cuda:0 and cuda, cuda:1"
```
sharpenb: cuda, cuda:1 is unclear. Is that a typo?

src/pruna/engine/utils.py
sharpenb: I was also confused by the naming. We can probably call it set_to_available_device

src/pruna/evaluation/metrics/metric_sharpness.py
```diff
@@ -161,3 +156,18 @@ def compute(self) -> MetricResult:
         # Otherwise, compute the mean sharpness score over all accumulated samples.
         mean_val = float(np.mean(self.scores))
         return MetricResult(self.metric_name, self.__dict__, mean_val)
+
+    def move_to_device(self, device: str | torch.device) -> None:
```
sharpenb: This function might be pretty similar for all metrics which do not use a model or so. Would that make sense to factor this in the parent class and overwrite only when needed.

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -165,6 +168,16 @@ def prepare_model(self, model: Any) -> PrunaModel:
...
             )
 
+        ensure_device_consistency(model, self.task)
+        model_device = get_device(model)
+
+        # The device map is set before smashing, so for the base models, we need to set it here.
+        if model_device == "accelerate" and is_base and model.smash_config.device_map is None:
+            model.smash_config.device_map = get_device_map(model)
+
+        self.device = self.task.device
```
sharpenb, line 182: Since we should now reject if the task and model device are not consistent, it is not clear when this will be used. The comment bloew suggest in the agent only, is that the optimization agent?

src/pruna/evaluation/task.py
```diff
@@ -49,21 +49,29 @@ class Task:
...
         request: str | List[str | BaseMetric | StatefulMetric],
         datamodule: PrunaDataModule,
         device: str | torch.device | None = None,
+        low_memory: bool = False,
     ) -> None:
-        device = set_to_best_available_device(device)
-        self.metrics = get_metrics(request, device)
+        self.auto_device = device is None  # We set this for the compatibility checks later on.
+        self.low_memory = low_memory  # When we wish to be memory efficient on stateful metrics.
+        self.device = set_to_best_available_device(
```
sharpenb, line 67: I am always a bit confused by best_available_device since here we should just use the device of the user.

tests/evaluation/test_device_compatibility.py
```diff
@@ -0,0 +1,185 @@
...
+@pytest.mark.skipif(torch.cuda.device_count() < 2, reason="needs ≥2 GPUs to build a split model")
+@pytest.mark.parametrize(
+    "datamodule_fixture, model_fixture, evaluation_request, low_memory",
+    [
+        ("LAION256",  "sd_tiny_random", list(('latency', 'cmmd', 'psnr', 'pairwise_clip_score')), False),
+        ("LAION256", "sd_tiny_random", list(('latency', 'cmmd', 'psnr', 'pairwise_clip_score')), True),
+    ],
+    indirect=["datamodule_fixture", "model_fixture"],
+)
+def test_auto_device_task_adapts_to_accelerate_model(datamodule_fixture: PrunaDataModule, model_fixture: Any, evaluation_request: List[str], low_memory: bool):
```
sharpenb, line 27: These tests are a bit heavy to understand. It does test that if the model is on accelerate and then we use auto_device option, metrics are on the rigth devices? Where is the auto device on? Makes sense to have it explicit. Maybe also make sense to add comment on the asserts to explain what si tested/expected.

tests/evaluation/test_task.py
```diff
@@ -30,4 +38,69 @@ def make_mock_metric(metric_class):
...
+        ("cpu", "cuda", "accelerate"),
+        ("cuda", "cpu", "cuda"),
+        ("cuda", "cpu", "cpu"),
+        ("cuda", "cpu", "accelerate"),
+        ("cuda", "cuda", "cuda"),
+        ("cuda", "cuda", "accelerate"),
+        ("cuda", "cuda", "cpu"),
+    ],
+)
+def test_metric_device_adapts_to_task_device(inference_device: str, stateful_metric_device: str, task_device: str):
```
sharpenb, line 81: Some comments would help to understand all these tests

sharpenb (commented):
Thanks for this big udpate! I left some comments. Since the PR is huge, I might have missed some points.

docs/user_manual/customize_metric.rst
guennemann: I agree. Please update the phrasing a bit.

src/pruna/evaluation/metrics/metric_stateful.py
```diff
@@ -31,14 +33,27 @@ class StatefulMetric(ABC):
...
 
-    def __init__(self) -> None:
+    def __init__(self, device: str | torch.device | None = None, **kwargs) -> None:
         """Initialize the StatefulMetric class."""
         super().__init__()
+        self.device = set_to_best_available_device(device)
+        if not self.is_device_supported(self.device):
+            raise ValueError(
+                f"Metric {self.metric_name} does not support device {self.device}. Must be one of {self.runs_on}."
+            )
```
guennemann, line 56: This part now actually confused me. You first assign the device to the "best available one" (I know, we are discussing this function in general) and later only check whether it is supported. Shouldn't the "best available" include the supported devices, i.e. "best available and supported device"?

src/pruna/config/compatibility_checks.py
begumcig: Not a typo 🙂. This covers cases where the model’s device might be "cuda:0", while the smash config was set manually to just "cuda". We normalize these so they don’t trigger a false error.
However, if the devices are actually different (e.g. "cuda:0" vs "cuda:1"), then we correctly raise an error.

src/pruna/evaluation/evaluation_agent.py
begumcig: If the user doesn’t explicitly specify a device for the Task, the set_to_best_available function assigns the task to the best available device. When the EvaluationAgent is created with this task, it inherits that device.

However, we decided that when no device is explicitly set by the user, the model’s device should take precedence over the “best available” device. That means if the task device is later updated once we pass a model to the EvaluationAgent, the agent’s device should also be updated accordingly.

So self.device = self.task.device ensures that the agent always stays in sync with the task’s current device.

src/pruna/evaluation/metrics/metric_stateful.py
begumcig: set_to_best_available_device only checks what devices are available on the machine (e.g. prefer CUDA if present, otherwise fall back to CPU). It’s a general engine utility and doesn’t know about the evaluation context or what specific metrics support.

That’s why we still need the explicit check in the metric itself; to ensure that even if a device is technically available on the machine, the metric can actually run on it.

PR #239, opened 2025-07-08
feat: new quantizer for vllm

llcnt:
## Description
This PR is the little sister of this [PR]() in pruna_pro.
This PR in pruna add some flexibility around hqq quantization and saving:
- a new parameter `patch_for_inference` is added to allow the user to quantize in 4bits without patching the layers (because can make them un-loadable);
- a new parameter `default_to_hf` allows the user to select a specific implementation of hqq quantization (previously only a try:...except was done, but no freedom was given to the user). this is important in particular because vllm expects a models that has been saved with the 'transformers' structure, not the AutoHQQModel one.

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
Unit test pass locally.

## Checklist

### DIFF
--- a/src/pruna/algorithms/quantization/hqq.py
+++ b/src/pruna/algorithms/quantization/hqq.py
@@ -84,6 +84,12 @@ class HQQQuantizer(PrunaQuantizer):
                 default_value=True,
                 meta=dict(desc="Whether to patch the model for inference (with torchaoint4 kernels)."),
             ),
+            CategoricalHyperparameter(
+                "default_to_hf",
+                choices=[True, False],
+                default_value=False,
+                meta=dict(desc="Whether or not to bypass the HQQ quantization and use the generic HF quantization."),
+            ),
         ]
 
     def model_check_fn(self, model: Any) -> bool:
@@ -129,14 +135,20 @@ class HQQQuantizer(PrunaQuantizer):
         safe_memory_cleanup()
         with ModelContext(model) as (pipeline, working_model, denoiser_type):
             try:  # Try to quantize the model using HQQ
+                if smash_config["default_to_hf"]:
+                    raise Exception(
+                        "AutoHQQHFModel is bypassed, defaulting to generic HF quantization. "
+                        "Set default_to_hf to False to (try to) use AutoHQQHFModel."
+                    )
                 working_model = imported_modules["AutoHQQHFModel"].quantize_model(
                     working_model,
                     quant_config=quant_config_hqq,
                     device=smash_config["device"],
                     compute_dtype=torch.float16 if smash_config["compute_dtype"] == "torch.float16" else torch.bfloat16,
                 )
-            except Exception:  # Default to generic HF quantization if it fails
-                pruna_logger.info("Could not quantize model using specialized HQQ pipeline, trying generic interface...")
+            except Exception:  # Default to generic HF quantization if it fails or if default_to_hf is True
+                if not smash_config["default_to_hf"]:
+                    pruna_logger.info("Could not quantize model using specialized HQQ pipeline, trying generic interface...")
                 # Create a temporary directory in a specific location
                 base_temp_dir = smash_config["cache_dir"]
                 temp_dir = tempfile.mkdtemp(dir=base_temp_dir)
--- a/src/pruna/algorithms/quantization/hqq.py
+++ b/src/pruna/algorithms/quantization/hqq.py
@@ -78,6 +78,12 @@ class HQQQuantizer(PrunaQuantizer):
                 default_value="torch.float16",
                 meta=dict(desc="Compute dtype for quantization."),
             ),
+            CategoricalHyperparameter(
+                "patch_for_inference",
+                choices=[True, False],
+                default_value=True,
+                meta=dict(desc="Whether to patch the model for inference (with torchaoint4 kernels)."),
+            ),
         ]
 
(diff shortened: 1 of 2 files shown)


src/pruna/algorithms/quantization/hqq.py
```diff
@@ -78,6 +78,18 @@ def get_hyperparameters(self) -> list:
                 default_value="torch.float16",
                 meta=dict(desc="Compute dtype for quantization."),
             ),
+            CategoricalHyperparameter(
```
sharpenb: As is, the description fo these two aprameters are unclear to me. I think that it will be confusing for users. What would be the impact for them? We can probably find better name and description.

src/pruna/algorithms/quantization/hqq.py
```diff
@@ -123,14 +135,22 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         safe_memory_cleanup()
         with ModelContext(model) as (pipeline, working_model, denoiser_type):
             try:  # Try to quantize the model using HQQ
+                if smash_config["default_to_hf"]:
```
sharpenb: When is that important for the user to bypass AutoHQQHFModel?

sharpenb: Could we maybe find a way to avoir having this additional hyperparameter?

src/pruna/algorithms/quantization/hqq.py
```diff
@@ -123,14 +135,22 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
                 working_model = imported_modules["AutoHQQHFModel"].quantize_model(
                     working_model,
                     quant_config=quant_config_hqq,
                     device=smash_config["device"],
                     compute_dtype=torch.float16 if smash_config["compute_dtype"] == "torch.float16" else torch.bfloat16,
                 )
-            except Exception:  # Default to generic HF quantization if it fails
-                pruna_logger.info("Could not quantize model using specialized HQQ pipeline, trying generic interface...")
+            except Exception:  # Default to generic HF quantization if it fails or if default_to_hf is True
+                if not smash_config["default_to_hf"]:
```
sharpenb: Same questions/thoughts here :)

sharpenb: Do we need this log as well? It complexify the implementation.

src/pruna/algorithms/quantization/hqq.py
```diff
@@ -149,7 +169,12 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
 
             # Prepare the model for fast inference
             try:
-                if weight_quantization_bits == 4:
+                if weight_quantization_bits == 4 and smash_config["patch_for_inference"]:
+                    pruna_logger.info(
+                        "Patching model for fast inference with torchaoint4 kernels. "
+                        "This operation can make the model incompatible with re-load. "
+                        "If you plan to save and re-load the model, set patch_for_inference to False."
```
sharpenb: This comes to the cost to no/less speedup when reloading, right?

sharpenb (commented):
This is functional codeso it is approved. Some comments that would be important to consider before merging:
- Simplify/Clairfy the interface in this PR
- Test in this PR the compatibility fo the new introduced configs with the compatible algorithms in HQQ.
- Add step in the roadmap which corresponds to refactorization of the quantization code aiming at homogenizing the quant UX between quantizers, and the quant features (e.g. save/load), making easier the implementation of custom quants.

src/pruna/algorithms/quantization/hqq.py
llcnt: In the context of vllm, we need to bypass AutoHQQ and use AutoModel (because the former produces a qmodel.pt while vllm needs .safetensors)

llcnt: I did't find a clean way to avoid the use of additional hyperparameter. If you have any idea I will be happy to add it :)
In general, do you remember why we still have this try...except with AutoHQQ and AutoModel in hqq.py? Why didn't we have since the beginning a hyperparameter to let the user select which type of implementation s/he want to use?

src/pruna/algorithms/quantization/hqq.py
llcnt: I change it. I hope it is more clear now ;)

src/pruna/algorithms/quantization/hqq.py
llcnt: "this log", you mean the "Could not quantize model using..."? If yes, it was already here before ;)

llcnt:
> This is functional codeso it is approved. Some comments that would be important to consider before merging:
> 
>     * Simplify/Clairfy the interface in this PR
> 
>     * Test in this PR the compatibility fo the new introduced configs with the compatible algorithms in HQQ.
> 
>     * Add step in the roadmap which corresponds to refactorization of the quantization code aiming at homogenizing the quant UX between quantizers, and the quant features (e.g. save/load), making easier the implementation of custom quants.

I re-ran locally all unit test related to hqq, and there are valid (the 2 hyperparameters introduced have some default value that yields to the exact previous code) :)

johnrachwan123:
bugbot run

johnrachwan123 (approved):
LGTM

sharpenb (approved):
Reapproving

PR #243, opened 2025-07-09
fix: change janus import into automodel import

llcnt:
## Description
This PR makes use and replace the `JanusForConditionalGeneration` by a `AutoModelForImageTextToText` : in case there is a package version mismatch, only the Janus test would fail with some model-not-available-error.

## Related Issue

The previous PR #145 on janus introduced some test that yields import error if the `transformers` is outdated.

## Type of Change

## How Has This Been Tested?

Local test on Janus pass

## Checklist

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -89,7 +89,7 @@ def get_diffusers_model(model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, S
     return model, smash_config
 
 
-def get_automodel_transformers(model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, SmashConfig]:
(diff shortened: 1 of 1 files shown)


tests/fixtures.py
```diff
@@ -119,7 +119,7 @@ def get_torchvision_model(name: str) -> tuple[Any, SmashConfig]:
 
 def get_janus_model(model_id: str) -> tuple[Any, SmashConfig]:
```
johannaSommer: can we just make the function name and docstring more general? then we can reuse this function similar to the automodel function :)

llcnt: Yes good idea :) I changed it, and I also adapted the `get_automodel_transformers` to explicitely mention causalLM

llcnt: I have reverted get_automodel_transformers name to avoid issues with pruna_pro

PR #246, opened 2025-07-10
feat: run test cases in parallel

GreatBahram:
## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -69,6 +69,9 @@ jobs:
 
       - uses: ./.github/actions/setup-uv-project
 
+      - name: Set test worker count
+        run: echo "PYTEST_WORKERS=$(( $(nproc) - 1 ))" >> "$GITHUB_ENV" # leave 1 for the system process
+
       - name: Run tests
         run: |
-          pytest -n 2 -m "cpu and not slow and not high_cpu"
+          pytest -n ${{ env.PYTEST_WORKERS }} -m "cpu and not slow and not high_cpu"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -71,4 +71,4 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -m "cpu and not slow and not high_cpu"
+          pytest -n 2 -m "cpu and not slow and not high_cpu"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -58,6 +58,9 @@ jobs:
       matrix:
         python-version: ["3.10"]
 
+    env:
+      HF_ENDPOINT: https://hf-mirror.com
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ addopts = [
 python_files = "test_*.py"
 
 testpaths = [
-    "src/pruna/tests",
+    "tests/",
 ]
 
 [[tool.uv.index]]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,6 +158,7 @@ dev = [
     "mypy",
     "types-PyYAML",
     "logbar",
+    "pytest-xdist>=3.8.0",
 ]
 cpu = []
 
--- a/tests/algorithms/testers/caching.py
+++ b/tests/algorithms/testers/caching.py
@@ -43,7 +43,8 @@ class TestFasterCache(AlgorithmTesterBase):
 
     def post_smash_hook(self, model: PrunaModel) -> None:
         """Hook to modify the model after smashing."""
-        assert model.transformer.is_cache_enabled
+        if hasattr(model, "transformer"):
+            assert model.transformer.is_cache_enabled
 
 
 class TestPAB(AlgorithmTesterBase):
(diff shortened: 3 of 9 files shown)


tests/fixtures.py
```diff
@@ -82,6 +82,8 @@ def dummy_model() -> tuple[Any, SmashConfig]:
 def get_diffusers_model(model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, SmashConfig]:
     """Get a diffusers model for image generation."""
     # snapshot download of the model
+    # TODO: is there a hugging face credential that we can use?
```
GreatBahram: It seems that when we run them in parallel, Hugging Face may not be very happy about it. Here are some alternatives:

1. Use a Hugging Face account by storing the credentials somewhere in the pipeline and logging in.
2. Download the necessary models and load them into the test as we iterate through them. I'm not sure about their size; perhaps we could create a Makefile to handle the downloads, preventing them from being added directly into the source code. We could then call this Makefile before executing the tests to download all the models, and subsequently run the test cases. 🤔
3. Cancel the test case execution in parallel

davidberenstein1957: Perhaps we can add some cache?

      - name: Cache HuggingFace Transformers
        uses: actions/cache@v3
        with:
          path: |
            ~/.cache/huggingface
            ~/.cache/torch/transformers
          key: ${{ runner.os }}-huggingface-${{ hashFiles('**/poetry.lock', '**/requirements.txt') }}
          restore-keys: |
            ${{ runner.os }}-huggingface-

davidberenstein1957: Awesome, you already opened a PR for that :)

GreatBahram: Thank you very much, David :sunflower:

GreatBahram:
My takeaway so far:

- The mirror feature worked and everything seems to be fine with it.
- The full_integration test case is a bit complex; I don't understand it at the moment and takes like a minute or so to run and it fails at the end :shrug:

johannaSommer:
hey @GreatBahram thanks a lot for all your efforts so far! I am catching up with this PR just now, it seems you have figured out how to manage the HF Hub to run the test cases in parallel and now the way the tests are set up (`path="saved_model/`) is causing you some troubles, is that correct? Let me know if I can help in any way, we can definitely adjust this saving mechanism 👍🏻

GreatBahram:
> hey @GreatBahram thanks a lot for all your efforts so far! I am catching up with this PR just now, it seems you have figured out how to manage the HF Hub to run the test cases in parallel and now the way the tests are set up (`path="saved_model/`) is causing you some troubles, is that correct? Let me know if I can help in any way, we can definitely adjust this saving mechanism 👍🏻

:wave: Hi Johanna, 

Thank you for your support :sunflower:! Today, I found some time to push [this solution](https://github.com/PrunaAI/pruna/pull/246/commits/89fe1d005ae0fcd2012ad10cb2ff915eaba869cf). It would be great if you could take a look since I changed some class variables and methods to instance ones; I hope that's alright.

I noticed that with these changes, the test time has decreased. Previously, it took about 4 minutes and 30 seconds on average, and now it's down to 3 minutes (approximately 30% improvement)

Before improvement:
before the improvement
After improvement:
after-improvement

.github/workflows/tests.yaml
```diff
@@ -69,6 +72,9 @@ jobs:
 
       - uses: ./.github/actions/setup-uv-project
 
+      - name: Set test worker count
+        run: echo "PYTEST_WORKERS=$(( $(nproc) - 1 ))" >> "$GITHUB_ENV" # leave 1 for the system process
+
       - name: Run tests
         run: |
-          pytest -m "cpu and not slow and not high_cpu"
+          pytest -n ${{ env.PYTEST_WORKERS }} -m "cpu and not slow and not high_cpu"
```
GreatBahram, line 80: > [!TIP]
> We can also update the documentation to state one can use `pytest -n auto` to run the test cases locally in parallel.

pyproject.toml
```diff
@@ -42,7 +42,7 @@ addopts = [
 python_files = "test_*.py"
 
 testpaths = [
-    "src/pruna/tests",
```
GreatBahram, line 45: I noticed this during the development and I guess it's not worth a separate PR (?)

tests/algorithms/testers/caching.py
```diff
@@ -43,7 +43,8 @@ class TestFasterCache(AlgorithmTesterBase):
 
     def post_smash_hook(self, model: PrunaModel) -> None:
         """Hook to modify the model after smashing."""
-        assert model.transformer.is_cache_enabled
+        if hasattr(model, "transformer"):
```
johannaSommer: Just out of curiosity, was this change necessary for you to make this test pass?

johannaSommer (approved):
Looks flawless to me! I think the change to make the `saving_path` an instance variable in this case makes complete sense and should not break anything else!

tests/algorithms/testers/caching.py
GreatBahram: I don't think it's necessary; I'll remove it and see if we really need to keep it

GreatBahram: Seems we didn't need it, thanks for raising it, Johanna :sunflower: 
https://github.com/PrunaAI/pruna/pull/246/commits/eaba158c40cb056a92db5e303d078d0a42f5c683

davidberenstein1957 (approved):

PR #249, opened 2025-07-10
Feature: Cache the Hugging Face directory so it doesn't download the test model each time

GreatBahram:
## Description
This idea stems from [this PR](https://github.com/PrunaAI/pruna/pull/246), the thing is when we run the test cases in parallel these hosts (huggingface & torch) are not happy about it.

So I searched around, it seems people cache the data to prevent that and make the pipeline more efficient.

First, I test it with hugging-face if succeeds; then we can extend it to torch too (while never got any complain from torch :shrug:).
```
      - name: Cache Torch models
        uses: actions/cache@v4
        key: torch
        with:
          path: ~/.cache/torch
```

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -71,6 +71,7 @@ jobs:
 
       - name: Cache models
         uses: actions/cache@v4
+        key: huggingface
         with:
           path: ~/.cache/huggingface
 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -69,6 +69,11 @@ jobs:
 
       - uses: ./.github/actions/setup-uv-project
 
+      - name: Cache models
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+
       - name: Run tests
         run: |
           pytest -m "cpu and not slow and not high_cpu"


.github/workflows/tests.yaml
```diff
@@ -69,6 +69,12 @@ jobs:
 
       - uses: ./.github/actions/setup-uv-project
 
+      - name: Cache models
+        uses: actions/cache@v4
```
GreatBahram, line 73: https://github.com/actions/cache

davidberenstein1957: great, you already found this 🔥

davidberenstein1957: @GreatBahram could you see and check if there is some way to reset the cache every x perio (let's say 3 months)?

GreatBahram:
🥹:

Warning: Failed to save: Cache size of ~34594 MB (36274943792 B) is over the 10GB limit, not saving cache.

.github/workflows/tests.yaml
```diff
@@ -72,7 +72,7 @@ jobs:
       - name: Cache models
         uses: actions/cache@v4
         with:
-          path: ~/.cache/huggingface
+          path: ~/.cache/huggingface/hub/
```
GreatBahram, line 75: This reduced it 25GB, but still too much :exploding_head: 

`Warning: Failed to save: Cache size of ~25675 MB (26922483338 B) is over the 10GB limit, not saving cache.`

GreatBahram:
Before we give up on this, it seems one can use a different mirror by setting the env variable:

https://github.com/huggingface/huggingface_hub/blob/83bcc0bc40619bd68bfd6af51329f7afd9eed998/src/huggingface_hub/constants.py#L67

and this Chinese website is hosting the models :thinking: :

https://hf-mirror.com/

I'll give it try to see if it works and then we can decide if it is worth or not.

davidberenstein1957:
@GreatBahram, I think the mirror could be an option if we only use it for our tests.

[This PR](https://github.com/PrunaAI/pruna/pull/250) might also resolve some of the issues since I am removing bigger models. I also added these updates (for example, removing BFL) from the cache. Maybe we can enforce a couple of big models not to be added to the cache? I assume we have issues with the number of requests and we already save some time if we only have to download some of the bigger models.

GreatBahram:
Made some progress on [the main PR](https://github.com/PrunaAI/pruna/pull/246) and I close this one.

PR #260, opened 2025-07-17
feat: replace os.path with pathlib.Path

GreatBahram:
## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


tests/common.py
```diff
@@ -173,29 +174,27 @@ def get_all_imports(package: str) -> set[str]:
...
-                        continue
-
-                    # Handle __init__.py: its module is the package (or subpackage) itself.
-                    if file == "__init__.py":
-                        module_name = module_name.rsplit(".", 1)[0]  # Remove the trailing '__init__'
-
-                    # Combine the base package name with the module path.
-                    # If module_name is empty (i.e. __init__.py at the root), just use the package name.
-                    full_import = f"{package}.{module_name}" if module_name else package
-                    imports.add(full_import)
```
GreatBahram, line 198: Here, I feel something was wrong; as we first check `module_name == '__init__'` I feel we never hit the second if-stmt :thinking:

src/pruna/algorithms/compilation/c_translate.py
```diff
@@ -190,14 +190,14 @@ def load_model(self: Any, model_class: Any, model_name_or_path: str, **kwargs: A
 
         converter.convert_from_args(args)
         if self.task_name == "translate":
-            optimized_model = imported_modules["Translator"](temp_dir, device=smash_config["device"])
-            optimized_model = TranslatorWrapper(optimized_model, temp_dir, smash_config.tokenizer)
+            optimized_model = imported_modules["Translator"](str(temp_dir), device=smash_config["device"])
```
davidberenstein1957: might be cleaner to just convert this to `str` once?

src/pruna/algorithms/compilation/c_translate.py
```diff
@@ -212,26 +212,26 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
-                os.path.join(os.path.dirname(sys.executable), "lib"),
+                Path("/usr/local/cuda/lib64"),
+                Path("/usr/lib/x86_64-linux-gnu"),
+                Path("/usr/lib/cuda/lib64"),
+                Path(sys.executable).parent / "lib",
             ]
 
             for path in possible_paths:
-                if os.path.exists(path) and any(f.startswith("libcudnn") for f in os.listdir(path)):
+                if path.exists() and any(str(f).startswith("libcudnn") for f in path.iterdir()):
```
davidberenstein1957, line 236:
```suggestion
                if path.exists() and any(f.is_relative_to("libcudnn") for f in path.iterdir()):
```

src/pruna/config/smash_config.py
```diff
@@ -50,6 +49,7 @@
 PROCESSOR_SAVE_PATH = "processor/"
 SMASH_CONFIG_FILE_NAME = "smash_config.json"
```
davidberenstein1957, line 50: should we define these as Path directly too?

src/pruna/data/datasets/text_to_image.py
```diff
@@ -80,22 +80,22 @@ def setup_coco_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
         The COCO dataset.
     """
     # original license Creative Commons Attribution 4.0 License (CC BY 4.0)
-    directory_dataset = os.path.join(config.HF_DATASETS_CACHE, "coco")
-    if not os.path.exists(directory_dataset):
-        os.makedirs(directory_dataset)
+    directory_dataset = Path(config.HF_DATASETS_CACHE) / "coco"
```
davidberenstein1957, line 83: should HF_DATASET_CACHE be directly defined as path too?

src/pruna/engine/load.py
```diff
@@ -238,13 +237,16 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
...
+    path = Path(path)
+    pipeline_info_path = path / PIPELINE_INFO_FILE_NAME
+    if pipeline_info_path.exists():
+        with pipeline_info_path.open("r") as f:
             pipeline_info = json.load(f)
         # transformers discards kwargs automatically, no need for filtering
         return pipeline(pipeline_info["task"], str(path), **kwargs)
     else:
-        with open(os.path.join(path, "config.json"), "r") as f:
+        config_path = path / "config.json"
```
davidberenstein1957, line 248: perhaps we can add all these `*.json` at the top of this script? Not sure if it improves when we implement this more consistentyl.
e.g.
```
CONFIG_FILE_NAME = "config.json"
MODEL_INDEX_FILE_NAME = "model_index.json"
```

src/pruna/algorithms/compilation/c_translate.py
GreatBahram: Thanks, David! tbh, I'm not sure if this code is correct, since each time `f` contains the real path, I wonder if we should do `f.name.startswith`; it seems here we are looking for some `.so` files :thinking:; love to hear your thoughts on this :sunflower: 

image

src/pruna/algorithms/compilation/c_translate.py
GreatBahram: Thanks, fixed here:
https://github.com/PrunaAI/pruna/pull/260/commits/930c5dc76ec833067084b8b9b0d20cc7b13a5880

src/pruna/config/smash_config.py
GreatBahram: I checked the codebase these are fine, since they're not a complete path, but we join them with other PathLike objects :pray:

src/pruna/data/datasets/text_to_image.py
GreatBahram: It seems this one is coming from a third-party library [[link](https://github.com/PrunaAI/pruna/blob/2ee12e09ea1c0c6fa1ad63110d63aa8ae0c880c1/src/pruna/data/datasets/text_to_image.py#L21)]

src/pruna/engine/load.py
GreatBahram: Honestly, not sure. I took a look it seems when they are used elsewhere people defined them globally, I don't have enough context to make the decision here, wdyt?

https://github.com/PrunaAI/pruna/blob/6217fee1e29cf553aa1539ed2ae0d55c201181f8/src/pruna/engine/load.py#L36-L38

src/pruna/algorithms/pruna_base.py
```diff
@@ -200,7 +200,7 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
             The model after the algorithm has been applied.
         """
         if self.save_fn == SAVE_FUNCTIONS.save_before_apply and smash_config._prepare_saving:
-            save_dir = os.path.join(smash_config.cache_dir, SAVE_BEFORE_SMASH_CACHE_DIR)
+            save_dir = Path(smash_config.cache_dir) / SAVE_BEFORE_SMASH_CACHE_DIR
```
johannaSommer: would it make sense to directly keep the cache_dir as a `Path` object in the smash config to avoid calls like this?

johannaSommer: okay I checked below and it seems that is the case (love it), in that case we can remove the Path casting here, right?

src/pruna/algorithms/batching/ws2t.py
```diff
@@ -135,12 +135,12 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         elif task == "whisper_ct" and hasattr(processor.tokenizer, "name_or_path"):
             # this requires a little trick, a transformers tokenizer can not be directly converted
             # we can either go via a download or via a file and then parsing out the Tokenizer
-            if os.path.exists(processor.tokenizer.name_or_path):
+            if Path(processor.tokenizer.name_or_path):
```
johannaSommer: from what I understand we would need `if Path("..").exists()`?

src/pruna/engine/load.py
johannaSommer: catching up on this one - if it is only about the "config.json" and "model_index.json" both of those come from huggingface diffusers/ transformers, they will always be called that and we scan for these files. If we can find these constants in diffusers/transformers, I would import them and use those, wdyt?

src/pruna/evaluation/metrics/metric_energy.py
```diff
@@ -84,16 +84,16 @@ def compute(self, model: PrunaModel, dataloader: DataLoader) -> Dict[str, Any] |
...
+        model.save_pretrained(str(save_path))
 
         tracker = EmissionsTracker(project_name="pruna", measure_power_secs=0.1)
         tracker.start()
 
         # Measure the loading energy
         tracker.start_task("Loading model")
         temp_model = model.__class__.from_pretrained(
-            save_path,
+            str(save_path),
```
johannaSommer: I might be reading this wrong but I think this model is a Pruna Model (see type hint), in which case we support the save_path being a Path, correct? so possible no need for this string wrapping (also see changes below if this is correct)

tests/common.py
```diff
@@ -173,29 +174,17 @@ def get_all_imports(package: str) -> set[str]:
...
+        pkg_path = Path(location)
```
johannaSommer, line 176: I love it!!!!

johannaSommer:
One more question - [ruff has support ](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth)for `pathlib` linting, let's add this to make sure we stay this consistent in the future? To you think this would be possible / would make sense?

src/pruna/algorithms/compilation/c_translate.py
davidberenstein1957: I agree. Perhaps your solution is better. I figured it might be a good solution but did not directly check. Feel free to ignore this.

davidberenstein1957 (approved):
Thanks a lot for resolving my nits. Feel free to merge after resolving the comment of Johanna.

src/pruna/algorithms/pruna_base.py
GreatBahram: Oh yes, you're right—I’ll go ahead and remove it :saluting_face:

src/pruna/evaluation/metrics/metric_energy.py
GreatBahram: No, you're absolutely right, my bad

src/pruna/engine/load.py
GreatBahram: I found [one of them](https://github.com/huggingface/transformers/blob/5a81d7e0b388fb2b86fc1279cdc07d9dc7e84b4c/src/transformers/utils/__init__.py#L295) which is stored in a global variable; but the other one sadly not [[link](https://github.com/search?q=repo%3Ahuggingface%2Fdiffusers+model_index.json&type=code)]; if you don't mind I'll keep them as they're, what do you think?

src/pruna/evaluation/metrics/metric_energy.py
GreatBahram: Fixed here:

https://github.com/PrunaAI/pruna/pull/260/commits/f4e15af73b3989878347cdd2d643c727c214e320

src/pruna/algorithms/batching/ws2t.py
GreatBahram: Fixed here:
https://github.com/PrunaAI/pruna/pull/260/commits/f4e15af73b3989878347cdd2d643c727c214e320

src/pruna/algorithms/pruna_base.py
GreatBahram: Fixed here:
https://github.com/PrunaAI/pruna/pull/260/commits/f4e15af73b3989878347cdd2d643c727c214e320

pyproject.toml
```diff
@@ -3,8 +3,8 @@ exclude = ["test_*.py"]
 line-length = 121
 
 [tool.ruff.lint]
-ignore = ["ANN002", "ANN003", "ANN401", "C901", "D100", "D104", "D401", "D406", "C408"]
-select = ["A", "C", "CPY", "D", "E", "ERA", "F", "FIX", "I", "N", "SIM", "T20", "W", "FA"]
+ignore = ["ANN002", "ANN003", "ANN401", "C901", "D100", "D104", "D401", "D406", "C408", "PTH123"]
```
GreatBahram, line 6: I ignored `PTH123` as that is for the builtin open function which is fine :shrug:

GreatBahram:
> One more question - [ruff has support ](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth)for `pathlib` linting, let's add this to make sure we stay this consistent in the future? To you think this would be possible / would make sense?

I added the rule to the ruff here, we can give it a try if it bothers people then we can remove it in the future

https://github.com/PrunaAI/pruna/pull/260/commits/db89d65ae78aed1df39d18b136d31e429974c83b

PR #263, opened 2025-07-18
feat: add unconstrained hyperparameter

gsprochette:
## Description

Separate smash_space and custom hyperparameters, and add a new hyperparameter class:
- the user can give any value to this parameter in smash_config
- ConfigSpace can still handle the dependencies between hyperparameters

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,72 @@
...
+        super().__init__(name, choices=[True, False], default_value=default, meta=meta)
+
+    def __new__(cls, name: str, default: bool = False, meta: Any = None) -> CategoricalHyperparameter:  # type: ignore
+        """Create a new boolean hyperparameter."""
+        return CategoricalHyperparameter(name, choices=[True, False], default_value=default, meta=meta)
+
+
+class UserEditable(Constant):
+    """
+    Represents a hyperparameter that is freely set by the user within a pre-specified type.
```
sharpenb: This docstring is a bit confusing to me. By "freely", we mean that it is not constraint by a predefined space, right? If yes, I am not sure about the name UserEditable. Maybe SpaceUnconstrainedParameter (or sthg like this) would be more transparent.

src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,72 @@
...
+
+from __future__ import annotations
+
+from typing import Any
+
+from ConfigSpace import CategoricalHyperparameter, Constant
+from typing_extensions import override
+
+
+class Boolean(CategoricalHyperparameter):
```
sharpenb, line 23: Having a proper dedicated file with these extended hyperparameters makes indeed sense :)

sharpenb (commented):
Thanks for this PR! This is a really important one :)

src/pruna/config/hyperparameters.py
gsprochette: Renamed to UnconstrainedHyperparameter

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -102,28 +140,36 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
         """
         imported_modules = self.import_algorithm_packages()
-
-        if hasattr(model, "unet"):
-            working_model = model.unet
-        elif hasattr(model, "transformer"):
-            working_model = model.transformer
-        else:
-            working_model = model
+        target_modules = smash_config["target_modules"] or self.get_default_hyperparameters(model, smash_config)[0]
```
sharpenb: I was wondering if we could get rid of this 0 index by removing () in the return of get_default_hyperparameters :)

src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,188 @@
...
+                f"Target modules must be a dictionary with lists of fnmatch patterns as values. Got: {value}"
+            )
+        elif "include" not in value or len(value["include"]) == 0:
+            raise ValueError("Target modules must have at least one 'include' pattern.")
+        else:
+            all_patterns = [pattern for patterns in value.values() for pattern in patterns]
+            unrecognized_patterns = [pattern for pattern in all_patterns if not isinstance(pattern, str)]
+            if unrecognized_patterns:
+                raise TypeError(
+                    "Target modules must be a dictionary with lists of fnmatch patterns as values. "
```
sharpenb: It is not clear what is fnmatch here. It looks like we mostly check that the pattern is a string above :)

src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self,
+        name: str,
+        default_value: Any = None,
+        meta: Any = None,
+    ) -> None:
+        super().__init__(name, default_value, meta)
+
+    @override
+    def legal_value(self, value):  # numpydoc ignore=GL08
```
sharpenb, line 68: I was wondering if we could have docstrings for these legal_value as well? :)

src/pruna/algorithms/pruna_base.py
```diff
@@ -179,6 +179,24 @@ def get_hyperparameters(self) -> list:
         """
         return []
 
+    def get_default_hyperparameters(self, model: Any, smash_config: SmashConfig | SmashConfigPrefixWrapper) -> tuple:
```
johannaSommer: I think this naming can be a bit confusing because i would expect this to return e.g. also the default of the other hyperparameters like `cache_interval=2` or so. Something like `get_user_specified_hyperparameter_default` (ok thats way too long but something along those lines?

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -63,6 +66,14 @@ def get_hyperparameters(self) -> list:
             Constant("act_bits", value=None),
             Boolean("calibrate", default=True, meta=dict(desc="Whether to calibrate the model.")),
             Constant(name="calibration_samples", value=64),
+            TargetModules(
+                name="target_modules",
+                default_value=None,
+                meta=dict(
+                    desc=f"Precise choices of which modules to quantize. "
+                    f"See {TargetModules.path_to_target_modules_documentation} for more details."
```
johannaSommer: that is a great solution, thanks for that!

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -102,28 +140,36 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                    activations=activations,
+                    include=subpaths,
+                )
+            except Exception as e:
+                pruna_logger.error("Error during quantization: %s", e)
+                raise
 
         if smash_config["calibrate"]:
             if smash_config.tokenizer is not None and smash_config.data is not None:
+                if hasattr(model, "unet"):
```
johannaSommer: We should use the functionality as defined above here. I know you can't directly use the target module result here but we should use it in some capacity so that the logic for quanto is centralized if we adjust this in the future (probs we wont adjust it but you get me, for duplication reasons)

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -139,11 +185,12 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
             else:
                 pruna_logger.error("Calibration requires a tokenizer and dataloader. Skipping calibration.")
 
-        try:
-            imported_modules["freeze"](working_model)
-        except Exception as e:
-            pruna_logger.error("Error while freezing the model: %s", e)
-            raise
+        for module, _ in modules_with_subpaths:
```
johannaSommer: I would say we should only have _one_ loop through the modules_with_subpaths and then to the quanto logic. From what I understand, it makes only sense to freeze it if we actually quantized it

johannaSommer: hm nvm we raise an error... Still, maybe it would be clearer to loop _once_ and then apply the algorithm logic inside?

src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,188 @@
...
+        # edit the internal state of the Constant to allow for the new value
+        self._contains_sequence_as_value = isinstance(value, (list, tuple))
+        self._transformer.value = value
+        return super().legal_value(value)
+
+
+TARGET_MODULES_TYPE = Dict[Literal["include", "exclude"], List[str]]
```
johannaSommer: should this be at the top of the file?

src/pruna/config/hyperparameters.py
```diff
@@ -0,0 +1,188 @@
...
+        elif not isinstance(value, dict):
+            raise TypeError(f"Target modules must be a dictionary with keys 'include' and/or 'exclude'. Got: {value}")
+        elif any(key not in ["include", "exclude"] for key in value):
+            raise ValueError(f"Target modules must only use keys 'include' and/or 'exclude'. Got: {list(value.keys())}")
+        elif any(not isinstance(patterns, list) for patterns in value.values()):
+            raise TypeError(
+                f"Target modules must be a dictionary with lists of fnmatch patterns as values. Got: {value}"
+            )
+        elif "include" not in value or len(value["include"]) == 0:
+            raise ValueError("Target modules must have at least one 'include' pattern.")
```
johannaSommer: so the user can not specify only exclude? If I understand correctly do we really want that? I could see that you want everything except some specific layer

johannaSommer: In this case you can specify it with `include="*"` i guess? just wondering if this is clear to the user

johannaSommer (changes requested):
Sensational work!

Outside of the comments above, I feel one thing that we are missing is informing the user about which Target Modules they are selecting (because who can really write correct regex/fnmatch patterns lol). I think what I would be expecting is that either in a seperate function before smash, or better, in the `to_list_of_module_paths` function, we add some significant logging.
Some ideas but you know best:
- raise an error if a pattern (exclude/include) doesnt find any hits? or at least log some error?
- at either "INFO" or possibly "DEBUG" level log the modules we are actually selecting. I want to avoid that the user misspecifies the pattern and e.g. the algo isnt applied at all.

src/pruna/config/hyperparameters.py
gsprochette: You're correct. Indeed it's not super clear so I added some simple logic at the end to set ['*'] as a default in this case for include

src/pruna/config/hyperparameters.py
gsprochette: I made it clearer

src/pruna/algorithms/pruna_base.py
gsprochette: Makes sense. All the hyperparameters are user-specified so I changed it to `get_unconstrained_hyperparameter_defaults` to reference the fact that is only gets the defaults for hyperparameters which are instances of UnconstrainedHyperparameter. It's long but making it shorter would make it less explicit...

src/pruna/algorithms/quantization/quanto.py
gsprochette: In fact working model has no place here, it can simply be the model itself when it's a LLM. For diffusers, the if statement should ignore that anyway (although you can technically add a tokenizer in the smash config after the dataloader and get an error here because it doesn't run).

In all cases, we can't use parts of the model here because it's called directly on the data, meaning if we get the diffuser's transformer the prompt won't be tokenized.

As you say, we should fix that but this is for an other PR. I'll leave it in the "calibrate LLM only" state as it was before, if you agree.

src/pruna/algorithms/quantization/quanto.py
gsprochette: I agree it would be cleaner but we still need to calibrate inbetween. Calibration needs the whole quantization to already be done since we need to run the full model, meaning we can't really freeze inside the first loop. 

Here we only freeze the attributes (e.g. unet or transformer) that contain targeted modules, and the freeze function checks each nn.Module to see if it's been quantized anyway. The only reason for iterating over modules_with_subpath is to guarentee that module is a nn.Module, because for diffusers model is not a nn.Module. I changed it to get_nn_modules which is a simpler function and achieves the same thing. Is it ok for you?

src/pruna/algorithms/quantization/quanto.py
gsprochette: Sure, I loosened the return type of the method in PrunaAlgorithmBase so that I don't have to return a tuple. In the future if we need to return multiple defaults we can still make it a tuple for other algorithms.

johannaSommer (approved):
Let's goooooo

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+import torch
+from typing_extensions import override
+
+from pruna.config.hyperparameters import UnconstrainedHyperparameter
+from pruna.engine.utils import get_nn_modules
+
+TARGET_MODULES_TYPE = Dict[Literal["include", "exclude"], List[str]]
+
+
+class TargetModules(UnconstrainedHyperparameter):
```
simlang, line 29: I'm not sure if a new file for just this HyperParameter is necessary, as we already add the new hyperparameters.py file whose purpose it is to contain all extended HyperParams. The helper methods i would move to config.utils.
Very much my personal opinion - i can see how it clutters other files then - i am fine with both solutions.

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+        elif not isinstance(value, dict):
+            raise TypeError(f"Target modules must be a dictionary with keys 'include' and/or 'exclude'. Got: {value}")
+        elif any(key not in ["include", "exclude"] for key in value):
+            raise ValueError(f"Target modules must only use keys 'include' and/or 'exclude'. Got: {list(value.keys())}")
+        elif any(not isinstance(patterns, list) for patterns in value.values()):
+            raise TypeError(
+                f"Target modules must be a dictionary with lists of fnmatch patterns as values. Got: {value}"
+            )
+        else:
+            all_patterns = [pattern for patterns in value.values() for pattern in patterns]
```
simlang: this line is very hard to read

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+def expand_list_of_modules_paths(target_modules: TARGET_MODULES_TYPE, model: Any) -> List[str]:
+    """
+    Convert the target modules to a list of module paths.
+
+    Parameters
+    ----------
+    model : Any
+        The model to get the module paths from.
+    target_modules : TARGET_MODULES_TYPE
+        The target modules to convert to a list of module paths.
```
simlang, line 104: switch order to match the signature
```suggestion
    Parameters
    ----------
    target_modules : TARGET_MODULES_TYPE
        The target modules to convert to a list of module paths.
    model : Any
        The model to get the module paths from.
```

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+) -> Dict[str | None, Tuple[torch.nn.Module, List[str]]]:
+    """
+    Get torch modules within the model and their associated subpaths.
+
+    Parameters
+    ----------
+    model : Any
+        The model to get the module paths from.
+    target_modules : TARGET_MODULES_TYPE
+        The target modules to convert to a list of module paths.
```
simlang: update docstring, appears to be a copy from `expand_list_of_modules_paths`

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+            pass  # chosing a default value is left to the algorithm based on the model
+        elif "include" not in value:
+            value["include"] = ["*"]
+        elif "exclude" not in value:
+            value["exclude"] = []  # for consistency
+
+        return super().legal_value(value)
+
+
+def expand_list_of_modules_paths(target_modules: TARGET_MODULES_TYPE, model: Any) -> List[str]:
```
simlang: not super happy with the name of this function. imo a bit vague

src/pruna/config/target_modules.py
gsprochette: So my reasoning is that target modules will probably have a lot of utils function, being used in most algorithms. So I thought it made sense to have them all here separated from the "simple" hyperparameters, and this way in the files we can do something like `target_modules.do_something` and it's clearer that it's related to target modules than `utils.do_something`.

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+        The target modules to convert to a list of module paths.
+
+    Returns
+    -------
+    Dict[str | None, Tuple[torch.nn.Module, List[str]]]
+        The dictionary of modules attributes in the model with their associated targeted subpaths.
+        A module attribute which doesn't contain any targeted subpath won't be included in the dictionary.
+        Each module-subpaths pair is indexed by the module attribute name within the model.
+        Following the convention of get_nn_modules, if the model itself is a torch.nn.Module, the dictionary
+        will contain a single item with key None, pointint to the model itself and the targeted paths.
```
simlang: typo
```suggestion
        will contain a single item with key None, pointing to the model itself and the targeted paths.
```

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,211 @@
...
+    model: Any,
+    target_modules: TARGET_MODULES_TYPE,
+) -> Any:
+    """
+    Apply a function to the model, or to each of its targeted nn.Modules in the case of a Pipeline.
+
+    Parameters
+    ----------
+    apply_single_root_fn : Callable[[str | None, torch.nn.Module, List[str]], Any]
+        The function to apply to map. It must take as input the attribute name of the root in the model,
```
simlang: "The function to apply to map" Is a little bit unclear to me

src/pruna/algorithms/pruna_base.py
```diff
@@ -179,6 +179,28 @@ def get_hyperparameters(self) -> list:
...
+            The model to get the default hyperparameters from.
+        smash_config : SmashConfig
+            The SmashConfig object.
+
+        Returns
+        -------
+        Any
+            The default unconstrained hyperparameters values for the algorithm.
+        """
+        return None
```
simlang: This function is both too broad and too narrow. At the moment we only have target_modules as the only unconstrained hyperparameter, so for now we could change this to `get_target_module_defaults`.
I am however for making it applicable to any future unconstrained hyperparameter. This is not how this function is implemented, or at least not clear by the signature (as the return type is Any, technically everything is supported)
In the case where there is a future second unconstrained hyperparam, we don't know which default belongs to what.
-> change the return type to Dict[str, Any] - so for any hyperparameter we can get a default, but indexed by the param it concerns

Is there an argument to extend it to something like `get_model_dependent_hyperparameter_defaults` - why keep it constrained to the unconstrained hyperparameters? This way any parameter which might need information from the model/SmashConfig can get a default

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -102,12 +139,9 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
         imported_modules = self.import_algorithm_packages()
-        if hasattr(model, "unet"):
-            working_model = model.unet
-        elif hasattr(model, "transformer"):
-            working_model = model.transformer
-        else:
-            working_model = model
+        target_modules = smash_config["target_modules"]
+        if target_modules is None:
+            target_modules = self.get_unconstrained_hyperparameter_defaults(model, smash_config)
```
simlang: This is unnecessary duplicate code in every _apply.
I would be for moving this into the apply wrapper. Connected to the proposition to change the return type of get_unconstrained_hyperparameter_defaults.
In `pruna_base.apply` before calling _apply add something like:

```
defaults = self.get_unconstrained_hyperparameter_defaults(model, smash_config)
for key in defaults.keys() if smash_config[key] is None:
    smash_config[key] = defaults[key]
self._apply(model, smash_config)
```

src/pruna/algorithms/quantization/quanto.py
```diff
@@ -102,12 +139,9 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
             The quantized model.
         """
         imported_modules = self.import_algorithm_packages()
```
simlang, line 141: Not connected to this PR at all, but i would also see this as unnecessary code, and for a future refactoring i would propose importing the modules once in smash and storing it in self.imported_modules, Especially as sometimes this is called multiple times during smashing etc.
But as I said, not part of this PR

src/pruna/config/target_modules.py
gsprochette: Updated to "The function to apply to each root in the model."

src/pruna/config/target_modules.py
gsprochette: changed to "expand_list_of_targeted_paths"

src/pruna/config/target_modules.py
gsprochette: It was also completely overkill.

src/pruna/algorithms/pruna_base.py
gsprochette: I'm torn, because on one hand I agree with you and I do think a dictionary would be the better option, both general and very readable.
On the other hand, this will probably most ofter be used only for target modules, in which case it's much clearer to just return `default_target_modules` instead of `{"target_modules": default_target_modules}`.

Note that this is why the return type is `Any` currently, you can handle both: returning just the default hyperparam if there is a single one, and return a dict if there are multiple. The method in child classes can always make the return type more specific. I didn't want to force a heavy TypedDict class every time we redefine this method.


I agree with the `get_model_dependent_hyperparameter_defaults`, it's a better name.

gsprochette: As a conclusion: I'm in favor of
1. renaming the method
2. keep the signature as is to leave more flexibility

Refactoring this return type wouldn't be too much of a problem anyway.

src/pruna/algorithms/quantization/quanto.py
gsprochette: The reason I don't want to do that is that it modifies the smash_config in-place.
In the case where a user would use the same smash_config to smash two different models, the second would have the defaults of the first one, instead of looking for its own default.
We can discuss doing a copy of the smash config at the start of `smash`, but that would be a refactor PR and not this one.
For now, I personnally prefer the duplicated "check for default None and replace the value" which is pretty standard in python I would say. WDYT?

src/pruna/algorithms/quantization/quanto.py
gsprochette: Could make sense, but as you said, not part of this PR :)

src/pruna/algorithms/quantization/quanto.py
simlang: the in-place problem is a very good point - have not thought of this!
For now it won't hurt anyone, i still think it's cleaner - maybe part of a future algorithm application refactoring

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,213 @@
...
+) -> Any:
+    """
+    Apply a function to the model, or to each of its targeted nn.Modules in the case of a Pipeline.
+
+    Parameters
+    ----------
+    apply_single_root_fn : Callable[[str | None, torch.nn.Module, List[str]], Any]
+        The function to apply each root in the model.
+        It must take as input the attribute name of the root in the model, the nn.Module itself, and a list of
+        paths withing the root, each pointing to a targeted submodule. It must return the modified root.
```
simlang: typo
```suggestion
        paths within the root, each pointing to a targeted submodule. It must return the modified root.
```

src/pruna/config/target_modules.py
```diff
@@ -0,0 +1,213 @@
...
+    model: Any,
+    target_modules: TARGET_MODULES_TYPE,
+) -> Any:
+    """
+    Apply a function to the model, or to each of its targeted nn.Modules in the case of a Pipeline.
+
+    Parameters
+    ----------
+    apply_single_root_fn : Callable[[str | None, torch.nn.Module, List[str]], Any]
+        The function to apply each root in the model.
```
simlang: typo
```suggestion
        The function to apply to each root in the model.
```

PR #270, opened 2025-07-22
feat: add ImageReward metric for text-to-image evaluation

### LINKED ISSUE #268: [FEATURE] Add ImageReward metric

[NeurIPS 2023] ImageReward: Learning and Evaluating Human Preferences for Text-to-image Generation

Implement as Metric in Pruna.

davidberenstein1957:
## Description
* Introduced ImageRewardMetric to evaluate text-to-image generation quality, outperforming existing methods.
* Updated pyproject.toml to include new dependencies: image-reward and clip.
* Integrated ImageRewardMetric into the evaluation task processing.
* Added unit tests for ImageRewardMetric, covering registration, prompt extraction, scoring, and error handling.

## Related Issue

Fixes #268 

## Type of Change

## Checklist


davidberenstein1957:
@begumcig I found that all prompts are truncated to 35 tokens... not sure if that makes sense but otherwise we could overwrite the score function as most of the prompt mentioned in their dataset actually extend to longer than 35 tokens.

https://github.com/THUDM/ImageReward/blob/c1392c6dd0fd6ecd6d416c96959ab744a6d0a8fb/ImageReward/ImageReward.py#L110

davidberenstein1957:
@sharpenb @sdiazlor @begumcig I've implemented a more generalisable reward metric here: https://github.com/PrunaAI/pruna/pull/272

src/pruna/evaluation/metrics/metric_imagereward.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def __init__(
+        self,
+        device: str | torch.device | None = None,
+        model_name: str = "ImageReward-v1.0",
+        call_type: str = SINGLE,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        self.device = set_to_best_available_device(device)
```
begumcig, line 72: This will change depending on whether the multi-GPU PR is merged before this! Just a heads up :)

src/pruna/evaluation/metrics/metric_imagereward.py
```diff
@@ -0,0 +1,193 @@
...
+    model_name : str, optional
+        The ImageReward model to use. Default is "ImageReward-v1.0".
+    call_type : str
+        The type of call to use for the metric. IQA metrics, like image_reward, are only supported for single mode.
+    **kwargs : Any
+        Additional keyword arguments for the metric.
+    """
+
+    higher_is_better: bool = True
+    default_call_type: str = "y"
```
begumcig, line 56: Just to double-check, should the call_type be set to "y"? From what I understand (and please correct me if I’m off), the ImageReward model estimates human preferences for images based on how well they match a given prompt. That sounds more like a text-to-image alignment metric to me, rather than a purely image-based one. What do you think? So maybe the call type should be something like x_y?

src/pruna/evaluation/metrics/metric_imagereward.py
```diff
@@ -0,0 +1,193 @@
...
+            The input prompts for text-to-image generation.
+        gt : Tensor
+            The ground truth images (not used for ImageReward).
+        outputs : Tensor
+            The generated images to evaluate.
+        """
+        # Prepare inputs
+        metric_inputs = metric_data_processor(x, gt, outputs, self.call_type, device=self.device)
+        prompts = self._extract_prompts(x)
+        images = metric_inputs[1] if len(metric_inputs) > 1 else outputs
```
begumcig, line 104: Can you explain what's happening in this line? I am probably missing something here but ideally the metric data processor returns what we need to use in our metric so we don't have to touch x, gt or outputs again :)

src/pruna/evaluation/metrics/metric_imagereward.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Format images as PIL Images
+        formatted_images = [self._format_image(image) for image in images]
+
+        # Score images with prompts
+        for prompt, image in zip(prompts, formatted_images):
+            score = self.model.score(prompt, image)
```
begumcig, line 111: Supporting my comment above, the metric does seem to use the inputs and the outputs, we should consider changing the call_type.

src/pruna/evaluation/metrics/metric_imagereward.py
```diff
@@ -0,0 +1,193 @@
...
+        image : Tensor
+            The image to score.
+
+        Returns
+        -------
+        float
+            The ImageReward score for the image.
+        """
+        # Convert tensor to PIL Image
+        if image.dim() == 4:
```
begumcig, line 173: I completely understand what you are doing here and I completely agree that we should do these size checks. Since above we call this function with `_format_image(image) for image in images` do you think we should check for the case for when `dim()==2`? I think it's quite possible the user calls the metric with a single RGB image.

uv.lock
begumcig, line 1: I am sure there is a reason for this, I have never pushed a lock file before so just putting a comment incase it's a fluke

tests/evaluation/test_image_reward_metric.py
```diff
@@ -0,0 +1,82 @@
...
+    assert extracted == prompts
+
+    # Test with tensor (should generate default prompts)
+    tensor = torch.randn(2, 3, 224, 224)
+    extracted = metric._extract_prompts(tensor)
+    assert len(extracted) == 2
+    assert all(prompt.startswith("prompt_") for prompt in extracted)
+
+
+def test_score_image():
```
begumcig, line 45: These tests are really comprehensive thanks a lot! Shall we also add a case or cases using either `PrunaModel.run_inference() `or the` EvaluationAgent`, similar to our tests in` cmmd` metric? That way we would also ensure the metric is compatible with our engine and agent!

begumcig (changes requested):
Really great work David! Thank you so much for being so attentive to the details of the evaluation module. It's already almost there, requested some small changes :)

sharpenb (changes requested):
The PR description suggests some big overlap between https://github.com/PrunaAI/pruna/pull/272 and https://github.com/PrunaAI/pruna/pull/270. It would be nice to merge this two PR togethers (and address all @begum comments ;))

PR #272, opened 2025-07-22
feat: 271 feature implement hps hpsv2 VQA

### LINKED ISSUE #271: [FEATURE] implement hps hpsv2

implements
- HPS: https://github.com/tgxs002/align_sd
- HPSv2: https://github.com/tgxs002/HPSv2

NA

davidberenstein1957:
## Description

Implements hps hpsv, VQA and image reward as a single unified class with various adaptations per specific score metric.

## Related Issue

Fixes #271 
Fixes #270 

## Type of Change

## Checklist


pyproject.toml
```diff
@@ -60,6 +60,8 @@ gptqmodel = [
   { index = "pruna_internal", marker = "sys_platform != 'darwin' or platform_machine != 'arm64'"},
   { index = "pypi", marker = "sys_platform == 'darwin' and platform_machine == 'arm64'"},
 ]
+image-reward = { git = "https://github.com/PrunaAI/ImageReward" }
+hpsv2 = { git = "https://github.com/PrunaAI/HPSv2" }
```
davidberenstein1957, line 64: There were some issues with both repositories.
ImageReward, had pushed their fixes to `main` 
HPSv2 had very strict constraints for protobuf where they set it to lower than 6

davidberenstein1957: @johnrachwan123 perhaps we should publish these on our own index?

src/pruna/evaluation/metrics/__init__.py
```diff
@@ -37,4 +45,10 @@
     "TotalMACsMetric",
     "PairwiseClipScore",
     "CMMD",
+    "ImageRewardMetric",
+    "HPSMetric",
+    "HPSv2Metric",
+    "HPS_REWARD",
+    "HPSv2_REWARD",
+    "IMAGE_REWARD",
```
davidberenstein1957:
```suggestion
```

pyproject.toml
davidberenstein1957: also we could add t2v-metrics with a more relaxed Python constraint

src/pruna/evaluation/metrics/__init__.py
```diff
@@ -37,4 +39,8 @@
     "TotalMACsMetric",
     "PairwiseClipScore",
     "CMMD",
+    "ImageRewardMetric",
+    "HPSMetric",
```
sharpenb, line 43: If HPS does not work, I would prefer to leave it out for now, or fix it.

tests/evaluation/test_reward_metric.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+    request.cls._reward_metrics = metrics
+
+
+@pytest.mark.usefixtures("reward_metrics")
+@pytest.mark.parametrize(
+    "metric_cls, metric_name, model_load_kwargs",
+    METRIC_CLASSES_AND_NAMES,
+)
+class TestRewardMetrics:
```
sharpenb, line 55: We do not structure the tests in classes for now.

src/pruna/evaluation/task.py
```diff
@@ -199,6 +200,7 @@ def _process_single_request(request: str, device: str | torch.device | None) ->
             TorchMetricWrapper("clip_score"),
             TorchMetricWrapper("clip_score", call_type="pairwise"),
             CMMD(device=device),
+            ImageRewardMetric(device=device),
```
sharpenb, line 217: Why HPS, HPSv2, and VQA are not added here?

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        self.call_type = get_call_type_for_single_metric(call_type, self.default_call_type)
+        self.to_pil = ToPILImage()
+
+        # Initialize state for accumulating scores
+        self.add_state("scores", [])
+        self.add_state("prompts", [])
+        self._load(**model_load_kwargs)
+
+    def _load(self, **kwargs: Any) -> None:
+        raise NotImplementedError("Subclasses must implement this method")
```
sharpenb, line 69: Do we need these? We cna probably use abstract class and methods.

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        """
+        Compute the final ImageReward metric.
+
+        Returns
+        -------
+        MetricResult
+            The computed ImageReward score.
+        """
+        if not self.scores:
+            pruna_logger.warning("No scores available for ImageReward computation")
```
sharpenb, line 111: It mentiones ImageReward here but it is for the Base class.

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            return MetricResult(self.metric_name, self.__dict__.copy(), 0.0)
+
+        # Calculate mean score
+        mean_score = torch.mean(torch.tensor(self.scores)).item()
```
sharpenb, line 115: I discussed this with @begumcig. Not necessary for now but there is the quesiton of adding a aggregation function option which would be mean by default.

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+)
+from pruna.logging.logger import pruna_logger
+
+IMAGE_REWARD = "image_reward"
+HPS_REWARD = "hps"
+HPSv2_REWARD = "hpsv2"
+VQA_REWARD = "vqa"
+
+
+class BaseModelRewardMetric(StatefulMetric):
```
sharpenb, line 40: Either this class specialize already for text to image, or it should be made more general in its implementation (e.g. there is notion of prompt and images in some methods)

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        elif isinstance(x, Tensor):
+            # If x is a tensor, we need to handle it differently
+            # This might be the case for some data formats
+            pruna_logger.warning("Input x is a tensor, assuming it contains encoded prompts")
+            return [f"prompt_{i}" for i in range(x.shape[0])]
+        else:
+            pruna_logger.error(f"Unexpected input type for prompts: {type(x)}")
+            return []
+
+    def _format_image(self, image: Tensor) -> PIL.Image.Image:
```
sharpenb, line 144: I think that there should be a torch funciton doing this. https://docs.pytorch.org/vision/main/generated/torchvision.transforms.ToPILImage.html

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        Additional keyword arguments for the metric.
+    """
+
+    metric_name: str = IMAGE_REWARD
+
+    def _load(self, **kwargs: Any) -> None:
+        model_name = kwargs.get("model_name", "ImageReward-v1.0")
+        pruna_logger.info(f"Loading ImageReward model: {model_name}")
+
+        import ImageReward as RM  # noqa: N814
```
sharpenb, line 228: I am not sure that the import are managed like this for now in metrics. We should put them at the top. If we start to do this, we should redesign metrics to have imported modules a bit like the algorithms. Let's use the simple solution for now.

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            The score of the image.
+        """
+        model = self.model
+        preprocess = self.preprocess
+        device = self.device
+
+        # Preprocess image and move to device
+        images = preprocess(image).unsqueeze(0).to(device)
+        # Tokenize prompt and move to device
+        import clip
```
sharpenb, line 306: Same here, I also think that we already ahve some package capable of doing clip. Makes sense to check what is used in clip metric

src/pruna/evaluation/metrics/metric_reward.py
sharpenb: This also holds for other metrics below.

sharpenb (changes requested):
Looking forward to have these metrics in the package! I left some comment and @begum should also have a check :)

pyproject.toml
```diff
@@ -106,7 +108,6 @@ dependencies = [
     "codecarbon",
     "pynvml",
     "thop",
-    "timm",
     "bitsandbytes; sys_platform != 'darwin' or platform_machine != 'arm64'",
```
begumcig, line 111: Why did we remove timm from the dependencies?

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+
+
+class BaseModelRewardMetric(StatefulMetric):
+    """Base class for model reward metrics."""
+
+    higher_is_better: bool = True
+    default_call_type: str = "y"
+    metric_units: str = "score"
+
+    # Type annotations for dynamically added attributes
```
begumcig, line 47: slay

src/pruna/evaluation/metrics/metric_reward.py
begumcig: +1 for abstract classes if we are going to have a new metric type :)

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        x : List[str] | Tensor
+            The input prompts for text-to-image generation.
+        gt : Tensor
+            The ground truth images (not used for ImageReward).
+        outputs : Tensor
+            The generated images to evaluate.
+        """
+        # Prepare inputs
+        metric_inputs = metric_data_processor(x, gt, outputs, self.call_type, device=self.device)
+        prompts = self._extract_prompts(x)
```
begumcig, line 89: The job of metric_data_processor is to handle device casting and separation of the inputs required for the metric. I see that in the next line you are using x, which could lead to device casting problems since it's not coming from the metric_data_processor. If the metric also needs the inputs we should use a different `call_type` for the metric (like x_y or y_x), then extract x from what the metric_data_processor returns :)

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        images = metric_inputs[1] if len(metric_inputs) > 1 else outputs
```
begumcig, line 90: again we shouldn't use the outputs, once the metric data processor is called

src/pruna/evaluation/metrics/metric_reward.py
begumcig: +1

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            The input data containing prompts.
+
+        Returns
+        -------
+        List[str]
+            The extracted prompts.
+        """
+        if isinstance(x, list):
+            return x
+        elif isinstance(x, Tensor):
```
begumcig, line 135: I am assuming in the case where the type of the prompts is a list, it's a list of strings, am I correct? I just don't know the inner workings of the metric so I am not exactly sure, is the metric equipped to handle both cases of having a tensor with numerical values and also raw preprocessed strings? Do we get the same result?

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+        image : Tensor
+            The image to score.
+
+        Returns
+        -------
+        PIL.Image.Image
+            The formatted PIL image.
+        """
+        # Convert tensor to PIL Image
+        if image.dim() == 4:
```
begumcig, line 159: I am sorry if this is a stupid question but do we lose all the other images in the batch if we pass a batch of images?

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            # Batch dimension, take first image
+            image = image[0]
+
+        # Handle different tensor formats
+        if image.dim() == 3:
+            # Check if image is in HWC format (H, W, C)
+            if image.shape[-1] in [1, 3, 4]:
```
begumcig, line 166: I think this is the case for numpy arrays, if we type check in the function signature that we require torch.Tensors would we ever be in this case?

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+                # Convert HWC to CHW format
+                image = image.permute(2, 0, 1)
+            elif image.shape[0] in [1, 3, 4]:
+                # Image is already in CHW format
+                pass
+            else:
+                # Unknown format, try to handle gracefully
+                pruna_logger.warning(f"Unexpected image shape: {image.shape}")
+                return PIL.Image.new("RGB", (224, 224))
```
begumcig, line 175: I think instead of creating a fake image we should just raise an error

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            elif image.shape[0] == 4:
+                # RGBA, take only RGB channels
+                image = image[:3]
+
+            # Normalize to [0, 1] if needed
+            if image.max() > 1.0:
+                image = image / 255.0
+
+            # Ensure values are in valid range
+            image = torch.clamp(image, 0.0, 1.0)
```
begumcig, line 190: if the image is normalized to have [-1, 1] range, wouldn't clamping here make us lose some information?

src/pruna/evaluation/metrics/metric_reward.py
begumcig: yup!

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+    **model_load_kwargs : Any
+        Additional keyword arguments for the model.
+    **kwargs : Any
+        Additional keyword arguments for the metric.
+    """
+
+    metric_name: str = HPSv2_REWARD
+
+    def _load(self, **kwargs: Any) -> None:
+        from functools import partial
```
begumcig, line 346: i think we can import partial on the top of the file :)

src/pruna/evaluation/metrics/metric_reward.py
```diff
@@ -0,0 +1,444 @@
...
+            The prompt to score the image with.
+        image : PIL.Image.Image
+            The image to score.
+
+        Returns
+        -------
+        float
+            The score of the image.
+        """
+        score = self.model(imgs_path=image, prompt=prompt)
```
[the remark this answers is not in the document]
begumcig: I remember some of the rewards metrics expecting a path rather than a image, could be beneficial to look into this warning

tests/evaluation/test_reward_metric.py
```diff
@@ -0,0 +1,138 @@
...
+        # Test with list of strings
+        prompts = ["a beautiful sunset", "a cat playing"]
+        extracted = metric._extract_prompts(prompts)
+        assert extracted == prompts
+
+        # Test with tensor (should generate default prompts)
+        tensor = torch.randn(2, 3, 224, 224)
+        extracted = metric._extract_prompts(tensor)
+        assert len(extracted) == 2
+        assert all(prompt.startswith("prompt_") for prompt in extracted)
```
begumcig, line 85: again i am a bit confused about this use case 🥺. do we just return prompt: and then the tensor in string format?

tests/evaluation/test_reward_metric.py
```diff
@@ -0,0 +1,138 @@
...
+        # Convert to PIL Image properly
+        image = image.clamp(0, 1)  # Ensure values are in [0, 1]
+        pil_image = Image.fromarray(
+            (image.permute(1, 2, 0).cpu().numpy() * 255).astype("uint8")
+        )
+
+        score = metric._score_image(prompt, pil_image)
+        assert isinstance(score, float)
+        # Score should be a reasonable value (ImageReward/HPS typically outputs scores around 0-10)
+        assert -10 <= score <= 10
```
begumcig, line 104: can the score actually be negative ?

begumcig (changes requested):
Thank you so much for such a thought-out implementation David, I really like the RewardsMetric structure. I left some comments mostly about the base class but the metrics themselves already look pretty good overall!

PR #283, opened 2025-07-29
docs: create end to end reasoning tutorial

sdiazlor:
## Description

This PR add the tutorial for reasoning LLMs.
- It uses Qwen3-1.7B reasoning model.
- It uses a `hqq` and `torch_compile`
- It evaluates using the reasoning dataset `zwhe99/DeepMath-103K`
- It increases the performance evaluated with 4 metrics.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


davidberenstein1957 (commented):
Hi @sdiazlor looking good!
some comments
- Qwen/Qwen3-0.6B is used but 1.7B is mentioned
- COuld you look at the evaluation. I think we are currently evaluating using a dataset that does not contain reasoning, so I assume this is leading to some issues when it comes quality metrics. Also, this might be a good excuse to showcase how to use a custom config for the datamodule.
- Could you mention the device?
- The current way we create the tables is a bit off. Could you take a look at a small PR where we add the metric units etc to the MetricResult, this might help with the bloatedness of the code :) 
-

davidberenstein1957 (approved):

PR #287, opened 2025-08-06
feat: add `flash_attention 3` kernel for diffusers pipelines

johannaSommer:
## Description
In this PR, we make use of HuggingFace's kernel hub to utilize flash attention 3 in diffusers pipelines where possible. Importantly, diffusers is refactoring their attention handling right now, hence we have two different implementations. 

In the current diffusers version, we wrap the call of a pipeline to intercept each call to `torch.nn.funcional.scaled_dot_product_attention`. If shapes, dtype and keyword arguments can be supported by flash attention, we reroute the computation to the fa3 kernel.

In the newer diffusers version (0.35.0, unreleased so far), we can register a new attention backend that calls the native attention or fa3 on the same criteria.

Overall, the speedup is minor for T2I pipelines but is substantial for Video Gen pipelines, in particular Wan.

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
Added algorithm tests with Flux and Wan and added combination tests.

## Checklist

## Additional Notes
Flash Attention 3 on WAN gives around a 1.4 speedup, no warmup or quality degradation of course.

Minimal Script:
``` python
from diffusers import WanPipeline
import torch

pipeline = WanPipeline.from_pretrained(
    "Wan-AI/Wan2.2-T2V-A14B-Diffusers", torch_dtype=torch.bfloat16
)
pipeline.to("cuda");

config = SmashConfig()
config._prepare_saving = False
config["kernel"] = "flash_attn3"
pipeline = smash(pipeline, config)

prompt = "A cat is doing an acrobatic dive into a swimming pool at the olympics, from a 10m high diving board, flips and spins"
negative_prompt = "Bright tones, overexposed, static, blurred details, subtitles, style, works, paintings, images, static, overall gray, worst quality, low quality, JPEG compression residue, ugly, incomplete, extra fingers, poorly drawn hands, poorly drawn faces, deformed, disfigured, misshapen limbs, fused fingers, still picture, messy background, three legs, many people in the background, walking backwards"
output = pipeline.__call__(
    prompt=prompt,
    negative_prompt=negative_prompt,
    height=480,
    width=640,
    num_frames=69,
    guidance_scale=5.0,
    num_inference_steps=10,
    generator=torch.Generator(device="cpu").manual_seed(1),
).frames[0]
```
Output during Generation:
`100%|██████████| 10/10 [01:04<00:00,  6.45s/it]`

----
Generation times on the prompt / size above on 1 H100:
- base: 8.04 s/it
- `flash_attn3`: 6.46 s/it (no warmup!)
- `torch.compile`: 6.38 s/it
- `flash_attn3` + `torch.compile`: 5.15 s/it

### DIFF
--- a/src/pruna/algorithms/kernels/flash_attn3.py
+++ b/src/pruna/algorithms/kernels/flash_attn3.py
@@ -46,7 +46,9 @@ class FlashAttn3Kernel(PrunaKernel):
     processor_required: bool = False
     runs_on: list[str] = ["cuda", "accelerate"]
     dataset_required: bool = False
-    compatible_algorithms: dict[str, list[str]] = dict(compiler=["torch_compile"], cacher=["fora"])
+    compatible_algorithms: dict[str, list[str]] = dict(
+        compiler=["torch_compile"], cacher=["fora"], quantizer=["torchao"]
+    )
 
     def model_check_fn(self, model: Any) -> bool:
         """
@@ -91,6 +93,9 @@ class FlashAttn3Kernel(PrunaKernel):
         """
         imported_packages = self.import_algorithm_packages()
 
+        # register the flash attention 3 operation with torch ops to make it compatible with full-graph compilation
+        register_pruna_flash_attn_op(imported_packages["flash_attention_3"])
+
         # in the new version of diffusers, we can use the modular attention backend to inject flash_attn3
         if Version(diffusers_version) >= Version("0.35.0.dev0"):
             # register our "custom" attention function as a backend
@@ -105,9 +110,6 @@ class FlashAttn3Kernel(PrunaKernel):
                     component.set_attention_backend("flash_attn3_pruna")
 
         else:
-            # register the flash attention 3 operation with torch ops to make it compatible with fullgraph compilation
-            register_pruna_flash_attn_op(imported_packages["flash_attention_3"])
-
             # wrap the model generate function to replace attention computations with flash_attn3 where possible
             wrap_pipeline_call(model, imported_packages)
         return model
@@ -152,6 +154,8 @@ def register_custom_backend(imported_packages: Dict[str, Any]) -> None:
     """
     Register the attention backend for flash_attn3 by mimicing the native backend.
 
+    Applies to diffusers >= 0.35.0.dev0.
+
     Parameters
     ----------
     imported_packages : Dict[str, Any]
@@ -162,7 +166,6 @@ def register_custom_backend(imported_packages: Dict[str, Any]) -> None:
     _check_shape = imported_packages["_check_shape"]
     _check_qkv_dtype_bf16_or_fp16 = imported_packages["_check_qkv_dtype_bf16_or_fp16"]
     _native_attention = imported_packages["_native_attention"]
-    flash_attention_3 = imported_packages["flash_attention_3"]
     attention_backend_name = imported_packages["AttentionBackendName"]
 
     if attention_backend_registry.get_active_backend()[0].name != "NATIVE":
@@ -203,7 +206,7 @@ def register_custom_backend(imported_packages: Dict[str, Any]) -> None:
             # if any constraints are not met or unsupported input arguments are being used, reroute to native attention
             if attn_mask is not None or dropout_p != 0.0 or not dtype_pass or not num_heads_pass or not head_dim_pass:
                 pruna_logger.debug(
-                    "Rerouting to native attention... Check the following criteria: "
+                    "Rerouting to native attention. Check the following criteria in algorithms/kernels/flash_attn3.py: "
                     f"attn_mask_pass: {attn_mask is not None}, dropout_p_pass: {dropout_p != 0.0}, "
                     f"dtype_pass: {dtype_pass}, num_heads_pass: {num_heads_pass}, head_dim_pass: {head_dim_pass}"
                 )
@@ -220,13 +223,8 @@ def register_custom_backend(imported_packages: Dict[str, Any]) -> None:
                 )
             else:
                 pruna_logger.debug("Using FA3...")
-                out, _, *_ = flash_attention_3.flash_attn_func(
-                    q=query,
-                    k=key,
-                    v=value,
-                    softmax_scale=scale,
-                    causal=is_causal,
-                    deterministic=False,
+                out, _, *_ = torch.ops.flash_attn_pruna._flash_attn_forward(
+                    q=query, k=key, v=value, softmax_scale=scale, causal=is_causal
                 )
                 return out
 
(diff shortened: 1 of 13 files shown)


johnrachwan123:
bugbot run

src/pruna/algorithms/kernels/flash_attn3.py
```diff
@@ -0,0 +1,354 @@
...
+        ----------
+        model : Any
+            The model to check.
+
+        Returns
+        -------
+        bool
+            True if the model is a valid model for the algorithm, False otherwise.
+        """
+        if Version(diffusers_version) >= Version("0.35.0.dev0"):
```
sharpenb, line 67: I am not sure about these conditions. It might be able to work with transformers.

src/pruna/algorithms/kernels/flash_attn3.py
```diff
@@ -0,0 +1,354 @@
...
+
+            # fa3 only supports attention with num_query_heads % num_kv_heads == 0
+            num_heads_pass = all(query.shape[1] % t.shape[1] == 0 for t in (key, value))
+
+            # test head dimension
+            head_dim_pass = all(t.shape[3] <= 256 for t in (query, key, value))
+
+            # if any constraints are not met or unsupported input arguments are being used, reroute to native attention
+            if attn_mask is not None or dropout_p != 0.0 or not dtype_pass or not num_heads_pass or not head_dim_pass:
+                pruna_logger.debug(
```
sharpenb, line 208: This debugging message does not really say what are the constraints which are explained in the comments above.

src/pruna/algorithms/kernels/flash_attn3.py
```diff
@@ -0,0 +1,354 @@
...
+
+        Returns
+        -------
+        Any
+            The wrapped model.
+        """
+        imported_packages = self.import_algorithm_packages()
+
+        # in the new version of diffusers, we can use the modular attention backend to inject flash_attn3
+        if Version(diffusers_version) >= Version("0.35.0.dev0"):
```
sharpenb, line 100: Do we need to support both versions for certain model use-cases? It feels to me that the code could be more readable by saying we only support from version xyz.

src/pruna/algorithms/kernels/flash_attn3.py
```diff
@@ -0,0 +1,354 @@
...
+
+            # replace in all compatible components
+            for component in model.components.values():
+                if hasattr(component, "set_attention_backend") and component.dtype in [
+                    torch.bfloat16,
+                    torch.float16,
+                ]:
+                    component.set_attention_backend("flash_attn3_pruna")
+
+        else:
```
sharpenb, line 112: I was confused by some functions redundant but I guess they are designed for the two diffusers versions. Could we mark e.g. in their docstrings, or separate files.

src/pruna/algorithms/kernels/flash_attn3.py
```diff
@@ -0,0 +1,354 @@
...
+    def import_algorithm_packages(self) -> Dict[str, Any]:
+        """
+        Import the algorithm packages.
+
+        Returns
+        -------
+        Dict[str, Any]
+            The algorithm packages.
+        """
+        flash_attention_3 = get_kernel("kernels-community/flash-attn3")
```
sharpenb, line 126: This is designed only for fa3 but `kernels_hub` has many more kernels. I feel that extending it to all kernels directly would be nice. Probably not for this PR but worth spending a bit of time now if the logic is similar.

sharpenb (changes requested):
Thanks for the PR :)  
- I would be interested in a small benchmark (base, torch compile, fa3, fa3+torch.compile) for Wan (if that is the main model targeted by this update) 
- Was it tested with EvaluationAgent after compression?
- Could we have test notebook since it integrates a new compression algorithm?

src/pruna/algorithms/kernels/flash_attn3.py
johannaSommer: I don't understand this comment, can you elaborate? Because of refactoring in the upcoming diffusers version we have to apply seperate logic.

src/pruna/algorithms/kernels/flash_attn3.py
johannaSommer: I don't this this makes sense here - at least currently the other kernels available on kernel hub tackle other functionality, i.e. not the attention function. I think it is better to think of this `get_kernel` like an aggregation of building, registering and importing a certain kernel package, but which functions you use and how you use them will be different from kernel to kernel. If there are other interesting attention kernels coming soon, we will of course consolidate them.

src/pruna/algorithms/kernels/flash_attn3.py
johannaSommer: I know what you mean - I added a slight extension to point to the file where the user could review these criteria. Note that this is really only for debugging purposes, activating this debug message will print _a lot_ since the text is displayed at **every** `torch.nn.functional.scaled_dot_product_attention` call. I added this mainly so that we can support someone who tries this on a pipeline we haven't tested before and doesn't experience any speedup.

src/pruna/algorithms/kernels/flash_attn3.py
sharpenb: Sorry for the confusion. I mean that model checks focus on diffusers, while it might work for transformers architectures. My understanding now is that it was not tested for transformers but could be worth to have a quick check.

src/pruna/algorithms/kernels/flash_attn3.py
sharpenb: I think that it aligns with my point. Here, would it be possible to have a streamlined way to integrate other kernels e.g. by defining a mapping from kernel to the functions we use. I would expect that if we implement another kernel from the hub, we would go over similar implementation steps e.g. registering, identifying functions, and wrapping calls. One potential ideais to take inspiration from their blog which suggests to use `use_kernel_forward_from_hub`. Again no need to refactorize , but wanted to highlight this since it might be relevant for the future ;)

src/pruna/algorithms/kernels/flash_attn3.py
sharpenb: Super nice thanks!

src/pruna/algorithms/kernels/flash_attn3.py
sharpenb: Let's have a reminder to remove support for older version of diffusers when a couple fo diffusers releases will have been done :)

sharpenb (approved):
Let's go! I left some final thoughts

PR #291, opened 2025-08-07
Feat: include python 3.13

GreatBahram:
## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist


GreatBahram:
After solving the build issue; I wonder if we can use the `pyupgrade` to benefit from the latest type hints features [[link](https://adamj.eu/tech/2021/05/21/python-type-hints-how-to-upgrade-syntax-with-pyupgrade/)]

.github/workflows/build.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
   build:
     strategy:
       matrix:
-        python-version: ["3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
GreatBahram, line 21: > [!NOTE]
> Python 3.9 reaches end-of-life (EOL) soon (October 2025) :partying_face:

GreatBahram:
It seems the [whisper-s2t](https://pypi.org/project/whisper-s2t/) is not that much maintained, and it prevents us from supporting Python 3.13 (probably if it makes sense better to) replace it with [faster-whisper](https://github.com/SYSTRAN/faster-whisper). What do you think, @johannaSommer?

GreatBahram:
> Hey @GreatBahram thanks a lot for taking care of this! I think this upgrade is great, just a few changes we might want to do:
> 
>     * to avoid a lot of overhead for the GitHub actions, let's keep the python version we use for e.g. build, docs, etc. to just one for now? I do agree though that we can update this, what about python 3.11? 😊
> 
>     * I also agree that we should support python 3.13, but as you said we have to find a solution for whisper-s2t first. Waiting for @johnrachwan123 's feedback here
>       Otherwise, it looks perfect! 🌻🌻🌻

Hey @johannaSommer totally makes sense, then I'm gonna reduce the scope and push the changes and I'll ask you to have another look :sunflower:

PR #308, opened 2025-08-19
torch.load always to cpu first

simlang:
## Description
This PR does a little fix
When using torch.load, always load to cpu first and then move to target device.
Problem when loading to the device directly was, that every tensor within a pipeline e.g. was moved to cuda, also some which are not supposed to be - not every pipeline can handle this unexpected device (e.g. SanaPipeline)
By loading to cpu first and then moving to target device, we make sure that only the right parts are moved to the target_device

## Type of Change

## How Has This Been Tested?
Every test which is using torch.load has been re-run and is successful

## Checklist


simlang:
there are already tests with Sana, which are currently failing - this PR fixes these tests

johannaSommer (approved):
🫡 merci

PR #310, opened 2025-08-21
feat: add prompt only image generation datasets

nifleisch:
## Description
Currently, we only support image-generation datasets consisting of prompt–image pairs. For many applications (e.g., evaluation agents, distillation), only the prompt is needed, and some benchmarking datasets for image-generation models consist of prompts only. This PR relaxes the requirement to always include images by adjusting the collate function. It also adds three common benchmarking datasets for image-generation models: DrawBench, PartiPrompts, and GenAiBench.

## Type of Change

## How Has This Been Tested?
- I added tests for the newly added datasets that pass locally.
- I tried the old and new image generation datasets together with the optimization agent. 

## Checklist

### DIFF
--- a/src/pruna/data/__init__.py
+++ b/src/pruna/data/__init__.py
@@ -35,8 +35,11 @@ from pruna.data.datasets.text_generation import (
 )
 from pruna.data.datasets.text_to_image import (
     setup_coco_dataset,
+    setup_drawbench_dataset,
+    setup_genai_bench_dataset,
     setup_laion256_dataset,
     setup_open_image_dataset,
+    setup_parti_prompts_dataset,
 )
 
 base_datasets: dict[str, Tuple[Callable, str, dict[str, Any]]] = {
@@ -56,4 +59,7 @@ base_datasets: dict[str, Tuple[Callable, str, dict[str, Any]]] = {
     "Polyglot": (setup_polyglot_dataset, "question_answering_collate", {}),
     "OpenImage": (setup_open_image_dataset, "image_generation_collate", {"img_size": 1024}),
     "CIFAR10": (setup_cifar10_dataset, "image_classification_collate", {"img_size": 32}),
+    "DrawBench": (setup_drawbench_dataset, "image_generation_collate", {"img_size": None}),
+    "PartiPrompts": (setup_parti_prompts_dataset, "image_generation_collate", {"img_size": None}),
+    "GenAIBench": (setup_genai_bench_dataset, "image_generation_collate", {"img_size": None}),
 }
--- a/src/pruna/data/collate.py
+++ b/src/pruna/data/collate.py
@@ -54,7 +54,9 @@ def image_format_to_transforms(output_format: str, img_size: int) -> transforms.
         raise ValueError(f"Invalid output format: {output_format}")
 
 
-def image_generation_collate(data: Any, img_size: int, output_format: str = "int") -> Tuple[List[str], torch.Tensor]:
+def image_generation_collate(
+    data: Any, img_size: int | None, output_format: str = "int"
+) -> Tuple[List[str], torch.Tensor]:
     """
     Custom collation function for text-to-image generation datasets.
 
@@ -65,8 +67,8 @@ def image_generation_collate(data: Any, img_size: int, output_format: str = "int
     ----------
     data : Any
         The data to collate.
-    img_size : int
-        The size of the image to resize to.
+    img_size : int | None
+        The size of the image to resize to. None if dataset does not contain images.
     output_format : str
         The output format, in ["int", "float", "normalized"].
         With "int", output tensors have integer values between 0 and 255. With "float", they have float values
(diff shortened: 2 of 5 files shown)


src/pruna/data/datasets/text_to_image.py
```diff
@@ -124,3 +124,66 @@ def _process_example(example: Dict[str, Any], directory_dataset: str) -> Dict[st
...
+        The seed to use.
+
+    Returns
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The DrawBench dataset.
+    """
+    ds = load_dataset("sayakpaul/drawbench", trust_remote_code=True)["train"]
+    ds = ds.rename_column("Prompts", "text")
+    return ds.select([0]), ds.select([0]), ds
```
davidberenstein1957: Perhaps this is leftover from testing? I believe we would want split the dataset in train test and val datasets?

src/pruna/data/datasets/text_to_image.py
```diff
@@ -124,3 +124,66 @@ def _process_example(example: Dict[str, Any], directory_dataset: str) -> Dict[st
...
+        The seed to use.
+
+    Returns
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The GenAI Bench dataset.
+    """
+    ds = load_dataset("BaiqiL/GenAI-Bench")["train"]
+    ds = ds.rename_column("Prompt", "text")
+    return ds.select([0]), ds.select([0]), ds
```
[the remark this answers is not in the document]
nifleisch: This decision was made deliberately to ensure that the evaluation agent uses all prompts in the benchmark for evaluation. Because these are benchmarking datasets, they are not intended for training models.

src/pruna/data/datasets/text_to_image.py
nifleisch: This decision was made deliberately to ensure that the evaluation agent uses all prompts in the benchmark for evaluation. Because these are benchmarking datasets, they are not intended for training models. But maybe I am missing cases in which it is favorable to split the benchmark dataset into train, val, and test.

davidberenstein1957: Makes sense, for consistency, I would add a note why you are doing that. Also, wouldn't it be better to provide an empty dataset ?

src/pruna/data/collate.py
```diff
@@ -54,7 +54,9 @@ def image_format_to_transforms(output_format: str, img_size: int) -> transforms.
         raise ValueError(f"Invalid output format: {output_format}")
 
 
-def image_generation_collate(data: Any, img_size: int, output_format: str = "int") -> Tuple[List[str], torch.Tensor]:
+def image_generation_collate(
```
begumcig: I totally get why it feels natural to link prompt datasets with the image_generation_collate. At the same time, I wonder if it might make sense to introduce a really simple, pass-through style collate function specifically for prompt-only datasets. I can imagine us adding more and more datasets in the future,  also not just for images but also for video generation too! So extending the current collate might not be the most sustainable approach long term.

src/pruna/data/datasets/text_to_image.py
begumcig: I second the idea of empty datasets for training and validation here :)

src/pruna/data/__init__.py
```diff
@@ -56,4 +59,7 @@
     "Polyglot": (setup_polyglot_dataset, "question_answering_collate", {}),
     "OpenImage": (setup_open_image_dataset, "image_generation_collate", {"img_size": 1024}),
     "CIFAR10": (setup_cifar10_dataset, "image_classification_collate", {"img_size": 32}),
+    "DrawBench": (setup_drawbench_dataset, "image_generation_collate", {"img_size": None}),
+    "PartiPrompts": (setup_parti_prompts_dataset, "image_generation_collate", {"img_size": None}),
+    "GenAIBench": (setup_genai_bench_dataset, "image_generation_collate", {"img_size": None}),
```
begumcig: I left a more detailed comment below, but if we were to have a separate collate, we also wouldn't have to set the image size for prompt datasets!

begumcig (changes requested):
THEY ARE HERE! You're a champ 🥇🥇🥇🥇. One teeeny request about passing empty datasets for train/val modes. Also left one suggestion about the collate, but only if you also think that makes sense, left it up to you. Everything already looks so good!

src/pruna/data/collate.py
nifleisch: Really good point. I did not think about video models. With them in mind it really makes sense to treat the prompt datasets separately.

src/pruna/data/__init__.py
```diff
@@ -56,4 +61,7 @@
     "Polyglot": (setup_polyglot_dataset, "question_answering_collate", {}),
     "OpenImage": (setup_open_image_dataset, "image_generation_collate", {"img_size": 1024}),
     "CIFAR10": (setup_cifar10_dataset, "image_classification_collate", {"img_size": 32}),
+    "DrawBench": (setup_drawbench_dataset, "prompt_collate", {}),
```
begumcig, line 64: slay

src/pruna/data/datasets/prompt.py
```diff
@@ -0,0 +1,80 @@
...
+        The seed to use.
+
+    Returns
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The GenAI Bench dataset.
+    """
+    ds = load_dataset("BaiqiL/GenAI-Bench")["train"]
+    ds = ds.rename_column("Prompt", "text")
+    return ds.select([0]), ds.select([0]), ds
```
begumcig, line 85: After our discussion I see why we cannot pass an empty dataset for the train and validation. Do you think it would make sense to print an info / warning in the setup functions to let people know they should be using the test dataloader?

begumcig (approved):
Amazing job Nils! Super excited to use these datasets already 🏆🏆🏆. It could be useful to add an info message to the user about how they should use the test split, but everything looks super good to me, so already approved!

src/pruna/data/datasets/prompt.py
nifleisch: Yes, that sounds like a good idea. Will log an info message when loading the dataset.

PR #315, opened 2025-08-22
fix: nightly tests llmcompressor and gptq

llcnt:
## Description
(i) `llmcompressor `requires LLM with weights large enough for the awq kernels. So I created a new "tiny_llama", and use it for the tests. Also the asymetric quantization scheme was running well for inference, but it yields error when we save and re-load. So I have added it as an hyperparameter.
(ii) For `GPTQ `I encountered several limitations:
- the model was never tested correctly when using compilation (the compilation was not working on the `generate `function as we expect for LLMs, but on the whole model -> this gives error at loading time and prevent from having any speedups). I have added a `is_gptq_model `function (because after gptq_model quantization, the model is NOT anymore `CausalLM `in the sense of `transformers`, but a `GPTQmodel`!) that we use in `torch_compile `to patch this. Also gptq is not compatible with `torch_compile `if one sets `full_graph=True`. So we must keep `full_graph=False` in our tests.
- gptq also suffers from the same limitations as awq for small weights, so I also use "tiny_llama".
- gptq save/load is partially bug. The re-load forgets to place some parameters (like the permutation indexes) on the correct device. So I added a patch do handle this in the `move_to_device `function.
- pipelines are not well integrated in the evaluation agent (except if I missed something; tbh I am not very familiar with it). The input a pipeline need is not the same as a raw CausalLM does (tokenizer should be handled). 

## Related Issue
This PR fixes 3 failing GPU tests:
- tests/algorithms/test_algorithms.py::test_full_integration[TestLLMCompressor_noref_llama_3_tiny_random-cuda]
- tests/algorithms/test_combinations.py::test_full_integration_combo[llama_3_tiny_random-config_dict5-True-perplexity]
- tests/algorithms/test_combinations.py::test_full_integration_combo[llama_3_tiny_random_as_pipeline-config_dict6-True-perplexity]

## Type of Change

## How Has This Been Tested?
image

## Checklist


src/pruna/engine/handler/handler_utils.py
```diff
@@ -49,7 +50,14 @@ def register_inference_handler(model: Any) -> InferenceHandler:
 
     model_module = model._orig_mod.__module__ if hasattr(model, "_orig_mod") else model.__module__
 
-    if "diffusers" in model_module:
+    # Check if it's a transformers pipeline
+    if hasattr(model, "__class__") and "Pipeline" in model.__class__.__name__:
```
begumcig: I might be wrong here but doesn't every python object have a __class__ attribute?

src/pruna/engine/handler/handler_utils.py
```diff
@@ -49,7 +50,14 @@ def register_inference_handler(model: Any) -> InferenceHandler:
...
+        # Specific check for text generation pipelines
+        if "TextGeneration" in model.__class__.__name__:
+            return PipelineHandler(pipeline=model)
+        # For other pipelines, fallback to standard pipeline handler
+        return PipelineHandler(pipeline=model)
```
begumcig: Would this also catch models from diffusers as they are mostly pipelines?

tests/algorithms/test_combinations.py
```diff
@@ -48,8 +48,8 @@ def prepare_smash_config(self, smash_config: SmashConfig, device: str) -> None:
         ("sd_tiny_random", dict(quantizer="hqq_diffusers", compiler="torch_compile"), False, 'cmmd'),
         ("flux_tiny_random", dict(quantizer="hqq_diffusers", compiler="torch_compile"), False, 'cmmd'),
         ("sd_tiny_random", dict(quantizer="diffusers_int8", compiler="torch_compile"), False, 'cmmd'),
-        ("llama_3_tiny_random", dict(quantizer="gptq", compiler="torch_compile"), True, 'perplexity'),
-        ("llama_3_tiny_random_as_pipeline", dict(quantizer="gptq", compiler="torch_compile", torch_compile_fullgraph=False), True, 'perplexity'),
```
begumcig, line 52: If I remember correctly we discussed this during the eval/inference tests connections, and decided to set the fulgraph hyperparameter default to False. It should be the case that the default fullgraph setting in Pruna is false. Can we check why the default is not working?

src/pruna/engine/utils.py
```diff
@@ -185,6 +192,78 @@ def move_to_device(
     safe_memory_cleanup()
 
 
+def _ensure_gptq_device_consistency(model: Any, target_device: str) -> None:
```
begumcig, line 195: Asking just to understand a bit better, we need this because model.to() doesn't move certain layers in a gptq quantized models right? Would any other quantization algorithm have qweight or qlinear attributes?

src/pruna/engine/utils.py
```diff
@@ -185,6 +192,78 @@ def move_to_device(
...
+                        if isinstance(attr, torch.Tensor) and attr.device != torch.device(target_device):
+                            try:
+                                setattr(module, attr_name, attr.to(target_device))
+                            except Exception as e:
+                                # Some tensors might be read-only or have special handling
+                                pruna_logger.debug(f"Could not move {attr_name} in {name} to {target_device}: {e}")
+
+                # Comprehensive scan: move ALL tensor attributes that might be missed
+                # This catches TorchQuantLinear and other GPTQ variant tensors
+                for attr_name in dir(module):
```
begumcig, line 229: if we do this comprehensive check here, do we still need to do the check above?

src/pruna/engine/utils.py
```diff
@@ -185,6 +192,78 @@ def move_to_device(
...
+                    if buffer.device != torch.device(target_device):
+                        try:
+                            buffer.data = buffer.data.to(target_device)
+                        except Exception as e:
+                            pruna_logger.warning(
+                                f"Could not move buffer {buffer_name} in {name} to {target_device}: {e}"
+                            )
+
+                # Check parameters as well
+                for param_name, param in module.named_parameters(recurse=False):
```
begumcig, line 254: slay

src/pruna/engine/handler/handler_pipeline.py
```diff
@@ -0,0 +1,205 @@
...
+
+        Returns
+        -------
+        Any
+            The prepared inputs (tensors for evaluation, strings for normal generation).
+        """
+        x, _ = batch
+
+        # If x is already a string or list of strings, return as is
+        if isinstance(x, (str, list)) and all(isinstance(item, str) for item in (x if isinstance(x, list) else [x])):
```
begumcig: I am probably not seeing something here, could you please explain why do we do these checks if we are returning x 🥺🥺

src/pruna/engine/handler/handler_pipeline.py
```diff
@@ -0,0 +1,205 @@
...
+            # Check if we're being called with string inputs (normal generation) or tensor inputs (evaluation context)
+            inputs = args[0] if len(args) > 0 else kwargs.get("inputs", kwargs.get("text_inputs"))
+
+            # If input is tensor or we detect evaluation context, return logits
+            if hasattr(inputs, "shape") and hasattr(inputs, "dtype"):
+                # This is a tensor input - likely from evaluation pipeline
+                # Use the model's forward pass to get logits instead of text generation
+                try:
+                    # Prepare inputs for the model
+                    if hasattr(pipeline, "model") and hasattr(inputs, "to"):
```
begumcig: `PrunaModel.run_inference() ` already calls the` move_inputs_to_device ` function :)

src/pruna/engine/handler/handler_pipeline.py
```diff
@@ -0,0 +1,205 @@
...
+                        device = next(pipeline.model.parameters()).device
+                        inputs = inputs.to(device)
+
+                        # Call the model's forward method directly to get logits
+                        with torch.no_grad():
+                            outputs = pipeline.model(input_ids=inputs)
+
+                        # Return logits in the format expected by perplexity metric
+                        if hasattr(outputs, "logits"):
```
begumcig: Is there a reason why we handle the outputs here and not in `process_output()` ?

src/pruna/engine/handler/handler_pipeline.py
```diff
@@ -0,0 +1,205 @@
...
+                if "top_k" not in kwargs and hasattr(generation_config, "top_k"):
+                    kwargs["top_k"] = generation_config.top_k
+
+            # Call the original generate method with the extracted parameters
+            return original_generate(*args, **kwargs)
+
+        # Replace the generate method with our patched version
+        model.generate = patched_generate
+
+    def _patch_pipeline_call(self, pipeline: Any) -> None:
```
begumcig, line 50: I totally understand why this function exists, but it seems to overlap with `pruna_model.run_inference`. Typically, inference handlers are responsible for preparing inputs, handling outputs, and managing arguments passed to the model, while the actual model execution is delegated to PrunaModel. Introducing model invocation logic here could lead to redundancy and blur the separation of responsibilities between the handler layer and the model abstraction. My preference is that we should not  be patching call method if we can avoid it 🥹

src/pruna/algorithms/compilation/torch_compile.py
```diff
@@ -414,7 +420,7 @@ def causal_lm_or_janus_logic(model: Any, smash_config: SmashConfigPrefixWrapper)
         # https://huggingface.co/docs/transformers/en/main_classes/text_generation#transformers.GenerationConfig.temperature
         temperature = 1.0
 
-    if is_causal_lm(model):
+    if is_causal_lm(model) or is_gptq_model(model):
```
begumcig, line 423: 🥇🥇🥇

src/pruna/engine/handler/handler_pipeline.py
```diff
@@ -0,0 +1,205 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
```
begumcig, line 1: You are a champ for working on inference Louis!! 🙌
I want to better understand the handler_pipeline, since I might have a suggestion once I do. From what I can tell, we’re mainly using it to handle the generation_config for Transformer pipelines.

A couple of questions to clarify (incoming 32994929 questions) (i am sorry)

Generation config + smashed models
– Does passing a generation_config break when we use the smashed models?
– Or do we also need to extract the kwargs from the generation_config for base models?

Division of responsibility
– Since the Transformer handler already takes care of inputs and outputs, is there a reason we can’t also handle the generation_config logic directly inside TransformerHandler?
– I’m concerned that keeping this logic in the generic pipeline handler might cause DiffusionPipeline instances to get routed there incorrectly.

llcnt: For the generationconfig, you are right, I removed it from this file. But for the generate function when we apply compilation we are explicitely expecting a max_tokens args. So I slightly adapted the logic there to handle the case where the user (or eg here the handler) will not provide such arg, but a generation_config.
For the diffusers, yes it is a good point, I made some changes so that a diffusers Pipeline doest not get intercept by the transformers Pipeline :)

src/pruna/engine/handler/handler_pipeline.py
llcnt: I think I see your point. However for evaluation ppl we need acess to the logits. And for a transformers pipeline we can not do pipe.generate and output.logits. We have to call instead pipeline.model, and then get the logits after.
Do you have any suggestion how to do this without patching the call? I guess we want to keep the pipeline.generate call in the run_inference function, no?

src/pruna/engine/handler/handler_utils.py
llcnt: Absolutely :)

src/pruna/engine/handler/handler_utils.py
llcnt: I refactored this to be sure to intercept diffusers pipeline first !

tests/algorithms/test_combinations.py
llcnt: Oh I did know, I removed it!

src/pruna/engine/utils.py
llcnt: I would say no. But I have added this to be sure we cover all changes on gptq side (changing the names of some parameters for eg). I can remove it if you think it is cleaner this way :)

src/pruna/engine/handler/handler_pipeline.py
llcnt: Yeah I simplified this!

src/pruna/engine/utils.py
llcnt: Yes for gptq some attributes of the quantized layer are not moved when we apply `.to`.
quanto also has some qweight attributes, but I did not have issues while moving a quanto quantized model

src/pruna/engine/handler/handler_pipeline.py
llcnt: Ok, I removed the `.to` :)

src/pruna/engine/handler/handler_utils.py
```diff
@@ -49,8 +50,17 @@ def register_inference_handler(model: Any) -> InferenceHandler:
 
     model_module = model._orig_mod.__module__ if hasattr(model, "_orig_mod") else model.__module__
 
+    # Prefer diffusers handler first to avoid routing diffusers pipelines to generic pipeline handler
     if "diffusers" in model_module:
         return DiffuserHandler(call_signature=inspect.signature(model.__call__))
+
```
begumcig, line 56: I really like this!! Thank youu

src/pruna/engine/handler/handler_pipeline.py
begumcig: Okay totally understandable! I see that you have removed the generation_config logic from this file, but i also don't see any changes to the transformershandler. How do we handle this max_tokens logic now?

src/pruna/engine/handler/handler_pipeline.py
begumcig: I completely understand the need for the current approach. If we really wanted to avoid patching, we could add a guard in `PrunaModel.run_inference`, e.g.:

```
if isinstance(self.inference_handler, PipelineHandler) and hasattr(batch[0], "shape") and hasattr(batch[0], "dtype"):
    self.run_inference(self.model.model, batch, device)
```

That said, it’s not really cleaner than patching. So I leave it upto you totally. 

Also just to flag: the patch is already doing some input/output handling, and `run_inference` (in the evaluation context) also calls prepare_inputs and process_outputs. Let’s make sure there isn’t unintended duplication and rerun the tests to be safe. From what I’ve seen you have already thought about this, `PipelineHandler.prepare_inputs` only splits the batch, and` process_output` only runs in the non-evaluation context. but just commenting to check I didn't get anything wrong and  we are on the same page🥺.

begumcig (approved):
Thanks so much for the updates! I’ve already approved the PR, but I left a question about `generation_config`. I couldn’t find the previous snippet elsewhere and wasn’t sure where it’s handled now. Also, let’s make sure the failing GPU tests still pass and that we’re not duplicating anything between the `__call__` patch and `prepare_inputs` / `process_outputs `in the `PipelineHandler`. If those look good to you too, we’re good to go!

src/pruna/engine/handler/handler_pipeline.py
llcnt: I would be in favor of keeping this patch here, to keep the `pruna_model.py` clean :)
Regarding your flag, your are right, in the evaluation context the prepare_inputs will just split the batch, and the process_output will do nothing special. I have re-run the 3 tests and they passed successfully: 
image

src/pruna/engine/handler/handler_pipeline.py
llcnt: Test related to TransformerHandler (with torch_compile) already passed successfully. Why should we change something there ? The `max_tokens` logic is now implemented in the `compilation/utils.py` file ;)

begumcig: Okay perfect I see it now!! Don't have any further questions

src/pruna/engine/handler/handler_pipeline.py
begumcig: Great! LET'S MERGE this then! Amazing job

PR #323, opened 2025-08-28
Rework Model Context

simlang:
## Description
This PR refactors the ModelContext abstraction
Before the ModelContext used the incoming pipeline as a storage, for the smashed model, now the context itself is returned to get resources from.
Using pipeline as storage device, lead to problems when e.g. using the combination hqq+torch.compile

The changes include:
1) instead of providing: pipeline, working model and denoiser_type, only provide model_context and working_model (denoiser_type was only used by one algorithm and can now be accessed via model_context.denoiser_type if needed)
2) at end of smashing, set_smashed_working_model(_smashed_model_) has to be called to tell the context, that the working model has changed
3) on context exit, the internal state of the pipeline/model given to the context is updated, but only if set_smashed_working_model has been called before - otherwise nothing happens. This allows us to use the context also, if the working model is not adapted
4) since the model given to the context might be immutable (if it's the working model and not a pipeline) we can't directly change it - this means to get the updated pipeline/model we have to call model_context.get_smashed() to return after smashing

## Type of Change

## How Has This Been Tested?
Locally ran all tests for algorithms which use ModelContext

## Checklist

## Additional Notes
There might be a point of moving the ModelContext to the apply wrapper, to avoid duplicate code

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -571,9 +571,18 @@ class ModelContext(AbstractContextManager):
                 return
             else:
                 raise RuntimeError(
-                    "ModelContext is not in read-only mode, but the working model was not updated. "
+                    "ModelContext is not in read-only mode, but the working model has not been updated. "
                     "Make sure to call `update_working_model` with the adapted model "
-                    "before exiting the context manager."
+                    "before exiting the context manager. After exiting the context manager retrieve "
+                    "the updated abstracted model via `get_updated_model` to e.g. return "
+                    "it after finishing the current compression algorithm."
+                )
+        else:
+            if self.read_only:
+                raise RuntimeWarning(
+                    "ModelContext is in read-only mode, but the working model has been updated. "
+                    "This can lead to unexpected behavior. If you want to update the model "
+                    "make sure to enter the context manager with `read-only=False` which is the default."
                 )
 
         if self.path_to_working_model is not None:


src/pruna/engine/utils.py
```diff
@@ -522,34 +523,34 @@ def __init__(self, model: "ModelMixin") -> None:
         model : ModelMixin
             The model to handle. Can be a transformer model, UNet, or other pipeline.
         """
-        self.pipeline = model
+        self.smashed_pipeline = model
```
gsprochette: self.model would be clearer I think

src/pruna/engine/utils.py
```diff
@@ -522,34 +523,34 @@ def __init__(self, model: "ModelMixin") -> None:
...
+        self.smashed_working_model = None
+        self.denoiser_type: str | None = None
```
gsprochette: Is self.denoiser_type being used somewhere? Isn't it the point of the model context to not have to use this?

src/pruna/engine/utils.py
```diff
@@ -564,17 +565,39 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
...
+    def get_smashed(self) -> "ModelMixin":
+        """
+        Get the smashed model.
+
+        Returns
+        -------
+        ModelMixin
+            The smashed model.
+        """
+        return self.smashed_pipeline
```
gsprochette: is there a reason for using this instead of `mc.smashed_working_model = working_model` and `mc.smashed_pipeline` from outside?

src/pruna/engine/utils.py
```diff
@@ -564,17 +565,39 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
...
         else:
-            self.pipeline = self.pipeline.working_model
-        # Only delete working_model if it was actually set as an attribute
-        # In the else case above, working_model and pipeline are the same object,
-        # so we don't need to (and can't) delete it
-        if hasattr(self.pipeline, "working_model") and self.pipeline.working_model is not self.pipeline:
-            del self.pipeline.working_model
-            safe_memory_cleanup()
+            self.smashed_pipeline = self.smashed_working_model
+
```
gsprochette: I would be in favor of not repeating the tests from before, and applying directly to the attribute currently named `self.denoiser_type`.

src/pruna/engine/utils.py
```diff
@@ -522,34 +523,34 @@ def __init__(self, model: "ModelMixin") -> None:
...
-        elif hasattr(self.pipeline, "unet"):
-            self.working_model = self.pipeline.unet
+        elif hasattr(self.smashed_pipeline, "unet"):
+            working_model = self.smashed_pipeline.unet
             self.denoiser_type = "unet"
-        elif hasattr(self.pipeline, "model") and hasattr(self.pipeline.model, "language_model"):
-            self.working_model = self.pipeline.model.language_model
+        elif hasattr(self.smashed_pipeline, "model") and hasattr(self.smashed_pipeline.model, "language_model"):
+            working_model = self.smashed_pipeline.model.language_model
             self.denoiser_type = "language_model"
```
gsprochette: denoiser_type in this case could be `model.language_model` so we can retrieve it without duplicating the logic

src/pruna/engine/utils.py
gsprochette: `self.working_model_attribute_name` or something like that would be clearer and less diffuser-centric

src/pruna/engine/utils.py
```diff
@@ -564,17 +565,39 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
...
         """
-        if hasattr(self.pipeline, "transformer"):
-            self.pipeline.transformer = self.pipeline.working_model
-        elif hasattr(self.pipeline, "unet"):
-            self.pipeline.unet = self.pipeline.working_model
-        elif hasattr(self.pipeline, "model") and hasattr(self.pipeline.model, "language_model"):
-            self.pipeline.model.language_model = self.pipeline.working_model
+        if self.smashed_working_model is None:
+            return
+
```
gsprochette: In which case would this be none? Isn't this a sign that the user forgot to assign this? I feel raising an error would be good

src/pruna/engine/save.py
```diff
@@ -358,14 +358,11 @@ def save_model_hqq(model: Any, model_path: str | Path, smash_config: SmashConfig
         quantized_path = Path(model_path)
 
     # save the quantized model only.
-    with ModelContext(model) as (pipeline, working_model, denoiser_type):
+    with ModelContext(model) as (mc, working_model):
```
gsprochette: `mc` is not being used here. I assume it will be if we add an error in the case where nothing was set for `mc.smashed_working_model`. Otherwise it should be `(_, working_model)`

src/pruna/engine/model_checks.py
```diff
@@ -609,9 +609,7 @@ def has_fused_attention_processor(pipeline: Any) -> bool:
     bool
         True if the pipeline can use a Fused attention processor, False otherwise.
     """
-    with ModelContext(pipeline) as (_, working_model, _):
-        # add this line just to make the __exit__ method of ModelContext work
-        pipeline.working_model = working_model
+    with ModelContext(pipeline) as (mc, working_model):
```
gsprochette: same comment as in save

src/pruna/engine/utils.py
gsprochette: I see there are cases where we don't do anything on the model and just need to read it. I think we can add a `read_only` option, or method or anything to signify that this has not been forgotten. I insist because it's not super natural to have to assign something inside the context object...

src/pruna/engine/utils.py
```diff
@@ -503,7 +504,7 @@ def set_to_best_available_device(device: str | torch.device | None) -> str:
         raise ValueError(f"Device not supported: '{device_str}'")
 
 
-class ModelContext:
+class ModelContext(AbstractContextManager):
```
llcnt, line 507: What does this AbstractContextManager provide us ? Any reason why we want our mc to depend on it? :)

gsprochette (changes requested):
Good job fixing this ! We're using a context in a weird way so we should spend a minute making it extra clear and extra clean, all my comments go in that direction :)

src/pruna/engine/utils.py
simlang: agree

src/pruna/engine/utils.py
llcnt: I guess when we save we dont use the set_smashed_workingmodel, so we get None, no?

src/pruna/engine/utils.py
simlang: This `self.smashed_working_model = working_model` was super cryptic to me, when i first read it before. So i add functions to name what is happening - so just readability

src/pruna/engine/utils.py
simlang: yes i first had a warning, but then saw that there are use cases where you do the read_only. adding a read_only flag makes sense, will do!

src/pruna/engine/utils.py
simlang: tbh nothing - just readability that this is a ContextManager - should i remove it?

src/pruna/engine/utils.py
gsprochette: I think in all cases the user needs to know what they are doing with the ModelContext, and these setter and getter are adding complexity... Using these variables could be explained within an error raised in the `__exit__` if `smashed_working_model` wasn't set.

simlang: i would rename the functions, as with the current names there is not really a difference to just assigning and reading.
i like having this abstraction however, without ever seeing the inside of this context it is hard to understand from the outside what is happening

src/pruna/engine/utils.py
```diff
@@ -566,7 +567,14 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
             The traceback.
         """
         if self.smashed_working_model is None:
-            return
+            if self.read_only:
```
gsprochette, line 570: We could also check `if self.read_only and self.smashed_working_model is not None` because this is bound to produce a cryptic bug

src/pruna/engine/utils.py
```diff
@@ -566,7 +567,14 @@ def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any) -> None:
...
+                return
+            else:
+                raise RuntimeWarning(
+                    "ModelContext is not in read-only mode, but the working model was not updated. "
+                    "Make sure to call `update_working_model` with the adapted model "
+                    "before exiting the context manager."
```
gsprochette: This is indeed the opportunity to mention `update_working_model`, but it should also mention the other method to get the updated model after exiting because the user seeing this message probably does not know about it

gsprochette (approved):
That's almost ready to go, I left a couple of comments in the read_only check because it can still be improved. Once this is done, you can merge :)

src/pruna/engine/utils.py
simlang: very good point!

src/pruna/engine/utils.py
simlang: also good point!

PR #327, opened 2025-09-01
fix: add-dataset-and-inferencer-for-janus-tests

gsprochette:
## Description
This PR works on the Janus related tests, updating some required arguments in the inference handler along the way. It fixes the tests by:
- adding a dataset to the janus fixture
- adding an inference handler for janus
- reworking the hqq load function to align compute dtype
- droping the tiny-janus and using janus 1b while waiting for a working tiny-janus compatible with the processor

Edge cases: Currently only text-to-image is tested even though Janus can handle text and/or image to text or image. Multimodal will require more work on the handler and adding new datasets. The updates we made still aim to not be in the way of this future extension.

## Related Issue
Fixes Nightly HQQ Janus test

## Type of Change

## How Has This Been Tested?
Ran the HQQ janus tests that previously crashed, they now pass

## Checklist


src/pruna/engine/handler/handler_autoregressive.py
```diff
@@ -0,0 +1,131 @@
...
+        # recognize JanusProcessor by its class name because it requires transformers>=4.52.0
+        elif self.smash_config.processor.__class__.__name__ != "JanusProcessor":
+            raise ValueError(f"Expected a JanusProcessor, but got {type(self.smash_config.processor)}")
+
+        # processor is a JanusProcessor, meaning transformers>=4.52.0, so we can now us it for type casting
+        processor = cast(transformers.JanusProcessor, self.smash_config.processor)
+
+        # image generation mode
+        text, _ = batch
+        text = cast(str, text)
```
begumcig: I might be wrong here but I think the first element of the batch is always a list. Also what would happen if batch size is larger than 1?

tests/algorithms/testers/quantization.py
```diff
@@ -60,13 +60,30 @@ class TestDiffusersInt8(AlgorithmTesterBase):
...
+    """Test the HQQ quantizer."""
+
+    models = ["janus_pro"]
+    reject_models = ["sd_tiny_random"]
+    allow_pickle_files = False
+    algorithm_class = HQQQuantizer
+    metrics = ["cmmd"]
+    hyperparameters = {"hqq_compute_dtype": "torch.bfloat16"}
+
+    def post_smash_hook(self, model: PrunaModel) -> None:
```
johannaSommer, line 80: Can we absorb this into the inference handler so that the user does not have to set this?

tests/fixtures.py
```diff
@@ -176,7 +182,7 @@ def get_automodel_image_text_to_text_transformers(model_id: str) -> tuple[Any, S
     ),
     "dummy_lambda": dummy_model,
     # image generation AR models
-    "tiny_janus_pro": partial(get_automodel_image_text_to_text_transformers, "loulou2/tiny_janus"),
+    "janus_pro": partial(get_autoregressive_text_to_image_model, "deepseek-community/Janus-Pro-1B"),
```
johannaSommer: If the janus fixture is not compatible with the processor, can we update the fixture instead (and just keep it under the pruna org?)

johannaSommer (changes requested):

tests/algorithms/testers/quantization.py
gsprochette: Here is a breakdown for each argument:
- `generation_mode` has to do with the fact that Janus is multimodel, it could generate text or image so at some point we need to tell it what to do. I could add a default value in tghe handler, but having no default value means reusing the one defined by Janus, which is text. I think it makes more sense to keep the default that is already in the code and override it here, what do you think?
- `do_sample` is optional, it's a matter of choosing the token with highest predicted probability or sampling it randomly from those probability. It doesn't need to be set, it's just a matter of preferences. I'm in favor of leaving it here to test the full functionalities and still not override the defaults.
- `use_cache` is set to True for image generation anyway after checking the generation mode. I'd rather set it anyway just in case.

tests/fixtures.py
gsprochette: That's the plan, but until we do can we keep this so we still have a test?

begumcig (commented):
Already looks super good to me, I just have one question about the casting, probably missing something. Thank you so much! 💜

src/pruna/engine/handler/handler_autoregressive.py
gsprochette: You are very very correct

src/pruna/engine/handler/handler_autoregressive.py
```diff
@@ -0,0 +1,127 @@
...
+                "`model.inference_handler.processor = ...`"
+            )
+        elif not isinstance(self.smash_config.processor, JanusProcessor):
+            raise ValueError(f"Expected a JanusProcessor, but got {type(self.smash_config.processor)}")
+
+        # image generation mode
+        text, _ = batch
+        text = cast(str, text)
+        inputs = self.smash_config.processor(text=text, generation_mode=generation_mode, return_tensors="pt")
+        inputs = inputs.to(self.smash_config.device, dtype=self.model.dtype)
```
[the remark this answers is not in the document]
gsprochette: Fixed it by using the predefined `move_inputs_to_device` and `set_correct_dtype` methods.
@begumcig the `text, _ = batch` thing is inherited from other handlers, I agree with the bot that this makes assumptions which are not reflected in the signature. We should clean this up in a future PR.

tests/algorithms/testers/quantization.py
gsprochette: The last two are optional as developped in the previous message. I set the "generation_mode" as default for the handler since it's the only supported mode right now. In the future, the user will have to set a generation mode so I still raise a warning if it's not manually set.

tests/fixtures.py
gsprochette: I have tested multiple options and there is currently a dimension error, the `image_tokens` passed to the vqmodel are dimension 27 but they are expected to be dimension 576. The current `tiny-janus` was built to run the generate method, but it turns out to be incompatible with the `model.decoding` step. I tried manually duplicating the output dim to resolve that, but it doesn't work.

gsprochette: Pinging @llcnt , there's probably something about the transformers version because I tried re-making tiny-janus the same way and it works in my notebook. This means the problem is almost surely:
1. transformer version update changed the way the model is saved/loaded and now a file is missing because tiny-janus was saved before this, or
2. something is lost when save-loading the tiny-janus model

Also we can make tiny-janus with the 1b model as a base instead of 7b to make it even tinier

llcnt: I had a quick look at the `transformers` codebase and indeed there was a lot of commit affecting Janus since the earlier version I was using. What is your current version? Did you try to save and re-load your tiny-janus model and test if it works? If Yes, I guess you can just change the model in `fixtures.py`, no?

gsprochette: I remade the tiny_janus random test model and it now works. Two changes of note:
- I derived the config from the 1b model's, which makes it a little smaller than previously since it was made from the 8b model
- I removed more layers than before

The new model is now 1.2Gb total

tests/fixtures.py
```diff
@@ -187,7 +191,7 @@ def get_autoregressive_text_to_image_model(model_id: str) -> tuple[Any, SmashCon
     ),
     "dummy_lambda": dummy_model,
     # image generation AR models
-    "janus_pro": partial(get_autoregressive_text_to_image_model, "deepseek-community/Janus-Pro-1B"),
+    "tiny_janus_pro": partial(get_autoregressive_text_to_image_model, "gsprochette/tiny_janus"),
```
johannaSommer: I would recommend transferring this to the Pruna HF organization so that we can all edit this if necessary (bertrand asked us to do this previously)

johannaSommer (approved):
See just my one comment about the location of the tiny fixture, otherwise LGTM!

src/pruna/engine/handler/handler_autoregressive.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # processor is a JanusProcessor, meaning transformers>=4.52.0, so we can now us it for type casting
+        processor = cast(transformers.JanusProcessor, self.smash_config.processor)
+
+        # image generation mode
+        text, _ = batch
+        text = cast(List[str], text)
+        inputs = processor(text=text, generation_mode=generation_mode, return_tensors="pt")
+        inputs = cast(BatchFeature, inputs)  # BatchFeature has a 'to' method
+        inputs = inputs.to(device=get_device(self.model), dtype=determine_dtype(self.model))
```
begumcig: Out of curiosity, we have a move_inputs_to_device function in the standard handler, why we cannot use this here and need to go through this BatchFeature cast/decasting?

begumcig (approved):
Have one question about device casting the inference outputs but already looks good to me so approving! Thanks a lot Gaspar! (inference go brrr)

src/pruna/engine/handler/handler_autoregressive.py
gsprochette: You're super right :)

tests/fixtures.py
gsprochette: this has been transfered to the pruna-test org

PR #333, opened 2025-09-05
feat: add target modules to bnb quantizers

gsprochette:
## Description
Add Target Modules to BnB quantizers. To do so, I also introduced a routine to get the ignored modules, and restrict any target modules to only leaf modules (Linear, Conv1d, but not SelfAttention for example). These new functionalities are usefull to apply target modules to HF based quantizers.

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
Added new tests counting the number of quantized modules before and after, + notebook

## Checklist


src/pruna/config/target_modules.py
```diff
@@ -102,6 +121,8 @@ def expand_list_of_targeted_paths(target_modules: TARGET_MODULES_TYPE, model: An
         The model to get the module paths from.
     target_modules : TARGET_MODULES_TYPE
         The target modules to convert to a list of module paths.
+    leaf_modules : bool
```
simlang: maybe change to `only_leaf_modules`?
Otherwise one could think, that this in/excludes only the leaf_modules

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -118,51 +153,64 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
-                model.unet = smashed_latent
-            else:
-                model = smashed_latent
-            return model
+        target_modules = smash_config["target_modules"]
+        if target_modules is None:
+            target_modules = self.get_unconstrained_hyperparameter_defaults(model, smash_config)
+        target_modules = cast(TARGET_MODULES_TYPE, target_modules)
+
+        def quantize_latent_model(attr_name: str | None, latent_model: nn.Module, subpaths: list[str]) -> Any:
```
simlang: why apply to latent_model?
i would just call it `quantize_working_model`, i feel like latent model might only apply for latent diffusion models
also rename the parameter

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -118,51 +153,64 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+            subpaths : list[str]
+                The subpaths of the working model to quantize.
+            """
+            if not hasattr(latent_model, "save_pretrained") or not callable(latent_model.save_pretrained):
+                raise ValueError(
+                    "diffusers-int8 was applied to a module which didn't have a callable save_pretrained method."
+                )
+            skipped_modules = get_skipped_submodules(latent_model, subpaths, leaf_modules=True)
+
+            with tempfile.TemporaryDirectory(prefix=str(smash_config["cache_dir"])) as temp_dir:
```
simlang, line 198: Not part of this PR, but since we are already touching this, i would suggest adding a diffusers_int8 subdirectory so things are kept seperate for every algo
```suggestion
            with tempfile.TemporaryDirectory(prefix=str(smash_config["cache_dir"] / "difussers_int8")) as temp_dir:
```

src/pruna/algorithms/quantization/huggingface_llm_int8.py
```diff
@@ -104,35 +134,64 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                raise ValueError(
+                    "llm-int8 was applied to a model (or part of a model) which is not a causal language model."
+                )
+            causal_lm = cast(PreTrainedModel, causal_lm)
+
+            # get the skipped modules, only include leaf modules since the bnb quantizer skips all submodules
+            # in a skipped module. Only Linear and Conv1d layers can be quantized anyway.
+            skipped_modules = get_skipped_submodules(causal_lm, subpaths, leaf_modules=True)
+
+            with tempfile.TemporaryDirectory(prefix=str(smash_config["cache_dir"])) as temp_dir:
```
simlang, line 181: Same thing here
```suggestion
            with tempfile.TemporaryDirectory(prefix=str(smash_config["cache_dir"] / "transformers_int8")) as temp_dir:
```

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -118,51 +153,64 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                The latent model to quantize.
+            subpaths : list[str]
+                The subpaths of the working model to quantize.
+            """
+            if not hasattr(latent_model, "save_pretrained") or not callable(latent_model.save_pretrained):
+                raise ValueError(
+                    "diffusers-int8 was applied to a module which didn't have a callable save_pretrained method."
+                )
+            skipped_modules = get_skipped_submodules(latent_model, subpaths, leaf_modules=True)
+
```
simlang, line 197: should we maybe on a debug level log the skipped modules? If you agree, the same thing in llm_int8

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -102,6 +110,33 @@ def model_check_fn(self, model: Any) -> bool:
...
+            The default target_modules for the algorithm.
+        """
+        prefix: str
+        if hasattr(model, "transformer"):
+            prefix = "transformer."
+        elif hasattr(model, "unet"):
+            prefix = "unet."
+        else:
+            prefix = ""
+        return {"include": [prefix + "*"], "exclude": [prefix + "lm_head"]}
```
simlang: maybe a more general discussion on the defaults - the defaults here contain some key configuration so the model doesn't break from quantization, which we hide from the user. a user might decide to exclude some specific module for their use case but doesn't know about those defaults are necessary for the model to work and wonder why the model breaks.
Should we join the defaults with the user config instead of overwriting them?

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
```diff
@@ -102,6 +110,33 @@ def model_check_fn(self, model: Any) -> bool:
 
         return hasattr(model, "unet") and isinstance(model.unet, tuple(transformer_and_unet_models))
 
+    def get_unconstrained_hyperparameter_defaults(
```
simlang: update to new name
```suggestion
    def get_model_dependent_hyperparameter_defaults(
```

src/pruna/algorithms/quantization/huggingface_llm_int8.py
```diff
@@ -88,6 +97,27 @@ def model_check_fn(self, model: Any) -> bool:
         """
         return is_causal_lm(model) or is_transformers_pipeline_with_causal_lm(model)
 
+    def get_unconstrained_hyperparameter_defaults(
```
simlang: update to new name
```suggestion
    def get_model_dependent_hyperparameter_defaults(
```

simlang (changes requested):
Looks amazing! Mostly some naming comments, and a general discussion how we should handle defaults, let me know what you think!

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
gsprochette: One option I saw is a `add_to_default` flag, if it's true then you the target modules is `dict(include = user_defined_include + include_default, exclude = user_defined_exclude + exclude_default)`. I have that in mind it's coming in a future PR

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
gsprochette: I think this is super not part of this PR, we can do this in an other PR if it makes things cleaner / if we need this for a new feature

src/pruna/algorithms/quantization/huggingface_diffusers_int8.py
gsprochette: Makes sense yes

src/pruna/algorithms/quantization/huggingface_llm_int8.py
simlang: has not been applied

src/pruna/algorithms/quantization/huggingface_llm_int8.py
```diff
@@ -104,35 +135,67 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
-
-            smashed_model = AutoModelForCausalLM.from_pretrained(
-                temp_dir,
-                quantization_config=bnb_config,
-                trust_remote_code=True,
-                torch_dtype=smash_config["compute_dtype"],  # storage type of the non-int8 params
-                device_map=device_map,
+        target_modules = smash_config["target_modules"]
+        if target_modules is None:
+            target_modules = self.get_unconstrained_hyperparameter_defaults(model, smash_config)
```
simlang:
```suggestion
            target_modules = self.get_model_dependent_hyperparameter_defaults(model, smash_config)
```

simlang (changes requested):
Bugbot caught already a lot, just some renaming-issues.

src/pruna/algorithms/quantization/huggingface_llm_int8.py
gsprochette: Yes, sorry, it has now been applied

src/pruna/algorithms/quantization/huggingface_llm_int8.py
```diff
@@ -104,35 +135,67 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+            # in a skipped module. Only Linear and Conv1d layers can be quantized anyway.
+            skipped_modules = get_skipped_submodules(causal_lm, subpaths, only_leaf_modules=True)
+            pruna_logger.debug(
+                f"Skipped leaf modules within {attr_name or 'the model'} in {self.algorithm_name}: {skipped_modules}"
             )
 
-        return smashed_model
+            with tempfile.TemporaryDirectory(prefix=str(smash_config["cache_dir"])) as temp_dir:
+                # cast original model to CPU to free memory for smashed model
+                device_map = get_device_map(causal_lm)
```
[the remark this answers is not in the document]
gsprochette: I don't think so, we have a specific exception for the `Pipeline`s

PR #340, opened 2025-09-09
Fix generic loading function for benchmarking

sharpenb:
## Description
This PR enables to load any diffusers and transformers models even if they do not have smash configs. This is useful for benchmarking of all models regardless if they are wrapped or not with Pruna.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/engine/load.py
+++ b/src/pruna/engine/load.py
@@ -216,7 +216,7 @@ def resmash(model: Any, smash_config: SmashConfig) -> Any:
     return smash(model=model, smash_config=smash_config_subset)
 
 
-def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwargs) -> Any:
+def load_transformers_model(path: str | Path, smash_config: SmashConfig | None = None, **kwargs) -> Any:
     """
     Load a transformers model or pipeline from the given model path.
 
@@ -249,12 +249,16 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
         architecture = config["architectures"][0]
         cls = getattr(transformers, architecture)
         # transformers discards kwargs automatically, no need for filtering
-        device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
-        device_map = smash_config.device_map if smash_config.device == "accelerate" else device
+        if smash_config is not None:
+            device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
+            device_map = smash_config.device_map if smash_config.device == "accelerate" else device
+        else:
+            device = "cuda" if torch.cuda.is_available() else "cpu"
+            device_map = "auto"
         return cls.from_pretrained(path, device_map=device_map, **kwargs)
 
 
-def load_diffusers_model(path: str | Path, smash_config: SmashConfig, **kwargs) -> Any:
+def load_diffusers_model(path: str | Path, smash_config: SmashConfig | None = None, **kwargs) -> Any:
     """
     Load a diffusers model from the given model path.
 
@@ -296,7 +300,8 @@ def load_diffusers_model(path: str | Path, smash_config: SmashConfig, **kwargs) 
     # diffusers discards kwargs automatically, no need for filtering
     # diffusers does not support device_maps as dicts at the moment, we load on cpu and cast it ourselves
     model = cls.from_pretrained(path, device_map=None, **kwargs)
-    move_to_device(model, smash_config.device, device_map=smash_config.device_map)
+    if smash_config is not None:
+        move_to_device(model, smash_config.device, device_map=smash_config.device_map)
     return model
 
 


src/pruna/engine/load.py
```diff
@@ -249,12 +249,16 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
         architecture = config["architectures"][0]
         cls = getattr(transformers, architecture)
         # transformers discards kwargs automatically, no need for filtering
-        device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
-        device_map = smash_config.device_map if smash_config.device == "accelerate" else device
+        if smash_config is not None:
```
gsprochette: personal preference you can choose to ignore: start with the simpler `if smash_config is None` and switch the two blocks

src/pruna/engine/load.py
```diff
@@ -249,12 +249,16 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
...
+            device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
+            device_map = smash_config.device_map if smash_config.device == "accelerate" else device
+        else:
+            device = "cuda" if torch.cuda.is_available() else "cpu"
```
gsprochette: I would recommend using `device = pruna.engine.set_to_best_available_device(None)` to reuse the auto device choice, and to also support mps devices

src/pruna/engine/load.py
```diff
@@ -216,7 +216,7 @@ def resmash(model: Any, smash_config: SmashConfig) -> Any:
     return smash(model=model, smash_config=smash_config_subset)
 
 
-def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwargs) -> Any:
+def load_transformers_model(path: str | Path, smash_config: SmashConfig | None = None, **kwargs) -> Any:
```
gsprochette, line 218: Shouldn't the user be able to provide a device here in the case they don't provide a smash_config? Should we have `device` and `device_map` arguments with default None that can only be used when smash_config is None?

src/pruna/engine/load.py
```diff
@@ -249,12 +249,16 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig, **kwarg
...
+            device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
+            device_map = smash_config.device_map if smash_config.device == "accelerate" else device
+        else:
+            device = "cuda" if torch.cuda.is_available() else "cpu"
+            device_map = "auto"
         return cls.from_pretrained(path, device_map=device_map, **kwargs)
 
 
-def load_diffusers_model(path: str | Path, smash_config: SmashConfig, **kwargs) -> Any:
+def load_diffusers_model(path: str | Path, smash_config: SmashConfig | None = None, **kwargs) -> Any:
```
gsprochette, line 264: same comment as for the transformers function

src/pruna/engine/load.py
```diff
@@ -296,7 +300,8 @@ def load_diffusers_model(path: str | Path, smash_config: SmashConfig, **kwargs)
     # diffusers discards kwargs automatically, no need for filtering
     # diffusers does not support device_maps as dicts at the moment, we load on cpu and cast it ourselves
     model = cls.from_pretrained(path, device_map=None, **kwargs)
-    move_to_device(model, smash_config.device, device_map=smash_config.device_map)
+    if smash_config is not None:
+        move_to_device(model, smash_config.device, device_map=smash_config.device_map)
```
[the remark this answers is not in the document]
gsprochette: I fully agree with the bot, the if statement should determine where to get the device and device map from, but the move_to_device should be done regardless

src/pruna/engine/load.py
gsprochette: also here device is never used

gsprochette (changes requested):
Looks good, a single iteration should do it :) 
- I would recommend renaming the PR to something in the line of `feat: allow loading models without a smash config`.
- If we decide to add a device_map argument in the loading functions, that can be done in an other PR

src/pruna/engine/load.py
sharpenb: I updated the comment taht it will use by default cuda and auto if the smash config doe snot exist and hence no device/device map is provided :)

src/pruna/engine/load.py
```diff
@@ -251,18 +251,18 @@ def load_transformers_model(path: str | Path, smash_config: SmashConfig | None =
...
         # transformers discards kwargs automatically, no need for filtering
-        if smash_config is not None:
-            device = smash_config.device if smash_config.device != "cuda" else "cuda:0"
-            device_map = smash_config.device_map if smash_config.device == "accelerate" else device
-        else:
-            device = "cuda" if torch.cuda.is_available() else "cpu"
-            device_map = "auto"
+        if smash_config is None:
+            smash_config = SmashConfig()
+            smash_config["device"] = "auto"
```
gsprochette: why not just `device_map = "auto"` in the if statement and the device/device_map from the smash_config in the else? Here the "auto" value for device seems non legal for the device utility functions so we shouldn't set that value inside smash_config I think

src/pruna/engine/load.py
```diff
@@ -217,7 +217,7 @@ def resmash(model: Any, smash_config: SmashConfig) -> Any:
 
 def load_transformers_model(path: str | Path, smash_config: SmashConfig | None = None, **kwargs) -> Any:
     """
-    Load a transformers model or pipeline from the given model path.
+    Load a transformers model or pipeline from the given model path. If a SmashConfig is not provided, it will default to "auto" for the device_map.
```
gsprochette: This case should be described in the description of the smash_config argument

PR #341, opened 2025-09-11
fix: memory related issues in the nightlies for 3 tests

begumcig:
## Description

- The Transformers pipelines were incorrectly configured to raise an error for non–text-generation models. Instead, it should always fall back to the standard inference handler.
- Memory metrics were being collected from a CPU copy of the model, rather than the active model instance.
- Certain models triggered out-of-memory (OOM) errors in nightly tests because the machine lacked support for efficient attention. To mitigate this, attention slicing has been added to the test fixtures.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -79,6 +79,11 @@ def get_diffusers_model(model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, S
     # snapshot download of the model
     model_path = snapshot_download(model_id)
     model = load_diffusers_model(model_path, smash_config=SmashConfig(device="cpu"), **kwargs)
+    # safely enable attention slicing if supported
+    # the gpus of the CI do not support efficient attention backends.
+    # we try to avoid OOM errors by using attention slicing.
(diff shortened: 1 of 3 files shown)


tests/fixtures.py
```diff
@@ -79,6 +80,11 @@ def get_diffusers_model(model_id: str, **kwargs: dict[str, Any]) -> tuple[Any, S
     # snapshot download of the model
     model_path = snapshot_download(model_id)
     model = load_diffusers_model(model_path, smash_config=SmashConfig(device="cpu"), **kwargs)
+    # safely enable attention slicing if supported
+    # the gpus of the CI do not support efficient attention backends.
+    # we try to avoid OOM errors by using attention slicing.
+    with suppress(Exception):
```
gsprochette: Is there no way to be a bit more specific about the Exception that should be suppressed? That's fine if the answer is no but I need to check

begumcig: Yes super good idea, I have changed it so we only set slicing if the model has the attribute :)

PR #342, opened 2025-09-12
fix: algorithms compatibility with inference tests

begumcig:
## Description

- Added a check for algorithms that don’t support model saving/loading to prevent inference tests from failing.

- Removed automatic casting of EvaluationAgent to CPU.

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/engine/pruna_model.py
+++ b/src/pruna/engine/pruna_model.py
@@ -400,14 +400,3 @@ class PrunaModel:
     def destroy(self) -> None:
         """Destroy model."""
         pass
-
-    def is_empty(self) -> bool:
-        """
-        Check if the model is empty.
-
-        Returns
-        -------
-        bool
-            True if the model is empty, False otherwise.
-        """
-        return self.model is None


tests/common.py
```diff
@@ -85,7 +87,12 @@ def run_full_integration(
         smashed_model = algorithm_tester.execute_smash(model, smash_config)
         algorithm_tester.execute_save(smashed_model)
         safe_memory_cleanup()
-        reloaded_model = algorithm_tester.execute_load()
+        # For algorithms that do not support saving/loading
+        reloaded_model = (
+            smashed_model
+            if is_function_unimplemented(algorithm_tester.execute_load)
```
simlang: I'm not the biggest fan of this check.
Raising a NotImplemented in the execute_load or something similar feels a bit cleaner 🪥

simlang (changes requested):

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -153,8 +152,8 @@ def prepare_model(self, model: Any) -> PrunaModel:
             model = PrunaModel(model, smash_config=smash_config)
             pruna_logger.info("Evaluating a base model.")
             is_base = True
-
-        model.inference_handler.log_model_info()
+        if hasattr(model, "inference_handler"):  # Distributers do not have an inference handler
+            model.inference_handler.log_model_info()
```
gsprochette, line 156: I could see having an else statement here logging a warning that the model doesn't have an inference handler. Or maybe such a warning is already logged when trying to attach an inference handler to the smashed model?

tests/common.py
```diff
@@ -85,7 +87,12 @@ def run_full_integration(
         smashed_model = algorithm_tester.execute_smash(model, smash_config)
         algorithm_tester.execute_save(smashed_model)
```
gsprochette: Should the model still be saved if the algorithm doesn't have a save/load function?

tests/common.py
gsprochette: I agree, especially considering the complexity of the `is_function_unimplemented` a `try .. except NotImplementedError` feels a lot cleaner and maintainable

src/pruna/evaluation/evaluation_agent.py
begumcig: This only happens with distributors that run their own server inference setup. That’s why I felt a warning might be misleading. We do handle inference, just not in the traditional way :)

gsprochette: Ok that makes sense, this is fine then. Not at all a priority but maybe dispatching to the inference server can still be an `inference_handler` in that case, meaning that we also get a `log_model_info`. In all cases that's definitely not for this PR.

tests/common.py
begumcig: We don't do anything if the algorithm doesn't support it, but we need it for the algorithms that do save.

tests/common.py
```diff
@@ -85,7 +85,11 @@ def run_full_integration(
...
         safe_memory_cleanup()
-        reloaded_model = algorithm_tester.execute_load()
+        # For algorithms that do not support saving/loading
+        try:
+            reloaded_model = algorithm_tester.execute_load()
+        except NotImplementedError:
+            reloaded_model = smashed_model
```
simlang, line 90: yes, way cleaner 😍  Are you adding the NotImplemented throwing to the respective testers in another PR?

simlang (approved):
LGTM! 🚀 
Had only one question, still approving it though!

tests/common.py
```diff
@@ -83,9 +83,16 @@ def run_full_integration(
...
-        algorithm_tester.execute_save(smashed_model)
+        # For algorithms that do not support saving/loading
+        try:
+            algorithm_tester.execute_save(smashed_model)
+            reloaded_model = algorithm_tester.execute_load()
+        except NotImplementedError:
+            reloaded_model = smashed_model
+        else:
+            if reloaded_model.is_empty():  # We have a save but no load.
+                reloaded_model = smashed_model
```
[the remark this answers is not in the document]
gsprochette: The "we have a save but no load" case would crash with NotImplementedError at the loading step, so this if check should be inside the except block I think, and raise an error with an explicit message such as "Error when reloading the model, the algorithm seems to implement a save function but no load function.".
Also the `safe_memory_cleanup()` seems to have disappeared.

tests/common.py
```diff
@@ -83,9 +83,15 @@ def run_full_integration(
...
-        safe_memory_cleanup()
-        reloaded_model = algorithm_tester.execute_load()
+
+        try:
+            algorithm_tester.execute_save(smashed_model)
+        except NotImplementedError:  # For algorithms that do not support saving/loading
+            reloaded_model = smashed_model
+        else:  # saving was successful, so we can safely clean up memory and reload
+            safe_memory_cleanup()
+            reloaded_model = algorithm_tester.execute_load()
```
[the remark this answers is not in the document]
gsprochette: This is intended behavior

PR #349, opened 2025-09-18
feat: add new fixtures as pre-quantized llms

llcnt:
## Description
This PR just adds 2 new models in `fixtures.py`.
It is particular because it does not follow the standard way of handing fixtures (ie tied to hf original models, with an additional SmashConfig). Instead we load the path to a pre-quantized model. To fit with the current state of the codebase, we also load a dummy model, and put the pre-quantized model path into the SmashConfig.

Why this weird behaviour?
-> Because we implement some tests for vLLM in this [PR](https://github.com/PrunaAI/prunatree/pull/228). And vLLM can not 'smash' a model, but intend to run inference.

## Type of Change

## Checklist

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -169,9 +169,10 @@ def get_autoregressive_text_to_image_model(model_id: str) -> tuple[Any, SmashCon
 
 def get_pre_quantized_model(model_id: str) -> tuple[Any, SmashConfig]:
     """
-    Get a pre-quantized model.
+    Create a SmashConfig with a new attribute for the path to the (already) quantized model.
 
-    The fixture emits a dummy model and a SmashConfig carrying path to quantized model.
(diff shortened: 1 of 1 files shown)


tests/fixtures.py
```diff
@@ -155,6 +155,30 @@ def get_automodel_image_text_to_text_transformers(model_id: str) -> tuple[Any, S
     return model, smash_config
 
 
+def get_pre_quantized_model(model_id: str) -> tuple[Any, SmashConfig]:
+    """
+    Get a pre-quantized model.
+
+    The fixture emits a dummy model and a SmashConfig carrying path to quantized model.
```
sharpenb: What do you mean by prequantized model and dummy model? It would be nice to clarify what quantized model we mean here.

Issue #352, opened 2025-09-30
[FEATURE] Migrate type checking from `mypy` to `ty`


johannaSommer:
## **✨ What You’ll Do**
We currently use mypy as our static type checker. The project [ty](https://github.com/astral-sh/ty) provides a more modern developer experience.

You’ll migrate the repository from `mypy` to `ty`.

## **🛠️ Dependency**
- [ ] Remove `mypy` as a dependency.
- [ ] Add `ty` as a new development dependency.

## 🤖 Useful Resources
- [ty documentation](https://github.com/astral-sh/ty)
- CI configuration in .github/workflows/ needs to be updated to run ty instead of mypy.
- Contributor documentation (README, CONTRIBUTING.md) should be updated to reflect the new command.

Suggested main steps:
- [ ] Uninstall/remove `mypy`.
- [ ] Install/add ty to the dev environment.
- [ ] Update GitHub Actions workflow to run `ty`.
- [ ] Run `ty` locally to confirm type checks succeed.

## ➡ Desired Solution
Before:
```bash
mypy src/
````
After:
```bash
ty src/
````
---

## **✅ Acceptance Criteria**
- [ ] `mypy` is fully removed.
- [ ] `ty` is installed and running in CI.
- [ ] Local type checking works using `ty`.
- [ ] Contributor documentation reflects the updated command.
- [ ] Tests & Docs: All existing and new unit tests pass, and the documentation is updated.

And don’t forget to give us a ⭐️!

---

## **❓ Questions?**

Feel free to jump into the [#contributing Discord channel](https://discord.com/invite/JFQmtFKCjd) if you hit any roadblocks. Can’t wait to see your contribution! 🚀

---

## Share on X

[![Share on X](https://img.shields.io/badge/Share_on_X-000?logo=x&logoColor=white)](https://twitter.com/intent/tweet?text=I%20just%20contributed%20to%20%40PrunaAI%20open%20source%20project!%20%F0%9F%9A%80%20Check%20out%20my%20PR%20%F0%9F%91%89%20%5BADD%20LINK%20TO%20PR%5D)

Dwid4r:
Hi @johannaSommer 
can you assign me?

DhanashreePetare:
@johannaSommer, can you please assign this to me

supakornn:
@johannaSommer  Did you mean this issue?

PR #360, opened 2025-09-30
migrate from mypy to Ty.

Supakornn:
## Description

Migrate from mypy to ty.

## Related Issue

Fixes #(352)

## Type of Change

- [] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- `ty check src/pruna`

## Checklist


supakornn:
Thanks @johannaSommer for the feedback! I've investigated the ty check results and found that ty is indeed much stricter than mypy by default. I was getting 18 errors/warnings initially, but I've successfully configured ty to behave more similarly to mypy for the transition period.
 
  What I found:
  • ty has stricter rules for Any assignments, dynamic method calls, tuple indexing, module attributes, and union syntax
  • The main differences were in rules like invalid-assignment, call-non-callable, index-out-of-bounds, unresolved-attribute, possibly-unbound-attribute, redundant-cast, unsupported-operator, and invalid-argument-type

  Solution implemented:
  I've updated the pyproject.toml configuration to ignore the stricter rules that mypy doesn't enforce:

```py
     [tool.ty.rules]
     # Ignore rules that are stricter than mypy for transition period
     unresolved-import = "ignore"
     invalid-assignment = "ignore"  # mypy is more permissive with Any assignments
     call-non-callable = "ignore"   # mypy allows more dynamic method calls
     index-out-of-bounds = "ignore" # mypy is more permissive with tuple indexing
     unresolved-attribute = "ignore" # mypy is more permissive with module attributes
     possibly-unbound-attribute = "ignore" # mypy doesn't warn about this as much
     redundant-cast = "ignore"      # mypy doesn't warn about redundant casts
     unsupported-operator = "ignore" # mypy supports | syntax with from __future__ import annotations
     invalid-argument-type = "ignore" # mypy is more permissive with argument types
```

  Result:
  After this configuration, `ty check src/pruna` now passes with "All checks passed!" - going from 18 errors/warnings to 0.
  Would you like me to commit this configuration change, or would you prefer to review it first?

johannaSommer:
@Supakornn thank you so much for the detailled response, I will have a look! In the meantime, could you briefly comment in the issue #360 so that I can assign this issue to you?

supakornn:
You mean issue #352, right?

johannaSommer:
@Supakornn thanks again a lot for your detailed answer, I think in any case its a great idea to adopt the rules as you detailed for the reasons we discussed. Could you add them to the `pyproject.toml`? Then we can run the linting GitHub action on this PR and once that passes we should be good to go 😊

johannaSommer:
Hey there @Supakornn thanks for adding them! Feel free to uncomment them, then I can run the linting workflow and we can check if we're ready to go or if there are still errors/warnings in the GitHub action 🫡😊

supakornn:
@johannaSommer  My bad, forgot to uncomment them initially! Fixed now and ready for the linting workflow check. 👍

johannaSommer:
@Supakornn Perfect, thanks! I ran the linting and there are still some failures, can you have a look? You can ignore the "test" failures, they are not rleated to the content of this PR 😊

supakornn:
@johannaSommer  All ty errors should be fixed now! The issues were mostly from load_dataset() and some helper functions returning union types that ty can’t fully check. I added specific # type: ignore[index|return-value|misc] comments where needed to handle those while keeping type safety everywhere else.

.pre-commit-config.yaml
```diff
@@ -20,13 +20,6 @@ repos:
         args: [--fix]
       - id: ruff-format
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
```
sdiazlor, line 23: Hey! Really cool @Supakornn! Just a small question: Did you check this thread? https://github.com/astral-sh/ty/issues/269. It might be useful to still have `ty` in our pre-commits.

minettekaum (approved):
lgtm as well 😄

PR #380, opened 2025-10-02
feat(evaluation): Add lm-eval to Pruna Metrics

### LINKED ISSUE #378: [FEATURE] Add lm-eval to our Metrics
> ‼️ If you want to work on this issue: please comment below and wait until a maintainer assigns this issue to you before opening a PR to avoid several contributions on the same issue. Thanks! 😊

## **✨ What You’ll Do**

Adding more metrics makes Pruna’s evaluations more insightful, so we’re excited to welcome **[evaluation harness](https://github.com/EleutherAI/lm-evaluation-harness)** into our metric lineup! 🎉

---

## **📐 Metric Class**

For a detailed guide, refer to the documentation on implementing a metric: https://docs.pruna.ai/en/stable/docs_pruna/user_manual/customize_metric.html.
Also, take a look at the metric code repository here: https://github.com/EleutherAI/lm-evaluation-harness
**Quick Overview**

1. **File & Class**
    - Create the new file and define the class.
2. **Configuration**
    - Choose a sensible `call_type` 
    - Define the state and update functions.
    - Don't forget to register the metric
3. **Testing & Docs**
    - Write the unit tests.
 
And that's all!

## **✅ Acceptance Criteria**

- **Correct Output:** `compute(...)` returns a sensible result 
- **Tests & Docs:** All existing and new unit tests pass, and the documentation is updated

---

## **❓ Questions?**

Feel free to jump into our Discord channel if you hit any roadblocks. Can’t wait to see your contribution of this metric! 🚀

---

## Share on X

sky-2002:
## Description
Adds [evaluation harness](https://github.com/EleutherAI/lm-evaluation-harness) to the metrics lineup.

## Related Issue

Fixes #378 

## Type of Change

## How Has This Been Tested?
- Have added basic unit tests for two tasks from eval harness.
- Will add more tests as needed

## Checklist

## Additional Notes
- `EvalHarnessMetric` inherits from `BaseMetric` since eval harness metrics are model level metrics.
- Open to discuss and update code and tests.

### DIFF
--- a/src/pruna/evaluation/metrics/metric_evalharness.py
+++ b/src/pruna/evaluation/metrics/metric_evalharness.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+from typing import Any, List, Dict, Union
+
+import torch
+from pruna.evaluation.metrics.metric_base import BaseMetric
+from pruna.evaluation.metrics.registry import MetricRegistry
+from pruna.evaluation.metrics.result import MetricResult
+from pruna.logging.logger import pruna_logger
+
+
+METRIC_EVALHARNESS = "eval_harness"
+
+
+@MetricRegistry.register(METRIC_EVALHARNESS)
+class EvalHarnessMetric(BaseMetric):
+    """
+    Wraps the EleutherAI LM Evaluation Harness as a Pruna metric.
+
+    This allows you to run benchmark tasks (HellaSwag, MMLU, GSM8k, etc.)
+    and return a scalar score usable in Pruna’s pipeline.
+
+    Parameters
+    ----------
+    tasks : list[str]
+        List of eval harness tasks to run.
+    model_args : dict
+        Arguments for the backend model in eval harness.
+    device : str | torch.device
+        Device to run evaluation on (cuda, cpu, etc.).
+    call_type : str
+        The call type (default "y").
+    """
+
+    metric_name: str = METRIC_EVALHARNESS
+    default_call_type: str = "y"
+    higher_is_better: bool = True
+    metric_units: str = "score"
+
+    def __init__(
+        self,
+        tasks: List[str],
+        model_args: Dict[str, Any],
+        device: Union[str, torch.device] = "cuda:0",
+        call_type: str = "y",
+        **kwargs,
+    ):
+        super().__init__()
+        self.tasks = tasks
+        self.model_args = model_args
+        self.device = torch.device(device) if isinstance(device, str) else device
+        self.call_type = call_type
+
+    def compute(self, model, dataloader) -> MetricResult:
+        """
+        Run LM Eval Harness and return the aggregated result.
+        """
+        try:
+            from lm_eval import evaluator
+        except ImportError:
+            raise ImportError(
+                "lm_eval package is not installed. Please install it with `pip install lm-eval`."
+            )
+
+        try:
+            eval_results = evaluator.simple_evaluate(
+                model="hf",
+                model_args=self.model_args,
+                tasks=self.tasks,
+                device=str(self.device),
+                batch_size=1,
+            )
+        except Exception as e:
+            pruna_logger.error(f"Eval Harness failed: {e}")
+            raise
+
+        scores, task_scores = [], {}
+
+        preferred_keys = ["acc,none", "acc", "f1", "em"]
+
+        for task in self.tasks:
+            if task not in eval_results["results"]:
+                pruna_logger.warning(f"Task {task} not found in eval results")
+                continue
+
+            task_results = eval_results["results"][task]
+            metric_keys = [
+                k
+                for k in task_results.keys()
+                if k not in {"alias", "samples", "higher_is_better"}
+            ]
+
+            if not metric_keys:
+                pruna_logger.warning(f"No metrics found for task {task}")
+                continue
+
+            key = next((k for k in preferred_keys if k in metric_keys), metric_keys[0])
+            score = task_results[key]
+            task_scores[task] = score
+            scores.append(score)
+
+        final_score = sum(scores) / len(scores) if scores else 0.0
+
+        return MetricResult(
+            self.metric_name,
+            {**self.__dict__, "task_scores": task_scores},
+            final_score,
+            higher_is_better=self.higher_is_better,
+            metric_units=self.metric_units,
+        )
--- a/tests/evaluation/test_evalharness_metrics.py
+++ b/tests/evaluation/test_evalharness_metrics.py
@@ -1,0 +1,65 @@
+import pytest
+import torch
+
+from pruna.evaluation.metrics.metric_evalharness import EvalHarnessMetric
+from pruna.evaluation.metrics.result import MetricResult
+
+
+@pytest.mark.cpu
+def test_eval_harness_basic():
+    """Test EvalHarnessMetric on a tiny HF model + task."""
+
+    # Use a very small HF model for speed
+    model_args = {
+        "pretrained": "sshleifer/tiny-gpt2",
+        "revision": "main",
+    }
+
+    metric = EvalHarnessMetric(
+        tasks=["lambada_openai"],  # lightweight benchmark
+        model_args=model_args,
+        device=torch.device("cpu"),
+    )
+
+    # compute ignores Pruna dataloader/model, so pass None
+    result: MetricResult = metric.compute(model=None, dataloader=None)
+
+    print(f"Result: {result}")
+
+    # Assertions
+    assert isinstance(result, MetricResult)
+    assert "task_scores" in result.params
+    assert "lambada_openai" in result.params["task_scores"]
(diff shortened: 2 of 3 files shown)


sky-2002:
@begumcig Yeah I agree, while implementing as well, I was not very comfortable with this approach. 
From what I see in pruna code, a task has a request, which is a list of metrics, exactly as an `lm-eval` task, which has metrics defined in the yaml files. 

> alternatively, we could integrate some metrics from lm-eval metrics registry as StatefulMetrics.

Yes, this is a much cleaner way and aligns with `lm-eval` interface(to add each metric as a metric in pruna), let us do this. Will update code. 
Also, you are right, I misunderstood `StatefulMetric`

> I was thinking maybe we could integrate one task from lm-eval as one Task (inheriting from our Task interface) 

The above two need to be done together no? we need to implement a task, and define its metrics or we need a way to create a task on the fly given a task name and the metrics needed for it. 

For example, given that we have all metrics implemented in pruna:

We can instantiate a task for a lm-eval task:
```python
from pruna.evaluation.task import Task

squad_metrics = ["lm_eval_exact_match", "lm_eval_f1"]

squad_task = Task(
    request=squad_metrics,
    datamodule=squad_datamodule, 
    device="cuda"
)
```

or have a task factory:
```python
def make_pruna_task_from_lmeval(task_name, datamodule, model_args, device="cuda"):
    lm_task = lm_tasks.get_task(task_name)
    metrics = [f"lm_eval_{m['metric']}" for m in lm_task.config.metric_list]
    return Task(request=metrics, datamodule=datamodule, device=device)
```

Wdyt, how should we go about this?

begumcig:
@sky-2002 Yes, that's exactly how our task and stateful metric interface work! 

I was suggesting that you could:
- Either create something like a` LMEvalTask` inheriting from our `Task`, that would also integrate the attributes and functions that are required for lm_eval task. 
- Or exactly like you suggested integrate the metrics as stateful metrics and initialize them using our own Task interface. I think the solution you suggested aligns perfectly with our interface and it would be amazing! 

Once we have the metrics as Pruna metrics, we can use them with our Task interface like you have shown in the example! This all looks super good to me, and thank you so much for taking the time to look into this, you are a champ 🏆🏆🏆🏆

sky-2002:
@begumcig  I tried implementing. But I have a few concerns. 
- We can create wrapper around `lm-eval` metrics, but then we have to do it around everything, for example, for a task say `hellaswag`, we need its dataset (specified in yaml), its pre-processing and post-processing functions etc. which might make our code brittle. Might get confusing and maybe too much duplicate code too, because it is already there in `lm-eval`.

> Either create something like a LMEvalTask inheriting from our Task, that would also integrate the attributes and functions that are required for lm_eval task.

I realise that this would be much cleaner and would leave `lm-eval` dependent code to itself, we don't need to keep updating it. But also, idk whether it should inherit from existing `Task` code, feel free to give suggestions, I am a little confused about what would be a good way to do this.

> Or exactly like you suggested integrate the metrics as stateful metrics and initialize them using our own Task interface. I think the solution you suggested aligns perfectly with our interface and it would be amazing!

Or if we continue this way, how do we handle the datasets for each task and other related attributes and functions?

src/pruna/evaluation/metrics/metric_evalharness.py
```diff
@@ -0,0 +1,121 @@
...
+                    "num_samples": 0,
+                    "higher_is_better": self.higher_is_better,
+                    "metric_units": self.metric_units,
+                },
+                result=0.0,
+            )
+
+        try:
+            # lm-eval metrics expect a list of (reference, prediction) tuples
+            raw_items = self.metric_fn(self.pairs)
```
begumcig: I love this new version! Being super nitpicky here, but I think it makes sense to move line 105 (self.metric_fn()) to update, so the pairs collect the calculated metric scores and in compute we only do the final aggregation?

src/pruna/evaluation/metrics/metric_evalharness.py
```diff
@@ -0,0 +1,121 @@
...
+            score = self.agg_fn(raw_items)
+        except Exception as e:
+            pruna_logger.error(f"Failed computing lm-eval metric {self.metric_name}: {e}")
+            raise
+
+        score_value = float(np.mean(list(score.values()))) if isinstance(score, dict) else float(score)
```
begumcig: I am asking just to learn, if we do aggregation already on line 106 why do we take the mean here?

src/pruna/evaluation/task.py
```diff
@@ -255,6 +263,11 @@ def _process_single_request(
             TorchMetricWrapper("clip_score", call_type="pairwise", device=stateful_metric_device),
             CMMD(device=stateful_metric_device),
         ]
+
+    elif request.startswith("lm_eval:"):
```
begumcig, line 333: This is insanely cool!

tests/evaluation/test_evalharness_metrics.py
```diff
@@ -0,0 +1,46 @@
+import pytest
```
begumcig, line 1: I love this test suite, could we possibly also add some tests for the "lm_eval:" case, using Task?

begumcig:
> @begumcig I tried implementing. But I have a few concerns.
> 
> * We can create wrapper around `lm-eval` metrics, but then we have to do it around everything, for example, for a task say `hellaswag`, we need its dataset (specified in yaml), its pre-processing and post-processing functions etc. which might make our code brittle. Might get confusing and maybe too much duplicate code too, because it is already there in `lm-eval`.
> 
> > Either create something like a LMEvalTask inheriting from our Task, that would also integrate the attributes and functions that are required for lm_eval task.
> 
> I realise that this would be much cleaner and would leave `lm-eval` dependent code to itself, we don't need to keep updating it. But also, idk whether it should inherit from existing `Task` code, feel free to give suggestions, I am a little confused about what would be a good way to do this.
> 
> > Or exactly like you suggested integrate the metrics as stateful metrics and initialize them using our own Task interface. I think the solution you suggested aligns perfectly with our interface and it would be amazing!
> 
> Or if we continue this way, how do we handle the datasets for each task and other related attributes and functions?

@sky-2002 That makes a lot of sense, I totally agree that bringing dataset handling into the evaluation part would make things brittle and messy.
I think we could handle that through the PrunaDataModule instead? We could even consider adding the required data as datasets directly in our library, so the evaluation code stays clean and focused on its purpose and people could use these datasets for any purpose they prefer :)

src/pruna/evaluation/metrics/metric_evalharness.py
sky-2002: Done

src/pruna/evaluation/metrics/metric_evalharness.py
sky-2002: Oh yeah, i don't think we need the mean, removed it

src/pruna/evaluation/task.py
sky-2002: thanks 😄

tests/evaluation/test_evalharness_metrics.py
sky-2002: Done

sky-2002:
@begumcig I have tried to address review comments, thanks for the review. 

~~Also I noticed that pruna's code structured changed a little bit (or is it just in algorithms), what changes will we need in this PR?~~

Edit: The code structure changes were in algorithms, ignore above question

davidberenstein1957:
@begumcig @sky-2002 any blockers here from either side?

tests/evaluation/test_evalharness_metrics.py
```diff
@@ -0,0 +1,104 @@
...
+    preds = ["the cat is on mat", "a quick brown fox"]
+
+    metric = LMEvalMetric(metric_name="bleu")
+    metric.update(preds, refs)
+    result = metric.compute()
+
+    assert isinstance(result, MetricResult)
+    assert "num_samples" in result.params
+    assert result.params["num_samples"] == 2
+    assert isinstance(result.result, float)
```
begumcig, line 27: Should we check the result is 1, rather than just checking for float? Since the example string are the exact same, we should get a perfect score, no?

sky-2002:
Hey @begumcig , sorry you had to push so many more commits, I sort of lost track of this PR, got busy with work. Will check that comment(if that is still relevant) when I am free.

begumcig:
@sky-2002 No worries at all! I was just making a few small updates. This looks really amazing, and thank you so much for this contribution! We haven’t had the chance to integrate many text evaluation metrics into Pruna yet, and this is a big step forward. Really appreciate the work you put into this 🥹

begumcig:
Also @sky-2002, while taking a look at the evaluation harness I noticed that the get_task_dict interface used here has been deprecated. If you’d be interested in updating it at some point, that could be a nice improvement, but absolutely no rush, we can also handle it in a future pass 💜

PR #383, opened 2025-10-02
feat: add pre-commit `trufflehog`

Almonok:
## Description

This PR adds trufflehog to the pre-commit hooks to prevent API keys/tokens leaks. 

## Type of Change

## Checklist

---

> [!NOTE]
> Add TruffleHog secret scanning to pre-commit/pre-push hooks and document its installation in CONTRIBUTING.
> 
> - **Tooling: pre-commit hooks**
>   - Add `trufflehog` hook in `.pre-commit-config.yaml` with stages `pre-commit` and `pre-push`, running `trufflehog git file://. --since-commit HEAD --results=verified,unknown --fail`.
> - **Docs**
>   - Update `CONTRIBUTING.md` to note TruffleHog is included in hooks and must be installed separately, with link to installation instructions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0b2a050eefd4d674db3470dc48916e7e19b18f5e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,7 +153,6 @@ dev = [
     "blobfile",
     "pillow>=9.5.0",
     "pre-commit",
-    "trufflehog",
     "twine",
     "pyc-wheel",
     "ruff",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -153,6 +153,7 @@ dev = [
     "blobfile",
     "pillow>=9.5.0",
     "pre-commit",
+    "trufflehog",
     "twine",
     "pyc-wheel",
     "ruff",
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,5 +51,5 @@ repos:
         name: TruffleHog
         description: Detect secrets in your data.
         entry: bash -c 'trufflehog git file://. --since-commit HEAD --results=verified,unknown --fail'
-        language: python
+        language: system
         stages: ["pre-commit", "pre-push"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,3 +47,9 @@ repos:
         stages: [pre-commit]
         types: [python]
         exclude: "^docs/"
+      - id: trufflehog
+        name: TruffleHog
+        description: Detect secrets in your data.
+        entry: bash -c 'trufflehog git file://. --since-commit HEAD --results=verified,unknown --fail'
+        language: python
+        stages: ["pre-commit", "pre-push"]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,6 +71,8 @@ You can then also install the pre-commit hooks with:
 pre-commit install
 ```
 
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -153,6 +153,7 @@ dev = [
     "blobfile",
     "pillow>=9.5.0",
     "pre-commit",
+    "trufflehog",
```
johannaSommer: Are we sure this is the correct package? It seems this package on pypi is not maintained anymore and seems to not be associated with https://github.com/trufflesecurity/trufflehog/blob/main/PreCommit.md. In any case this package should be installed when doing `pre-commit install` so maybe no need to put it into the dependencies?

johannaSommer (changes requested):

johannaSommer:
Thanks a lot @Almonok for tackling this, it's a great idea to add this! Just had one question about the pypi package we are using otherwise it already looks good to me! 😊

pyproject.toml
Almonok: Thank you for the review ! You're absolutely right, the PyPI package is not the correct one and is unmaintained.  I'll remove it from the dependencies. 

Unfortunately, it seems that we do not install the trufflehog binary when running pre-commit install; it requires an extra step that I can document in https://github.com/PrunaAI/pruna/blob/main/CONTRIBUTING.md. Would that be okay?

johannaSommer (approved):
Looks great to me, thanks a lot for this suggestion @Almonok

GreatBahram:
@Almonok Trufflehog package is a bit hard to install I wonder if you also considered [gitleaks](https://github.com/gitleaks/gitleaks) during your experiments? It seems it can be installed via [pre-commit hooks](https://github.com/gitleaks/gitleaks?tab=readme-ov-file#pre-commit) too, wdyt?

Almonok:
Hey @GreatBahram, sorry to hear that TruffleHog is causing some issues. The installation on macOS via Homebrew is pretty straightforward. What specific issues are you encountering, is it the installation or running the hook ?
I didn't test gitleaks but both tools seem comparable for the use case of this repo. If TruffleHog's setup is problematic, switching to Gitleaks could be reasonable as it appears to be lighter and easier to use but I don't have strong opinions on that one !

GreatBahram:
> Hey @GreatBahram, sorry to hear that TruffleHog is causing some issues. The installation on macOS via Homebrew is pretty straightforward. What specific issues are you encountering, is it the installation or running the hook ?
> I didn't test gitleaks but both tools seem comparable for the use case of this repo. If TruffleHog's setup is problematic, switching to Gitleaks could be reasonable as it appears to be lighter and easier to use but I don't have strong opinions on that one !

Hey, no worries it’s all good! 🙂 I just mentioned that since Gitleaks doesn’t need a separate installation, which makes the developer experience a bit smoother. But it’s just an idea: maybe TruffleHog is actually better for this use case.

Feel free to give it a try yourself, and if you need any support, just ping me!

PR #395, opened 2025-10-06
feat: video models and seeding for inference

begumcig:
## Description

This PR introduces several updates to the inference handlers:

DiffusersHandler Enhancements:

- Added support for video models.
- Added support for return_dict=False, which skips wrapping outputs in a pipeline output object and instead returns a tuple directly.

Seeding Support Across Handlers:

- All inference handlers now support optional seeding.

For DiffusersHandler, seeding is crucial since it directly affects the generation process.
For other models, inference is usually deterministic, but we added seeding for reproducibility in cases where randomness is involved.

Users can explicitly set or unset the seed. By default, no seed is applied, leaving seeding control to the user.

## Type of Change

## How Has This Been Tested?

Unit tests were added in tests/test_handler.py to validate the new functionality:

- Verified that seeding and unseeding work as expected for both standard inference handlers and the DiffusersHandler.

- Added tests for image and video model outputs using lightweight/tiny models to ensure correctness and efficiency.

## Checklist

Now you can configure the seed like: 
```
input_text = "photo of a cute prune."
pipe = FluxPipeline.from_pretrained("katuni4ka/tiny-random-flux")
pruna_model = PrunaModel(pipe)
pruna_model.inference_handler.configure_seed("per_sample",global_seed=42)
output_image = pruna_model.run_inference(input_text)

### DIFF
--- a/src/pruna/engine/handler/handler_diffuser.py
+++ b/src/pruna/engine/handler/handler_diffuser.py
@@ -19,7 +19,7 @@ from typing import Any, Dict, List, Literal, Optional, Tuple
 
 import torch
 
-from pruna.engine.handler.handler_inference import InferenceHandler, validate_seed_strategy
+from pruna.engine.handler.handler_inference import InferenceHandler
 from pruna.logging.logger import pruna_logger
 
 
@@ -93,35 +93,27 @@ class DiffuserHandler(InferenceHandler):
             # Maybe the user is calling the pipeline with return_dict = False,
             # which then returns the generated image / video in a tuple
             generated = output[0]
-        return generated
+        return generated.float()
 
     def log_model_info(self) -> None:
         """Log information about the inference handler."""
         pruna_logger.info(
             "Detected diffusers model. Using DiffuserHandler.\n- The first element of the batch is passed as input.\n"
+            "Inference outputs are expected to have either have an `images` attribute or a `frames` attribute."
+            "Or be a tuple with the generated image / video as the first element."
         )
 
-    def configure_seed(self, seed_strategy: Literal["per_sample", "no_seed"], global_seed: int | None) -> None:
+    def set_seed(self, seed: int) -> None:
         """
-        Set the random seed according to the chosen strategy.
-
-        - If `seed_strategy="per_sample"`,the `global_seed` is used as a base to derive a different seed for each
-        sample. This ensures reproducibility while still producing variation across samples,
-        making it the preferred option for benchmarking.
-        - If `seed_strategy="no_seed"`, no seed is set internally.
-        The user is responsible for managing seeds if reproducibility is required.
+        Set the random seed for the current process.
 
         Parameters
         ----------
-        seed_strategy : Literal["per_sample", "no_seed"]
-            The seeding strategy to apply.
-        global_seed : int | None
-            The base seed value to use (if applicable).
+        seed : int
+            The seed to set.
         """
-        self.seed_strategy = seed_strategy
-        validate_seed_strategy(seed_strategy, global_seed)
-        if global_seed is not None:
-            self.global_seed = global_seed
-            self.model_args["generator"] = torch.Generator("cpu").manual_seed(global_seed)
-        else:
-            self.model_args["generator"] = None  # Remove the seed.
+        self.model_args["generator"] = torch.Generator("cpu").manual_seed(seed)
+
+    def remove_seed(self) -> None:
+        """Remove the seed from the current process."""
+        self.model_args["generator"] = None
--- a/src/pruna/engine/handler/handler_inference.py
+++ b/src/pruna/engine/handler/handler_inference.py
@@ -122,38 +122,36 @@ class InferenceHandler(ABC):
         validate_seed_strategy(seed_strategy, global_seed)
         if global_seed is not None:
             self.global_seed = global_seed
-            set_seed(global_seed)
+            self.set_seed(global_seed)
         else:
-            remove_seed()
+            self.remove_seed()
 
+    def set_seed(self, seed: int) -> None:
+        """
+        Set the random seed for the current process.
 
-def set_seed(seed: int) -> None:
-    """
-    Set the random seed for the current process.
+        Parameters
+        ----------
+        seed : int
+            The seed to set.
+        """
+        #  With the default handler, we can't assume anything about the model,
+        #  so we are setting the seed for all RNGs available.
+        random.seed(seed)
+        np.random.seed(seed)
+        torch.manual_seed(seed)
+        if torch.cuda.is_available():
+            torch.cuda.manual_seed_all(seed)
 
-    Parameters
-    ----------
-    seed : int
-        The seed to set.
-    """
-    #  With the default handler, we can't assume anything about the model,
-    #  so we are setting the seed for all RNGs available.
-    random.seed(seed)
-    np.random.seed(seed)
-    torch.manual_seed(seed)
-    if torch.cuda.is_available():
-        torch.cuda.manual_seed_all(seed)
-
-
-def remove_seed() -> None:
-    """Remove the seed from the current process."""
-    random.seed(None)
-    np.random.seed(None)
-    #  We can't really remove the seed from the PyTorch RNG, so we are reseeding with torch.seed().
-    #  torch.seed() creates a non-deterministic random number.
-    torch.manual_seed(torch.seed())
-    if torch.cuda.is_available():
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/pruna/engine/handler/handler_diffuser.py [hunk 1]
```diff
@@ -83,13 +88,44 @@ def process_output(self, output: Any) -> torch.Tensor:
...
+        if hasattr(output, "images"):
+            generated = output.images
+        #  For video models.
+        elif hasattr(output, "frames"):
+            generated = output.frames
+        else:
+            # Maybe the user is calling the pipeline with return_dict = False,
+            # which then directly returns the generated image / video.
+            generated = output
+        return generated
```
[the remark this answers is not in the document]
begumcig: We add to model_args output_type = pt which returns tensors :)

src/pruna/engine/pruna_model.py
```diff
@@ -110,6 +110,8 @@ def run_inference(self, batch: Any, device: torch.device | str | None = None) ->
             )
         inference_function = getattr(self, inference_function_name)
 
+        self.inference_handler.model_args = filter_load_kwargs(self.model.__call__, self.inference_handler.model_args)
```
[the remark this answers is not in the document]
begumcig: well duh that's the purpose

src/pruna/engine/handler/handler_inference.py
```diff
@@ -116,3 +118,77 @@ def set_correct_dtype(
...
+def remove_seed() -> None:
+    """Remove the seed from the current process."""
+    random.seed(None)
+    np.random.seed(None)
+    #  We can't really remove the seed from the PyTorch RNG, so we are reseeding with torch.seed().
+    #  torch.seed() creates a non-deterministic random number.
+    torch.manual_seed(torch.seed())
+    if torch.cuda.is_available():
+        torch.cuda.manual_seed_all(torch.seed())
+
```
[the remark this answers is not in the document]
begumcig: Unfortunately we cannot remove seeds in a reliable way from torch, so this is intended.

src/pruna/engine/handler/handler_diffuser.py
```diff
@@ -40,12 +35,18 @@ class DiffuserHandler(InferenceHandler):
...
+        seed_strategy: Literal["per_sample", "no_seed"] = "no_seed",
+        global_seed: int | None = None,
+    ) -> None:
         self.call_signature = call_signature
-        if model_args:
-            default_args.update(model_args)
-        self.model_args = default_args
+        self.model_args = model_args if model_args else {}
+        # We want the default output type to be pytorch tensors.
+        self.model_args["output_type"] = "pt"
```
davidberenstein1957, line 48: Is this intended to overwrite? Perhaps checking if it is present and overwriting only if it is not is better?

src/pruna/engine/handler/handler_diffuser.py
```diff
@@ -83,13 +84,44 @@ def process_output(self, output: Any) -> torch.Tensor:
...
+            generated = output[0]
+        return generated
 
     def log_model_info(self) -> None:
         """Log information about the inference handler."""
         pruna_logger.info(
-            "Detected diffusers model. Using DiffuserHandler with fixed seed.\n"
-            "- The first element of the batch is passed as input.\n"
-            "- The generated outputs are expected to have .images attribute."
+            "Detected diffusers model. Using DiffuserHandler.\n- The first element of the batch is passed as input.\n"
```
davidberenstein1957, line 101: perhaps we can add, we work with .image, .frames or a sinlge image?

src/pruna/engine/handler/handler_diffuser.py
```diff
@@ -83,13 +84,44 @@ def process_output(self, output: Any) -> torch.Tensor:
...
         )
+
+    def configure_seed(self, seed_strategy: Literal["per_sample", "no_seed"], global_seed: int | None) -> None:
```
davidberenstein1957: this seems to be a duplicate from the method in `handle_inference`. Do we need both?

src/pruna/engine/handler/handler_inference.py
```diff
@@ -118,3 +120,77 @@ def set_correct_dtype(
...
+        self.seed_strategy = seed_strategy
+        validate_seed_strategy(seed_strategy, global_seed)
+        if global_seed is not None:
+            self.global_seed = global_seed
+            set_seed(global_seed)
+        else:
+            remove_seed()
+
+
+def set_seed(seed: int) -> None:
```
davidberenstein1957: solid seed setting 😎

src/pruna/engine/pruna_model.py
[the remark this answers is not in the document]
davidberenstein1957: sassy 😂

tests/engine/test_handler.py
```diff
@@ -0,0 +1,114 @@
...
+    ("per_sample", None),
+    ("no_seed", 42),
+])
+def test_validate_seed_strategy_invalid(strategy, seed):
+    '''Test to see validate_seed_strategy raises an error for invalid strategies'''
+    with pytest.raises(ValueError):
+        validate_seed_strategy(strategy, seed)
+
+
+def test_set_seed_reproducibility():
```
davidberenstein1957, line 28: should we test the other methods too, so for numpy and such?

tests/engine/test_handler.py
```diff
@@ -0,0 +1,114 @@
...
+    move_to_device(model, device)
+    pruna_model = PrunaModel(model, smash_config=smash_config)
+    pruna_model.inference_handler.configure_seed("per_sample", global_seed=seed)
+    result = pruna_model.run_inference(input_text)
+
+    # Get the output from the pipeline.
+    pipe_output = model(input_text, output_type="pt", generator=torch.Generator("cpu").manual_seed(seed), return_dict=return_dict)
+    if output_attr != "none":
+        pipe_output = getattr(pipe_output, output_attr)
+        pipe_output = pipe_output[0]
```
davidberenstein1957, line 77: why were we doin this again? this is only a problem when using them in tests like this, but this is not the behavior we expect users to deal with, correct?

davidberenstein1957 (approved):
some small comments, but feel free to mege after resolving,

src/pruna/engine/handler/handler_diffuser.py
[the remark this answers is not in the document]
simlang: loving the passive aggressive :)

src/pruna/engine/handler/handler_diffuser.py
simlang: not exact duplicate - would it make sense to move the set and remove seed into the class and override there in the child.
This would remove the boilerplate here, and move the logic into those two functions. what do you think?

src/pruna/evaluation/metrics/metric_cmmd.py
```diff
@@ -100,6 +101,8 @@ def update(self, x: List[Any] | torch.Tensor, gt: torch.Tensor, outputs: torch.T
             The output images.
         """
         inputs = metric_data_processor(x, gt, outputs, self.call_type)
+        if inputs[1].dtype == torch.bfloat16:
+            inputs[1] = inputs[1].to(torch.float16)
```
simlang, line 104: i would just cast it to f32, tbh.
This could actually change the value of the input, because the precision is different between bf16 and f16

src/pruna/engine/handler/handler_diffuser.py
begumcig: Yes! In the previous implementation we already automatically transformed everything to Tensors using torchvision.transforms.toTensor (line 87), so this is me just refactoring it using the more "official" way

src/pruna/engine/handler/handler_diffuser.py
begumcig: Agreed!

src/pruna/evaluation/metrics/metric_cmmd.py
begumcig: Actually moving this entire type casting part to inference handler! And yes will use f32 tysm

tests/engine/test_handler.py
begumcig: Exactly, we are doing all this here because the pipe is not a PrunaModel. Indeed when we use PrunaModel and Evaluation Agent we don't have to deal with this at all. This is the exact feature I am testing here :D

PR #396, opened 2025-10-06
feat: add saving and loading handling for artifacts

nifleisch:
## Description
Some algorithms (like torch.compile) produce binary artifacts that must be saved/loaded differently from our usual saving and loading functions. We previously hard-coded torch.compile as a special case. With more algorithms adopting artifact workflows, we need a general, extensible mechanism. 

This PR introduces two registries `ARTIFACT_LOAD_FUNCTIONS` and `ARTIFACT_SAVE_FUNCTIONS` - these enumerate save/load functions that operate on artifacts.

## Type of Change

## How Has This Been Tested?
I tested feature for both torch.compile and a new algorithm I am currently working on by saving and loading smashed models.

## Checklist

### DIFF
--- a/src/pruna/engine/load.py
+++ b/src/pruna/engine/load.py
@@ -64,10 +64,11 @@ def load_pruna_model(model_path: str | Path, **kwargs) -> tuple[Any, SmashConfig
     if len(smash_config.load_fns) == 0:
         raise ValueError("Load function has not been set.")
 
-    # load torch artifacts if they exist
-    if LOAD_FUNCTIONS.torch_artifacts.name in smash_config.load_fns:
-        load_torch_artifacts(model_path, **kwargs)
-        smash_config.load_fns.remove(LOAD_FUNCTIONS.torch_artifacts.name)
+    artifact_load_fns = []
+    for load_fn in ARTIFACT_LOAD_FUNCTIONS:
+        if load_fn.name in smash_config.load_fns:
+            artifact_load_fns.append(load_fn)
+            smash_config.load_fns.remove(load_fn.name)
 
     if len(smash_config.load_fns) > 1:
         pruna_logger.error(f"Load functions not used: {smash_config.load_fns[1:]}")
@@ -77,6 +78,9 @@ def load_pruna_model(model_path: str | Path, **kwargs) -> tuple[Any, SmashConfig
     # check if there are any algorithms to reapply
     if any(algorithm is not None for algorithm in smash_config.reapply_after_load.values()):
         model = resmash_fn(model, smash_config)
+
+    for load_fn in artifact_load_fns:
+        load_fn(model_path, model, smash_config, **kwargs)
 
     return model, smash_config
 
@@ -433,7 +437,7 @@ def load_hqq(model_path: str | Path, smash_config: SmashConfig, **kwargs) -> Any
         return quantized_model
 
 
-def load_torch_artifacts(model_path: str | Path, **kwargs) -> None:
+def load_torch_artifacts(model_path: str | Path, model: Any, smash_config: SmashConfig, **kwargs) -> None:
     """
     Load a torch artifacts from the given model path.
 
@@ -618,3 +622,6 @@ def filter_load_kwargs(func: Callable, kwargs: dict) -> dict:
         pruna_logger.info(f"Discarded unused loading kwargs: {list(invalid_kwargs.keys())}")
 
     return valid_kwargs
+
+
+ARTIFACT_LOAD_FUNCTIONS = [LOAD_FUNCTIONS.torch_artifacts]
--- a/src/pruna/engine/save.py
+++ b/src/pruna/engine/save.py
@@ -59,9 +59,10 @@ def save_pruna_model(model: Any, model_path: str | Path, smash_config: SmashConf
     if not model_path.exists():
         model_path.mkdir(parents=True, exist_ok=True)
 
-    if SAVE_FUNCTIONS.torch_artifacts.name in smash_config.save_fns:
-        save_torch_artifacts(model, model_path, smash_config)
-        smash_config.save_fns.remove(SAVE_FUNCTIONS.torch_artifacts.name)
+    for artifact_save_fn in ARTIFACT_SAVE_FUNCTIONS:
+        if artifact_save_fn.name in smash_config.save_fns:
+            artifact_save_fn(model, model_path, smash_config)
+            smash_config.save_fns.remove(artifact_save_fn.name)
 
     # in the case of no specialized save functions, we use the model's original save function
     if len(smash_config.save_fns) == 0:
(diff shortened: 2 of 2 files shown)


src/pruna/engine/load.py
```diff
@@ -64,10 +64,11 @@ def load_pruna_model(model_path: str | Path, **kwargs) -> tuple[Any, SmashConfig
     if len(smash_config.load_fns) == 0:
         raise ValueError("Load function has not been set.")
 
-    # load torch artifacts if they exist
-    if LOAD_FUNCTIONS.torch_artifacts.name in smash_config.load_fns:
-        load_torch_artifacts(model_path, **kwargs)
-        smash_config.load_fns.remove(LOAD_FUNCTIONS.torch_artifacts.name)
+    artifact_load_fns = []
+    for load_fn in ARTIFACT_LOAD_FUNCTIONS:
```
johannaSommer, line 68: sorry this is the most nitpicky comment ever but what about
```python
for load_fn in smash_config.load_fns:
    artifact_load_fns.append(getattr(ARTIFACT_LOAD_FUNCTIONS, load_fn)
    smash_config.load_fns.remove(load_fn)
```
saves one line hehe and is maybe a bit easier to read

src/pruna/engine/save.py
```diff
@@ -59,9 +59,10 @@ def save_pruna_model(model: Any, model_path: str | Path, smash_config: SmashConf
     if not model_path.exists():
         model_path.mkdir(parents=True, exist_ok=True)
 
-    if SAVE_FUNCTIONS.torch_artifacts.name in smash_config.save_fns:
-        save_torch_artifacts(model, model_path, smash_config)
-        smash_config.save_fns.remove(SAVE_FUNCTIONS.torch_artifacts.name)
+    for artifact_save_fn in ARTIFACT_SAVE_FUNCTIONS:
```
johannaSommer, line 62: same comment as above pls dont hate me

src/pruna/engine/save.py
```diff
@@ -84,8 +85,10 @@ def save_pruna_model(model: Any, model_path: str | Path, smash_config: SmashConf
     # in the case of multiple, specialized save functions, we default to pickled
     else:
         pruna_logger.debug(f"Several save functions stacked: {smash_config.save_fns}, defaulting to pickled")
-        save_fn = SAVE_FUNCTIONS.pickled
-        smash_config.load_fns = [LOAD_FUNCTIONS.pickled.name]
+        for fn in LOAD_FUNCTIONS:
```
johannaSommer: can you quickly clarify what we are doing here? i am a bit lost

src/pruna/engine/load.py
```diff
@@ -433,7 +437,7 @@ def load_hqq(model_path: str | Path, smash_config: SmashConfig, **kwargs) -> Any
         return quantized_model
 
 
-def load_torch_artifacts(model_path: str | Path, **kwargs) -> None:
+def load_torch_artifacts(model_path: str | Path, model: Any, smash_config: SmashConfig, **kwargs) -> None:
```
begumcig, line 440: I am a bit lost here, probably not seeing something. Can you explain a bit why do we need model and smashConfig here but we didn't change anything in the function itself using these?

src/pruna/engine/save.py
begumcig: Same, why are we checking if a load function is in save functions 🥺

src/pruna/engine/load.py
nifleisch: Good call, I should’ve added more context

This PR was done with a new algorithm in Pruna Pro (https://github.com/PrunaAI/prunatree/pull/237) in mind. Its artifact loader needs the model as it loads the saved artifact into a model component. Several existing load functions also read from smash_config. Because we invoke all loaders like:
```
for load_fn in artifact_load_fns:
    load_fn(model_path, model, smash_config, **kwargs)
```
I standardized the loader interface so every load_fn accepts (model_path, model, smash_config, **kwargs). That’s why load_torch_artifacts now takes model and smash_config even though it doesn’t currently use them.

src/pruna/engine/save.py
nifleisch: Thanks for catching this brain fart of mine. 🙏

I think you are there is something off here. What I was trying to do was the following:
- We have logic that, if we have more than one save_fn, we default to pickled. When we do that, we also replace all loading functions with the single pickled loader.
- The problem is that this also replaces all artifact loading functions with the pickled loader. But even if we pickle the model, we might still want to load the artifacts separately (e.g., the torch artifacts).
- This is why I introduced new code to avoid this: `smash_config.load_fns = [LOAD_FUNCTIONS.pickled.name]`, which would have removed the already added artifact_load_fns.

However, as I said, I don’t think the code I introduced does the job properly.
- `if fn not in ARTIFACT_SAVE_FUNCTIONS:` this is certainly a bug

I was thinking if we even need this line here
```
smash_config.load_fns = [LOAD_FUNCTIONS.pickled.name]
```

Shouldn't be the the load function added to the smash config when calling the pickle save function:

```
def save_pickled(model: Any, model_path: str | Path, smash_config: SmashConfig) -> None:
    """
    Save the model by pickling it.

    Parameters
    ----------
    model : Any
        The model to save.
    model_path : str | Path
        The directory to save the model to.
    smash_config : SmashConfig
        The SmashConfig object containing the save and load functions.
    """
    # helpers can not be pickled, we will disable and just reapply them later
    smash_helpers = get_helpers(model)
    for helper in smash_helpers:
        getattr(model, helper).disable()
    torch.save(model, Path(model_path) / PICKLED_FILE_NAME)
    smash_config.load_fns.append(LOAD_FUNCTIONS.pickled.name)
```
If this does apply and we can remove `smash_config.load_fns = [LOAD_FUNCTIONS.pickled.name]`, I think this should already solve it. But maybe I am missing something here.

src/pruna/engine/load.py
nifleisch: Always happy to save lines of code! ☺️

However I’m not sure this code snippet is equivalent to my implementation. I remove all artifact load functions from smash_config.load_fns and store them in a separate list. This ensures artifacts are loaded after the model, which is necessary for artifacts that must be loaded into components of the model (see PR https://github.com/PrunaAI/prunatree/pull/237 for an example).

Regarding your snippet I am unsure how:
- `getattr` works for the list ARTIFACT_LOAD_FUNCTIONS. Are the elements of this list considered attributes.
- Isn't it removing all load functions from the smash_config

But maybe I am missing something.

src/pruna/engine/save.py
nifleisch: See comment above.

nifleisch:
Thank you for your reviews! You raised good points. In hindsight, I think I really should have given more context to make it easier to understand this PR. I did a small fix and would be super thankful for you to give it another pass.

PR #400, opened 2025-10-08
feat: evalution agent for vbench and artifact saver

begumcig:
## Description

This PR introduces artifact saving and related metric handling capabilities, and evaluation agent workflow updates.

Artifact Saver

- Added an abstract ArtifactSaver class and a VideoArtifactSaver implementation for storing inference outputs.
- Introduced a create_alias method, which allows creating hardlinks or symlinks to saved artifacts under different filenames. This helps reduce memory usage.

Metrics & Modalities

- Added a modality attribute to the metrics module. Each quality metric now specifies its modality.
- Metrics that work across modalities are marked as general.

A task may only have one modality (other than general). This ensures consistency since the Evaluation Agent operates on one task with one model, and their modalities must match.

Evaluation Agent

- Can now be initialized with a seeding strategy (Related to https://github.com/PrunaAI/pruna/pull/395 )
- Accepts artifact-saving arguments directly, which are then forwarded to the model’s inference handler.
- Can run inference many times per input, as requested by the user.
- Saves artifacts for each input if requested.
- Creates aliases for none/one/some/all stateful metrics, depending on user configuration.

## Type of Change

## How Has This Been Tested?

This PR also adds comprehensive tests to ensure correctness of the new features:

FlashAttention3 Algorithm Test
- Runs one of the new VBench metrics with the Evaluation Agent to validate integration and correctness with the updated metrics system.

Task Modality Tests

- Verifies that each modality is supported and behaves as expected.
- Ensures that mixing incompatible modalities raises an error.

Evaluation Agent Tests

- Confirms that artifact saving is handled correctly within the agent.

Artifact Saving & Alias Tests

- Unit tests validate artifact saving and alias creation across all supported export formats (NumPy, Torch, and PIL video outputs).

## Checklist

---

> [!NOTE]
> Introduces an artifact saver (with video support) and extends the EvaluationAgent to save per-sample outputs with seeding/multi-sample control, while making metrics/task modality-aware and adding tests.
> 
> - **Evaluation**:
>   - Enhance `EvaluationAgent` to support saving artifacts (`save_artifacts`, `root_dir`, `artifact_saver_export_format`, `saving_kwargs`), multi-sample inference (`num_samples_per_input`), and seeding (`seed_strategy`, `global_seed`).
>   - Caches outputs for pairwise metrics; sets model eval mode and seeds via `inference_handler`.
> - **Artifact Saving**:
>   - Add `ArtifactSaver` base (`src/pruna/evaluation/artifactsavers/artifactsaver.py`) with `create_alias` (hardlink→symlink fallback).
>   - Implement `VideoArtifactSaver` with MP4/GIF export; conversion from torch/np/PIL; utility `assign_artifact_saver`.
> - **Metrics & Modalities**:
>   - Introduce modality constants and per-metric `modality` attributes; propagate through `CMMD`, `PairwiseClipScore`, `SharpnessMetric`, and `TorchMetrics`/`TorchMetricWrapper`.
>   - Add filename aliasing hook `StatefulMetric.create_filename`.
> - **Task**:
>   - Validate single-modality across stateful metrics and expose `Task.modality`.
> - **Tests**:
>   - New tests for artifact saving/aliasing, EvaluationAgent artifact flow, and task modality validation; kernel tester metric updated to `background_consistency`.
> - **Misc**:
>   - Use `get_device` when wrapping base models; minor config tweak in `pyproject.toml` lint rules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f6b2df39f2ce89e34679c95c9dedb5a1712e8397. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/pruna/evaluation/artifactsavers/video_artifactsaver.py
+++ b/src/pruna/evaluation/artifactsavers/video_artifactsaver.py
@@ -55,6 +55,8 @@ class VideoArtifactSaver(ArtifactSaver):
         ----------
         data: Any
             The data to save.
+        saving_kwargs: dict
+            The additional kwargs to pass to the saving utility function.
 
         Returns
         -------
--- a/src/pruna/evaluation/evaluation_agent.py
+++ b/src/pruna/evaluation/evaluation_agent.py
@@ -264,11 +264,13 @@ class EvaluationAgent:
                 for stateful_metric in single_stateful_metrics:
                     stateful_metric.update(x, gt, processed_outputs)
                     if self.save_artifacts and stateful_metric.create_alias:
-                        # Again, we have to create an alias for each sample in the batch.
+                        # The evaluation agent saves the artifacts with a canonical filenaming convention.
+                        # If the user wants to save the artifact with a different filename,
+                        # here we give them the option to create an alias for the file.
                         for prompt_idx, prompt in enumerate(x):
                             if self.artifact_saver.export_format is None:
                                 raise ValueError(
-                                    "Export format is not set.Please set the export format for the artifact saver."
+                                    "Export format is not set. Please set the export format for the artifact saver."
                                 )
                             alias_filename = stateful_metric.create_filename(
                                 filename=prompt, idx=sample_idx, file_extension=self.artifact_saver.export_format
--- a/src/pruna/evaluation/task.py
+++ b/src/pruna/evaluation/task.py
@@ -168,6 +168,8 @@ def _safe_build_metrics(
         str
             The modality of the task.
         """
+        if not self.get_single_stateful_metrics() and not self.get_pairwise_stateful_metrics():
+            return "general"
         modality_intersection = set.intersection(
             *[metric.modality for metric in self.metrics if isinstance(metric, StatefulMetric)]
         )
--- a/tests/evaluation/test_artifactsaver.py
+++ b/tests/evaluation/test_artifactsaver.py
@@ -39,7 +39,7 @@ def test_assign_artifact_saver_video(tmp_path: Path):
 def test_assign_artifact_saver_invalid():
     """ Test that we raise an error if the artifact saver is assigned incorrectly."""
     with pytest.raises(ValueError):
-        assign_artifact_saver("text")
+        assign_artifact_saver("nonexistent_modality")
 
 @pytest.mark.parametrize(
     "export_format, save_from_type, save_from_dtype",


src/pruna/evaluation/evaluation_agent.py
```diff
@@ -202,22 +233,37 @@ def update_stateful_metrics(
...
+                for stateful_metric in single_stateful_metrics:
+                    stateful_metric.update(x, gt, processed_outputs)
+                    if stateful_metric.create_alias:
+                        # Again, we have to create an alias for each sample in the batch.
+                        for prompt_idx, prompt in enumerate(x):
+                            assert isinstance(self.artifact_saver.export_format, str)
+                            alias_filename = stateful_metric.create_filename(
+                                filename=prompt, idx=sample_idx, file_extension=self.artifact_saver.export_format
+                            )
+                            self.artifact_saver.create_alias(canonical_paths[prompt_idx], alias_filename)
```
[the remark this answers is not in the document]
begumcig: I am confused because in line 251 we have if `self.save_artifacts`

src/pruna/evaluation/artifactsavers/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        The root directory to save the artifacts.
+    export_format: str
+        The format to save the artifacts.
+
+    Returns
+    -------
+    ArtifactSaver
+        The appropriate artifact saver.
+    """
+    if modality != "video":
```
davidberenstein1957: this will be removed later?

src/pruna/evaluation/artifactsavers/artifactsaver.py
```diff
@@ -0,0 +1,98 @@
...
+            The data to save.
+
+        Returns
+        -------
+        Path
+            The full path to the saved artifact.
+        """
+        pass
+
+    def create_alias(self, source_path: Path | str, filename: str) -> Path:
```
simlang, line 58: cool idea!

src/pruna/evaluation/artifactsavers/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        raise ValueError(f"Modality {modality} is not supported")
+    else:
+        return VideoArtifactSaver(root=root, export_format=export_format)
```
simlang: very nit-picky, but i would do it like this, so when we add new artifact savers for other modalities, it's just adding elifs on the modality and the error is the fallback

```suggestion
    if modality == "video":
        return VideoArtifactSaver(root=root, export_format=export_format)
    else:
        raise ValueError(f"Modality {modality} is not supported")
```

src/pruna/evaluation/metrics/metric_stateful.py
```diff
@@ -35,6 +35,8 @@ class StatefulMetric(ABC):
 
     metric_name: str
     call_type: str
+    create_alias: bool = False
+    modality: List[str]
```
simlang: can a metric have multiple modalities?

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -70,12 +92,20 @@ def __init__(
...
+        self.seed_strategy = seed_strategy
+        self.num_samples_per_input = num_samples_per_input
+        self.global_seed = global_seed
         self.device = set_to_best_available_device(self.task.device)
         self.cache: List[Tensor] = []
         self.evaluation_for_first_model: bool = True
+        self.save_artifacts: bool = save_artifacts
+        if save_artifacts:
+            self.root_dir = root_dir if root_dir is not None else tempfile.mkdtemp(prefix="inference_outputs")
+            self.artifact_saver = assign_artifact_saver(self.task.modality, self.root_dir, artifact_saver_export_format)
```
simlang: assign_aritficat_saver expects a single string, but here a list is passed - relating to the previous comment, is it actually a list, or should it be changed to a single string anyway

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -202,22 +233,39 @@ def update_stateful_metrics(
...
+                        canonical_paths.append(canonical_path)
+
+                batch = move_batch_to_device(batch, self.device)
+                processed_outputs = move_batch_to_device(processed_outputs, self.device)
+                (x, gt) = batch
+                # Non-pairwise (aka single) metrics have regular update.
+                for stateful_metric in single_stateful_metrics:
+                    stateful_metric.update(x, gt, processed_outputs)
+                    if self.save_artifacts and stateful_metric.create_alias:
+                        # Again, we have to create an alias for each sample in the batch.
```
simlang: why do we do this?

src/pruna/evaluation/metrics/metric_stateful.py
davidberenstein1957: @begumcig could you take a look here?

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -170,8 +200,9 @@ def prepare_model(self, model: Any) -> PrunaModel:
                 model.smash_config.batch_size,
             )
 
-        # ensure the model is on the cpu
-        model.move_to_device("cpu")
+        model.set_to_eval()
```
davidberenstein1957, line 218: does this capture setting the model to CPU, might still be good, since we are done evaluating, right?

src/pruna/evaluation/evaluation_agent.py
davidberenstein1957: @begumcig :)

src/pruna/evaluation/task.py
```diff
@@ -109,6 +110,32 @@ def is_pairwise_evaluation(self) -> bool:
...
+        modalities_no_general = {
+            modality for metric in stateful_metrics for modality in metric.modality if modality != "general"
+        }
+        #  We should also allow 0 because the user might have only general modality metrics.
+        if len(modalities_no_general) == 1:
+            modality = modalities_no_general.pop()
+        elif len(modalities_no_general) == 0:
+            modality = "general"
+        else:
+            raise ValueError("The task should have a single modality of quality metrics.")
```
davidberenstein1957:
```suggestion
            raise ValueError("The task should have a single modality across quality metrics.")
```

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
```diff
@@ -0,0 +1,80 @@
...
+        Returns
+        -------
+        Path
+            The path to the saved artifact.
+        """
+        canonical_filename = f"{int(time.time())}_{secrets.token_hex(4)}.{self.export_format}"
+        canonical_path = Path(str(self.root)) / "canonical" / canonical_filename
+
+        #  all diffusers saving utility functions accept a list of PIL.Images, so we convert to PIL to be safe.
+        if isinstance(data, torch.Tensor):
```
davidberenstein1957, line 70: Is there a specific reason we are converting to uint8?

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -70,12 +92,20 @@ def __init__(
...
         self.subsequent_model_results: List[MetricResult] = []
+        self.seed_strategy = seed_strategy
+        self.num_samples_per_input = num_samples_per_input
+        self.global_seed = global_seed
         self.device = set_to_best_available_device(self.task.device)
         self.cache: List[Tensor] = []
         self.evaluation_for_first_model: bool = True
+        self.save_artifacts: bool = save_artifacts
+        if save_artifacts:
+            self.root_dir = root_dir if root_dir is not None else tempfile.mkdtemp(prefix="inference_outputs")
```
davidberenstein1957: perhaps we can remove thix directory to the top of the file?

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -202,22 +233,39 @@ def update_stateful_metrics(
...
+                batch = move_batch_to_device(batch, self.device)
+                processed_outputs = move_batch_to_device(processed_outputs, self.device)
+                (x, gt) = batch
+                # Non-pairwise (aka single) metrics have regular update.
+                for stateful_metric in single_stateful_metrics:
+                    stateful_metric.update(x, gt, processed_outputs)
+                    if self.save_artifacts and stateful_metric.create_alias:
+                        # Again, we have to create an alias for each sample in the batch.
+                        for prompt_idx, prompt in enumerate(x):
+                            assert isinstance(self.artifact_saver.export_format, str)
```
davidberenstein1957: it is better to avoid using assert in production code for a bunch of reason, so a explicit raise is better here :) 

reasons : https://claude.ai/share/a2267a5a-5abf-439f-b755-bd3cc633de54

tests/evaluation/test_evalagent.py
```diff
@@ -0,0 +1,43 @@
...
+from pruna.engine.pruna_model import PrunaModel
+
+
+@pytest.mark.cuda
+@pytest.mark.parametrize("model_fixture",
+[pytest.param("wan_tiny_random", marks=pytest.mark.cuda)], indirect=["model_fixture"])
+def test_agent_saves_artifacts(model_fixture):
+    """ Test that the agent runs inference and saves the inference output artifacts correctly."""
+    model, smash_config = model_fixture
+    # Metrics don't work with bfloat16
```
davidberenstein1957: should we perhaps test for this somewhere in the code and warn for this during execution?

tests/evaluation/test_evalagent.py
```diff
@@ -0,0 +1,43 @@
...
+    dm.limit_datasets(data_points)
+
+
+    agent = EvaluationAgent(
+        request=['background_consistency'],
+        datamodule=dm,
+        device="cuda",
+        save_artifacts=True,
+        root_dir=temp_path,
+        artifact_saver_export_format="mp4",
```
davidberenstein1957: should we also check for gif?

tests/evaluation/test_evalagent.py
```diff
@@ -0,0 +1,43 @@
...
+    """ Test that the agent runs inference and saves the inference output artifacts correctly."""
+    model, smash_config = model_fixture
+    # Metrics don't work with bfloat16
+    model.to(dtype=torch.float16, device="cuda")
+    # Artifact path
+    temp_path = tempfile.mkdtemp()
+
+    # Let's limit the number of data points
+    dm  = smash_config.data
+    data_points = 2
```
davidberenstein1957, line 21: perhaps 1 is enough?

tests/evaluation/test_evalagent.py
```diff
@@ -0,0 +1,43 @@
...
+        datamodule=dm,
+        device="cuda",
+        save_artifacts=True,
+        root_dir=temp_path,
+        artifact_saver_export_format="mp4",
+        saving_kwargs={"fps":4}
+    )
+
+    pruna_model = PrunaModel(model, smash_config)
+    pruna_model.inference_handler.model_args["num_inference_steps"] = 2
```
davidberenstein1957: perhaps 1 is enough?

tests/evaluation/test_artifactsaver.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.mark.parametrize(
+    "export_format, save_from_type",
+    list(itertools.product(["gif", "mp4"], ["np", "pt", "pil"]))
+)
+def test_video_artifact_saver_tensor(export_format: str, save_from_type: str):
+    """ Test that we can save a video from numpy, torch and PIL in mp4 and gif formats."""
+    with tempfile.TemporaryDirectory() as tmp_path:
+        saver = VideoArtifactSaver(root=tmp_path, export_format=export_format)
+        # create a fake video:
+        if save_from_type == "pt":
```
davidberenstein1957, line 67: should we check for not using uint8?

src/pruna/evaluation/artifactsavers/utils.py
begumcig: yes, ideally the more artifact savers we have, the more modalities we will have here :)

src/pruna/evaluation/artifactsavers/utils.py
begumcig: makes so much sense

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
begumcig: https://github.com/huggingface/diffusers/blob/9c3b58dcf16ebd027fd3d85ec703ad5a142b1e1e/src/diffusers/utils/export_utils.py#L198 in diffusers utils they also convert each frame to uint8, I assumed it might be required by imageio or for the conversion from numpy to PIL. I am super open to changing it if you have suggestions or if you think this would introduce errors :)

src/pruna/evaluation/evaluation_agent.py
begumcig: We actually decided (@sharpenb can eloborate more I guess) that we don't want to move the model back to CPU after evaluating, but rather make sure that the model is on the same device before starting evaluation.

src/pruna/evaluation/evaluation_agent.py
begumcig: The evaluation agent saves the inference outputs with the artifact saver right after running inference and uses a "canonical" file format (which I defined to be a timestamp and a unique uuid). Here we give the user the option to create a hard/sym link to that "canonical" file under a different alias, incase they need the file with a certain filename (some benchmarks need the filename to be equal to the prompt for instance.) 

Obviously, if you are using Pruna you don't actually have to care about your filename at all (💅💅💅), since we take care of every step of the process from smashing, inference, post-processing and evaluation; but since we also introduced artifact savers in this PR, I also wanted to give the user to be able to save their files with a certain filenaming convention of their choosing, if they wish to do so. I thought it would be unncessarily strict to be able to only save the artifacts with "our" filename formatting, and it would take too much space on disk to save the same file multiple times. So here we just create an alias with a different name instead.

TLDR; I overengineered stuff, can remove if you think it doesn't make sense lmk <3

src/pruna/evaluation/evaluation_agent.py
begumcig: Great point!

tests/evaluation/test_evalagent.py
begumcig: yup completely agree!

tests/evaluation/test_evalagent.py
begumcig: possibly, but artifact saving do not support batching but evaluation agent does, so i wanted to test if passing outputs to the artifact saver one by one is done correctly

tests/evaluation/test_evalagent.py
begumcig: Yes, this is an insanely good point, I think I was a bit lazy to not handle this 😇, I have looked over our inference handlers for most of them we change the output dtype to be float, I think we can also apply this in the diffusers handler and log a message what do you think?

tests/evaluation/test_evalagent.py
begumcig: yes, definitely, we totally should!

tests/evaluation/test_evalagent.py
davidberenstein1957: I think that sounds good :)

src/pruna/evaluation/evaluation_agent.py
davidberenstein1957: Sure, but how would this work with comparing multiple models after on another? Wouldn't we likely be overloading memory? Or at least, we just assume this is a thing the user should think of themselves? Perhaps a flag could be nice?

src/pruna/evaluation/evaluation_agent.py
davidberenstein1957: @simlang an answer :)

src/pruna/evaluation/evaluation_agent.py
begumcig: Yes, you are absolutely correct about multiple models overcrowding the memory. I think it's 
- Either we just ensure consistency and the user gets their model on the same device they have started with 
-  Or we move things back to CPU after the agent is finished, so we possibly prevent some OOM issues, but if the user wanted to do more tasks with the model after evaluation, they need to move their model back.

So I think every point has its merits and shortcomings, and it's more about what we believe is more important. I initially also thought that moving the model back to the CPU was the better option (hence the original evaluation agent was implemented with that logic). Considering we now also support Multi-GPU and more complicated device settings, moving the user's model around without them explicitly asking for it might not be the best course of action going forward :)

tests/evaluation/test_artifactsaver.py
begumcig: Yes, added!

src/pruna/evaluation/metrics/metric_stateful.py
begumcig: This is a super good point. I think we either have a single modality in this attribute and for the metrics that support multiple modalities (like accuracy, precision, etc.) we have "general" to indicate that is a multimodality metric. Then the modality field can just be a string.
Or we explicitly state all possible modalities for each metric (incase something like (video image)(image text) happens), then we would have to carry them in an array and take some sort of intersection operation to validate / check the modality

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -202,22 +233,37 @@ def update_stateful_metrics(
...
+                            )
+                            self.artifact_saver.create_alias(canonical_paths[prompt_idx], alias_filename)
+
+                # Cache outputs once in the agent for pairwise metrics to save compute time and memory.
+                if self.task.is_pairwise_evaluation():
+                    if self.evaluation_for_first_model:
+                        self.cache.append(processed_outputs)
+                    else:
+                        for pairwise_metric in pairwise_metrics:
+                            pairwise_metric.update(x, self.cache[batch_idx], processed_outputs)
```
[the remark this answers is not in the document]
begumcig: if num_samples_per_input > 1, we raise an error

src/pruna/evaluation/artifactsavers/artifactsaver.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Create an alias for the artifact.
+
+        The evaluation agent will save the inference outputs with a canonical file
+        formatting style that makes sense for the general case.
+
+        If your metric requires a different file naming convention for evaluation,
```
simlang, line 65: i'm not sure if you changed the description here, but for some reason reviewing this today, makes it much much clearer. amazing!

src/pruna/evaluation/artifactsavers/artifactsaver.py
```diff
@@ -0,0 +1,98 @@
...
+            The path to the source artifact.
+        filename : str
+            The filename to create the alias for.
+
+        Returns
+        -------
+        Path
+            The full path to the alias.
+        """
+        alias = Path(str(self.root)) / f"{filename}.{self.export_format}"
```
simlang, line 85: i actually think that the Path(str(self.root)) is unnecessary , you should be able to just do Path(self.root)

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def save_artifact(self, data: Any, saving_kwargs: dict = dict()) -> Path:
+        """
+        Save the video artifact.
+
+        Parameters
+        ----------
+        data: Any
+            The data to save.
+
```
simlang, line 60: add saving_kwargs

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
```diff
@@ -0,0 +1,80 @@
...
+        Returns
+        -------
+        Path
+            The path to the saved artifact.
+        """
+        canonical_filename = f"{int(time.time())}_{secrets.token_hex(4)}.{self.export_format}"
+        canonical_path = Path(str(self.root)) / "canonical" / canonical_filename
```
simlang, line 67: same comment as before

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
```diff
@@ -0,0 +1,80 @@
...
+
+        #  all diffusers saving utility functions accept a list of PIL.Images, so we convert to PIL to be safe.
+        if isinstance(data, torch.Tensor):
+            data = np.transpose(data.cpu().numpy(), (0, 2, 3, 1))
+            data = np.clip(data * 255, 0, 255).astype(np.uint8)
```
simlang, line 72: this doesn't convert to PIL.Image as the comment before the if states. one of them has to be changed 🤠

src/pruna/evaluation/evaluation_agent.py
simlang: thanks for the clarification - makes sense!

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -212,22 +248,42 @@ def update_stateful_metrics(
...
+                        canonical_paths.append(canonical_path)
+
+                batch = move_batch_to_device(batch, self.device)
+                processed_outputs = move_batch_to_device(processed_outputs, self.device)
+                (x, gt) = batch
+                # Non-pairwise (aka single) metrics have regular update.
+                for stateful_metric in single_stateful_metrics:
+                    stateful_metric.update(x, gt, processed_outputs)
+                    if self.save_artifacts and stateful_metric.create_alias:
+                        # Again, we have to create an alias for each sample in the batch.
```
simlang: maybe update the comment to a super short version of the explanation you gave me

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -212,22 +248,42 @@ def update_stateful_metrics(
...
+                        for prompt_idx, prompt in enumerate(x):
+                            if self.artifact_saver.export_format is None:
+                                raise ValueError(
+                                    "Export format is not set.Please set the export format for the artifact saver."
```
simlang: no comment
```suggestion
                                    "Export format is not set. Please set the export format for the artifact saver."
```

tests/evaluation/test_artifactsaver.py
```diff
@@ -0,0 +1,90 @@
...
+    """ Test the artifact save is assigned correctly."""
+    saver = assign_artifact_saver("video", root=tmp_path, export_format="mp4")
+    assert isinstance(saver, VideoArtifactSaver)
+    assert saver.export_format == "mp4"
+
+
+def test_assign_artifact_saver_invalid():
+    """ Test that we raise an error if the artifact saver is assigned incorrectly."""
+    with pytest.raises(ValueError):
+        assign_artifact_saver("text")
```
simlang: maybe change to nonsense modality?
otherwise this will fail, when we add a text artifact saver at some point?

src/pruna/evaluation/artifactsavers/video_artifactsaver.py
begumcig: I see what you mean, it converts them to numpy images, which then is immediately picked by the next if statement which turns them to PIL image

tests/evaluation/test_artifactsaver.py
begumcig: makes so much sense

PR #401, opened 2025-10-09
refactor: algorithm compatibility

johannaSommer:
## Description
To support our efforts of making the package more flexible for both use and integration, this PR removes algorithm groups as a concept from the Pruna Package. In short, this means the following changes:
- Algorithms will exist at the same level (unless extensive utility warrants a folder) and only specify what was formerly known as "algorithm groups" as "tags", which we will use in the documentation to provide guidance to users.
- Importantly, the execution order is no longer determined by the algorithm groups, but each algorithm specifies its requirements ("I am compatible with xyz and it needs to run before/after me"). Before smashing, we build a graph and find a path that satisfies directional constraints, this is then the order of execution.
- The SmashConfig interface is now simplified and accepts various input formats (see below).

## Related Issue
None.

## Type of Change

## How Has This Been Tested?
All tests pass.

## Checklist

### To-Do
- [ ] In a follow up PR: update documentation 

### Examples of new Smash Config Interface
```python
config = SmashConfig.from_list(["fastercache", "diffusers_int8"])

config = SmashConfig.from_dict({"fastercache": {"interval": 4}, "diffusers_int8": True})

config = SmashConfig.from_dict({"fastercache": True, "diffusers_int8": True, "fastercache_interval": 4})

config = SmashConfig(["fastercache", "diffusers_int8"])

config = SmashConfig({"fastercache": {"interval": 4}, "diffusers_int8": True})

config = SmashConfig()
config.add("fastercache")
config.add("diffusers_int8")
```

Let me know what you think about the interface, looking at this code snippet it might feel like too many options, so feel free to agree/disagree and let me know what your favorites are. The old way of specifying the Smash Config will be deprecated but still supported for a while.

---

> [!NOTE]
> Removes algorithm groups in favor of a tag-based AlgorithmRegistry with before/after compat rules and graph-derived execution order, alongside a simplified SmashConfig that activates algorithms directly and updated algorithms/tests to the new model.
> 
> - **Core architecture**:
>   - Introduces `AlgorithmRegistry` with discovery, tag filtering, and lookup; removes group-based discovery (`PRUNA_ALGORITHMS`).
>   - Refactors `PrunaAlgorithmBase`: adds `group_tags`, `compatible_before/after`, compatibility expansion via tags, and new reapply recording (`reapply_after_load[algorithm_name]`).
>   - Adds `AlgorithmTag` enum.
> - **Execution flow**:
>   - Replaces fixed group order with graph-based ordering (`determine_algorithm_order`) using per-algorithm before/after constraints; adds cross-compat checks and package availability checks.
>   - `smash()` now uses registry, runs pre-hooks and applies algorithms in computed order.
> - **Config space/API**:
>   - `SmashConfig` now activates algorithms directly (booleans), with `add()`, `from_list()`, `from_dict()`, `get_active_algorithms()`, and overrideable algorithm order; migrates old group keys when loading.
>   - `SmashConfigurationSpace` registers per-algorithm booleans/conditions and exposes `get_all_algorithms()`.
> - **Algorithms**:
>   - Renames/modernizes classes (e.g., `HQQ`, `LLMInt8`, `DiffusersInt8`, `TorchCompile`, `StableFast`, `DeepCache`, `FORA`, `FasterCache`, `PAB`, `QKVFusing`, `Half`, `GPTQ`, `Torchao`, `TorchDynamic`, `CTranslate` variants), adds `group_tags`, `compatible_before/after`, and adjusts `save_fn`/imports; minor fixes (e.g., `flash_attn3`).
> - **Load/Save/Utils**:
>   - Updates HQQ load/save paths to new classes; adjusts reapply logic; minor device/torch artifacts fixes.
> - **Handlers/Metrics/Logging**:
>   - Minor signature/doc fixes; small robustness tweaks.
> - **Tests**:
>   - Overhauls testers to per-algorithm classes; updates combination tests to list-style configs; adds cycle/order/compatibility symmetry tests; removes legacy group-based tests.
> - **Tooling**:
>   - Lint/config tweaks (ruff ignores, pytest/coverage).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f44407e975eaa9e921b94ebf0605214532c23d3e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/config/test_pre_smash_routines.py
+++ b/tests/config/test_pre_smash_routines.py
@@ -122,7 +122,7 @@ class TestCheckModelCompatibility:
         mock_algorithms = {"algorithm1": mock_algorithm1}
         
         with patch("pruna.config.pre_smash_routines.AlgorithmRegistry", mock_algorithms):
-            with patch("pruna.config.pre_smash_routines.get_device", return_value="cuda"):
+            with patch("pruna.config.pre_smash_routines.get_device_type", return_value="cuda"):
                 check_model_compatibility(model, smash_config)
 
     def test_model_compatibility_model_check_failure(self):
@@ -155,9 +155,10 @@ class TestCheckModelCompatibility:
         mock_algorithms = {"algorithm1": mock_algorithm1}
         
         with patch("pruna.config.pre_smash_routines.AlgorithmRegistry", mock_algorithms):
-            with patch("pruna.config.pre_smash_routines.get_device", return_value="cuda"):
-                with pytest.raises(ValueError, match="algorithm1 is not compatible with model device cuda"):
-                    check_model_compatibility(model, smash_config)
+            with patch("pruna.config.pre_smash_routines.get_device_type", return_value="cuda"):
+                with patch("pruna.config.pre_smash_routines.get_device", return_value="cuda"):
+                    with pytest.raises(ValueError, match="algorithm1 is not compatible with model device cuda"):
+                        check_model_compatibility(model, smash_config)
 
 @pytest.mark.cpu
 class TestCheckAlgorithmPackagesAvailability:


src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +267,46 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
         wrapped_config = SmashConfigPrefixWrapper(smash_config, prefix)
         return self._apply(model, wrapped_config)
 
+    def _expand(self, items: Iterable[str | AlgorithmTag] | None) -> list[str]:
+        """Expand algorithms/tags -> concrete algorithm names."""
```
sharpenb: By expand here we mean collect all algorithms which have a specific tag right?

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +267,46 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
...
+
+    def get_incompatible_algorithms(self) -> list[str]:
+        """Algorithms incompatible with this one."""
+        cb = self._expanded_set(self.compatible_before)
+        ca = self._expanded_set(self.compatible_after)
+        allowed = cb | ca
+        all_algorithms = set(SMASH_SPACE.get_all_algorithms())
+        return sorted(all_algorithms - allowed)
+
+    def _required_execution_order(self, side: Literal["before", "after"]) -> list[str]:
```
sharpenb: Docstring? :)

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +267,46 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
...
+        compat = getattr(self, f"compatible_{side}")
+        opposite = "after" if side == "before" else "before"
+        return self._expand(compat)
+
+    def get_required_before(self) -> list[str]:
```
sharpenb: Here we mean algorithm that is compatible to run before/after.

src/pruna/algorithms/base/tags.py
```diff
@@ -0,0 +1,98 @@
...
+        "quantizer",
+        "Quantization reduces the precision of the model’s weights and activations, making them much smaller in terms of memory required.",
+    )
+    PRUNER = (
+        "pruner",
+        "Pruning removes less important or redundant connections and neurons from a model, resulting in a sparser, more efficient network.",
+    )
+    FACTORIZER = (
+        "factorizer",
+        "Factorization batches several small matrix multiplications into one large fused operation.",
```
sharpenb: It could be also low rank factorization like spit one big matrix multiplicatoin into two small ones :)

src/pruna/config/smash_space.py
```diff
@@ -79,161 +70,35 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        parent = Boolean(algorithm_name)
+        self.algorithms.append(algorithm_name)
+        self.add(parent)
+        # Wrap hyperparameter names with config group and algorithm name
+        for hp in hyperparameters:
+            hp.name = f"{algorithm_name}_{hp.name}"
+            self.add(hp)
+
+            # Store condition s.t. hyperparameter is active if algorithm is active (is True)
+            self.add(EqualsCondition(hp, parent, True))
```
simlang: I think you probably meant to use this IsTrueCondtion above, because otherwise it is never used

src/pruna/config/smash_config.py
simlang: update docstring

src/pruna/config/smash_config.py
```diff
@@ -103,20 +93,70 @@ def __init__(
...
+    @classmethod
+    def from_list(
+        cls,
+        configuration: list[str],
+        batch_size: int = 1,
+        device: str | torch.device | None = None,
+        cache_dir_prefix: str | Path = DEFAULT_CACHE_DIR,
+    ) -> SmashConfig:
+        """
+        Create a SmashConfig from a Configuration object.
```
simlang: more extensive docstring if wee keep this

src/pruna/config/smash_config.py
simlang, line 498: not sure if this method is still supported? If it is, the checks are wrong because they rely on algorithm groups

src/pruna/config/smash_config.py
```diff
@@ -541,39 +528,46 @@ def __setitem__(self, name: str, value: Any) -> None:
...
         else:
-            return self._configuration.__setitem__(name, value)
+            # support old way of activating algorithms
+            if value in SMASH_SPACE.get_all_algorithms():
+                pruna_logger.warning(f"Setting {name} to {value} is deprecated. Please use config.add({value}).")
+            else:
+                pruna_logger.warning(f"Setting {name} deprecated. Please use config.add(dict({name}={value})).")
+            self.add(value)
+
+    def add(self, request: Any) -> None:
```
simlang: make typing stricter to only accept str, list[str], and a dict[str, str]?

src/pruna/config/smash_config.py
```diff
@@ -541,39 +528,46 @@ def __setitem__(self, name: str, value: Any) -> None:
...
+        """
+        Add a value to the SmashConfig.
```
simlang: I'm not sure if i like "value", is there a reason to not just call it algorithm?

src/pruna/config/smash_config.py
```diff
@@ -541,39 +528,46 @@ def __setitem__(self, name: str, value: Any) -> None:
...
+
+        Parameters
+        ----------
+        request : Any
```
simlang: same thing here

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,95 +75,80 @@ def ensure_device_consistency(model, smash_config):
...
+        if not algorithm_class.model_check_fn(model):
+            raise ValueError(f"Model is not compatible with {algorithm_class.algorithm_name}")
+        if not any(device in get_device(model) for device in algorithm_class.runs_on):
+            raise ValueError(
+                f"{algorithm} is not compatible with model device {get_device(model)}, "
+                f"compatible devices are {algorithm_class.runs_on}"
+            )
+
+
+def check_algorithm_packages_availability(smash_config: SmashConfig) -> None:
```
simlang, line 103: maybe a bit nit-picky, but i would argue this should be just called import algorithm_packages, as nothing is checked in this function

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,95 +75,80 @@ def ensure_device_consistency(model, smash_config):
...
+        if "pruna_pro" in algorithm_class.__module__:
+            raise RuntimeError(f"Algorithm {algorithm} is unavailable with pruna.smash")
 
 
-def execute_algorithm_pre_smash_hooks(
-    model: Any, smash_config: SmashConfig, algorithm_dict: dict[str, Any] = PRUNA_ALGORITHMS
-) -> None:
+def execute_algorithm_pre_smash_hooks(model: Any, smash_config: SmashConfig, algorithm_order: list[str]) -> None:
     """
     Loops through all algorithm groups and calls the pre_smash_hook method for each algorithm.
```
simlang: update docstring

src/pruna/config/pre_smash_routines.py
```diff
@@ -170,15 +158,93 @@ def execute_algorithm_pre_smash_hooks(
...
+    """
+    Check if the active algorithms are compatible with each other.
+
+    Parameters
+    ----------
+    smash_config : SmashConfig
+        The SmashConfig object containing the algorithm configuration
+    """
+    active_algorithms = smash_config.get_active_algorithms()
+    algorithm_pairs = list(itertools.permutations(active_algorithms, 2))
```
simlang, line 171: does permutations return both pairs (a, b) and (b, a) or only one of them?
If only one the check below might not be complete

src/pruna/algorithms/base/pruna_base.py
simlang: agree, the naming is a bit unclear

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,98 @@
...
+            # Inspect classes defined in this module (avoid classes only re-exported here)
+            for _, obj in inspect.getmembers(module, inspect.isclass):
+                if obj.__module__ != module.__name__:
+                    continue
+
+                # Must be a subclass (but not the base itself)
+                if not issubclass(obj, PrunaAlgorithmBase) or obj is PrunaAlgorithmBase:
+                    continue
+                # Must be a **direct** child (first-grade)
+                if PrunaAlgorithmBase not in obj.__bases__:
```
simlang: Why do we enforce this? Isn't it sometimes a bit cleaner to have a new parentclass for similar algorithms?

src/pruna/algorithms/base/pruna_base.py
johannaSommer: yes exactly!

src/pruna/algorithms/__init__.py
```diff
@@ -14,30 +14,11 @@
...
-            module = importlib.import_module(f"{algorithms.__name__}.{algorithm_types}.{sub_name}")
-            # discover the algorithm class in the file
-            for obj_name, obj in inspect.getmembers(module, inspect.isclass):
-                # Check whether the algorithm is a grandchild of PrunaAlgo in order to skip the algorithm base classes
-                class_names = [cls.__name__ for cls in obj.__mro__]
-                # usually algorithms are grandchildren of PrunaAlgo, except for cases like ctranslate/cgenerate/cwhisper
-                if "PrunaAlgorithmBase" in class_names and class_names.index("PrunaAlgorithmBase") != 1:
-                    # instantiation of the algorithm makes it ready to call and registers HPs
-                    PRUNA_ALGORITHMS[obj.algorithm_group][obj.algorithm_name] = obj()
+AlgorithmRegistry.discover_algorithms(algorithms)
```
simlang, line 18: cleeeeeean 🪥

src/pruna/algorithms/c_translate.py
```diff
@@ -292,7 +296,7 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
         )
 
 
-class CGenerateCompiler(CTranslateCompiler):
+class CGenerate(CTranslate):
```
simlang, line 301: maybe i'm mis-understanding this, but will these be discovered, as it does not directly inherit from PrunaAlgorithmBase?

src/pruna/algorithms/flash_attn3.py
```diff
@@ -101,13 +104,13 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
             register_custom_backend(imported_packages)
 
-            # replace in all compatible components
-            for component in model.components.values():
-                if hasattr(component, "set_attention_backend") and component.dtype in [
-                    torch.bfloat16,
-                    torch.float16,
-                ]:
-                    component.set_attention_backend("flash_attn3_pruna")
+            for j, block in enumerate(model.transformer.blocks):
```
simlang: is this supposed to be here 👀 ?

src/pruna/algorithms/torchao.py
```diff
@@ -168,7 +152,7 @@ def model_check_fn(self, model: Any) -> bool:
 
     def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         """
-        Quantize the model.
+        Quantize the model with torchao.
```
simlang, line 156: what's the policy regarding docstrings of overriden functions? i saw that e.g. in torch_dynamic you removed the docstring completely.

simlang (commented):
Amazing refactoring - left a view comments, some a bit nit-picky tbh.

I can see some algorithm tests may be failing, because of the removal of algorithm groups.

Would it make sense to have something like "compatible in general", so without any order - didn't give this a long think, so maybe it doesn't make sense.

I have no idea how one would implement this, but what do you think of something like a global compatibly matrix? With this adding a new algorithm is maybe a bit more straightforward - doing it on a algorithm level would mean, one would have to actually go to every other algorithm and add the new algorithm there, to make it compatible, potentially touching a lot of files - but as i said, no clue how to do this 🫠 

I love the new interface for adding algorithms to a config! I don't see a necessity to remove certain ways, the more the merrier i guess - unless it's hard to maintain ofc.

src/pruna/config/pre_smash_routines.py
johannaSommer: actually no imo! This is some form of check - we call the import algorithm packages funtion to see whether we have all the packages installed correctly (e.g. the extras are sometimes missing). If this fails we raise a controlled error and inform the user about what they have to do to install the packages :)

src/pruna/config/pre_smash_routines.py
johannaSommer: permutations are ordered on this case so it returns both!

src/pruna/algorithms/base/registry.py
johannaSommer: good catch - this was still a leftover. We allow any subclass of Pruna Algorithm Base except the base class itself. Is fixed!

src/pruna/config/smash_config.py
```diff
@@ -185,34 +276,17 @@ def load_from_json(self, path: str | Path) -> None:
...
-        for group in expected_groups - current_groups:
-            config_dict[group] = None
-
-        self._configuration = Configuration(SMASH_SPACE, values=config_dict)
+        # Build a map of today’s hyperparameters so we can drop stale keys
+        space_hparams = {hp.name for hp in SMASH_SPACE.get_hyperparameters()}
+
+        # Keep only values that still exist in the space
+        saved_values = {k: v for k, v in config_dict.items() if k in space_hparams}
+
```
gsprochette:
```suggestion
        # Keep only values that still exist in the space, drop stale keys
        supported_hparam_names = {hp.name for hp in SMASH_SPACE.get_hyperparameters()}
        saved_values = {k: v for k, v in config_dict.items() if k in supported_hparam_names}

```

src/pruna/config/smash_config.py
```diff
@@ -58,41 +57,29 @@ class SmashConfig:
...
         device: str | torch.device | None = None,
         cache_dir_prefix: str | Path = DEFAULT_CACHE_DIR,
-        configuration: Configuration | None = None,
     ) -> None:
-        SMASH_SPACE.gather_algorithm_buffer()
         self._configuration: Configuration = (
-            SMASH_SPACE.get_default_configuration() if configuration is None else configuration
+            configuration if isinstance(configuration, Configuration) else SMASH_SPACE.get_default_configuration()
         )
         self.config_space: ConfigurationSpace = self._configuration.config_space
```
gsprochette: What's the reason for handling one type of configuration at the very beginning and an other at the very end of init?
I assume the object needs to be initialize to call add, so could we move these `self._configuration` and `self.config_space` declarations to the end?

src/pruna/config/smash_config.py
```diff
@@ -103,20 +90,104 @@ def __init__(
...
+        Examples
+        --------
+        >>> config = SmashConfig.from_list(["fastercache", "diffusers_int8"])
+        >>> config
+        SmashConfig(
+         'fastercache': True,
+         'diffusers_int8': True,
+        )
+        """
+        return cls(configuration=configuration, batch_size=batch_size, device=device, cache_dir_prefix=cache_dir_prefix)
```
gsprochette, line 154: I think here configuration is a `list[str]` which isn't a valid type for configuration in the init. Should it be a reformating of the `list[str]` into a `dict[str, Any]` and then returning `cls.from_dict(...)`?

src/pruna/config/smash_config.py
```diff
@@ -103,20 +90,104 @@ def __init__(
...
+
+    @classmethod
+    def from_dict(
+        cls,
+        configuration: Dict[str, Any],
+        batch_size: int = 1,
+        device: str | torch.device | None = None,
+        cache_dir_prefix: str | Path = DEFAULT_CACHE_DIR,
+    ) -> SmashConfig:
+        """
```
gsprochette, line 164: Shouldn't `from_list` and `from_dict` accept tokenizers and processors like `load_from_json` does? One advantage then is also that `load_from_json` could just call `from_dict` once the dict has been load and corrected. WDYT?

src/pruna/config/smash_config.py
```diff
@@ -586,20 +620,53 @@ def __getattr__(self, attr: str) -> object:  # noqa: D105
...
+        lines = []
+        for alg in self.get_active_algorithms():
+            lines.append(f"  {alg}")
+            if len(self._configuration.config_space.children_of[alg]) > 0:
+                for child in self._configuration.config_space.children_of[alg]:
+                    child_name = child.name
+                    child_value = self._configuration[child_name]
+                    lines.append(f"      {child_name.removeprefix(alg + '_')}: {convert_numpy_types(child_value)!r},")
+            else:
+                lines.append("      -")
```
gsprochette, line 637: Why do we want that indented "-" print? If we don't we can remove the `if..else`.

src/pruna/config/smash_config.py
```diff
@@ -586,20 +620,53 @@ def __getattr__(self, attr: str) -> object:  # noqa: D105
...
+
+        Returns
+        -------
+        list[str]
+            The active algorithms in this smash config.
+        """
+        all_algorithms = self.config_space.get_all_algorithms()
+        return [k for k, v in self._configuration.items() if v and k in all_algorithms]
+
+    def overwrite_algorithm_order(self, algorithm_order: list[str]) -> None:
```
gsprochette, line 656: love it !

src/pruna/config/smash_config.py
```diff
@@ -586,20 +620,53 @@ def __getattr__(self, attr: str) -> object:  # noqa: D105
...
+        ----------
+        algorithm_order : list[str]
+            The order of algorithms to be applied.
+        """
+        if not set(algorithm_order) == set(self.get_active_algorithms()):
+            raise ValueError("All active algorithms must be contained in the given algorithm order.")
+        self._algorithm_order = algorithm_order
+
+    def disable_saving(self) -> None:
+        """Disable the saving of the SmashConfig."""
```
gsprochette: The docstring could be a bit more explicit as to *why* we would want to call this function, especially so since the attribute that's being changed is an internal one.

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
 
-def check_model_compatibility(
-    model: Any,
-    smash_config: SmashConfig,
-    algorithm_dict: dict[str, Any] = PRUNA_ALGORITHMS,
-) -> None:
+def check_model_compatibility(model: Any, smash_config: SmashConfig) -> None:
     """
-    Check if the model is compatible with the given configuration.
+    Check if the model and its device state is compatible with the given configuration.
```
gsprochette:
```suggestion
    Check if the model and its device state are compatible with the given configuration.
```

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Registry for metrics.
+
+    The registry is a dictionary that maps metric names to metric classes.
+    """
+
+    _registry: Dict[str, Callable[..., Any]] = {}
+
+    @classmethod
+    def discover_algorithms(cls, algorithms_pkg: Any) -> None:
```
gsprochette, line 34: I like the discovery so that we don't have to mark new algorithms added directly inside pruna, but I would love having a function that can be used from outside to manually register a custom algorithm. Let's say I'm writing a new algorithm in a notebook, I'd love to be able to do `AlgorithmRegistry.register(CustomAlgorithm)`, or mark it with a decorator.

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,95 @@
...
+            for _, obj in inspect.getmembers(module, inspect.isclass):
+                if obj.__module__ != module.__name__:
+                    continue
+
+                # Must be a subclass (but not the base itself)
+                if obj is PrunaAlgorithmBase:
+                    continue
+
+                # Must inherit from PrunaAlgorithmBase
+                if PrunaAlgorithmBase not in obj.__mro__:
```
gsprochette: Is there a reason this isn't `if not issubclass(obj, PrunaAlgorithmBase)`?

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,95 @@
...
+        prefix = algorithms_pkg.__name__ + "."
+        for _finder, modname, _ispkg in pkgutil.walk_packages(algorithms_pkg.__path__, prefix):
+            try:
+                module = importlib.import_module(modname)
+            except Exception as e:
+                logging.warning("Skipping %s (import error): %s", modname, e)
+                continue
+
+            # Inspect classes defined in this module (avoid classes only re-exported here)
+            for _, obj in inspect.getmembers(module, inspect.isclass):
```
gsprochette: If I'm correct these are not object but classes right? Then can we call it `algorithm_cls` or something like that instead of suggesting it's been instanciated?

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,95 @@
...
+from pruna.algorithms.base.pruna_base import PrunaAlgorithmBase
+from pruna.algorithms.base.tags import AlgorithmTag
+
+
+class AlgorithmRegistry:
+    """
+    Registry for metrics.
+
+    The registry is a dictionary that maps metric names to metric classes.
+    """
```
simlang, line 29: remains from copy
```suggestion
    """
    Registry for algorithms.

    The registry is a dictionary that maps algorithm names to algorithm instances.
    """
```

src/pruna/algorithms/base/tags.py
```diff
@@ -0,0 +1,113 @@
...
+    names : str
+        The name of the enum member.
+    module : str
+        The module where the enum is defined.
+    qualname : str
+        The qualified name of the enum.
+    type : type
+        The type of the enum.
+    start : int
+        The start index for auto-numbering enum values
```
simlang: i don't get that part of the docstring, tbh

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +268,59 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
...
+        # avoid circular import
+        from pruna.algorithms import AlgorithmRegistry
+
+        out: list[str] = []
+        for it in items:
+            if isinstance(it, str):
+                out.append(it)
+            else:
+                out.extend(AlgorithmRegistry.get_algorithms_by_tag(it))
+        return out
```
simlang: not sure if this is the correct place for this. This feels more like a general utility and has nothing to do with the current algorithm

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +268,59 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
...
+        Get algorithms incompatible with this one.
+
+        Returns
+        -------
+        list[str]
+            The incompatible algorithms.
+        """
+        cb = set(self._expand_tags_into_algorithm_names(self.compatible_before))
+        ca = set(self._expand_tags_into_algorithm_names(self.compatible_after))
+        allowed = cb | ca
```
simlang: So nitpicky i'm doing it myself - but yes clearer variable naming :melting_face:
```suggestion
        compatible_before = set(self._expand_tags_into_algorithm_names(self.compatible_before))
        compatible_after = set(self._expand_tags_into_algorithm_names(self.compatible_after))
        allowed = compatible_before | compatible_after
```

src/pruna/config/smash_config.py
```diff
@@ -103,20 +90,104 @@ def __init__(
...
+            The prefix for the cache directory. Default is DEFAULT_CACHE_DIR.
+
+        Returns
+        -------
+        SmashConfig
+            The SmashConfig object instantiated from the dictionary.
+
+        Examples
+        --------
+        >>> config = SmashConfig.from_dict({"fastercache": True, "diffusers_int8": True})
```
simlang, line 185: should we also add to the examples, one where also parameters are provided as dict?

src/pruna/config/smash_config.py
```diff
@@ -103,20 +90,104 @@ def __init__(
...
         self.tokenizer: PreTrainedTokenizerBase | None = None
         self.processor: ProcessorMixin | None = None
         self.data: PrunaDataModule | None = None
         self._target_module: Any | None = None
+
         # internal variable *to save time* by avoiding compilers saving models for inference-only smashing
         self._prepare_saving = True
 
+        # internal variable to overwrite the graph-induced order of algorithms if desired
+        self._algorithm_order: list[str] | None = None
```
simlang, line 99: couldn't comment there, but add this to the equal function

src/pruna/config/smash_config.py
```diff
@@ -305,7 +344,7 @@ def flush_configuration(self) -> None:
         # flush also saving / load functionality associated with a specific configuration
         self.save_fns = []
         self.load_fns = []
-        self.reapply_after_load = dict.fromkeys(ALGORITHM_GROUPS)
+        self.reapply_after_load = {}
```
simlang, line 349: also couldn't comment in the right place, but update examples of this docstring

src/pruna/config/smash_config.py
```diff
@@ -541,39 +554,60 @@ def __setitem__(self, name: str, value: Any) -> None:
...
-            return setattr(self, name, value)
-        else:
-            return self._configuration.__setitem__(name, value)
+        >>> config = SmashConfig()
+        >>> config.add("fastercache")
+        >>> config.add("diffusers_int8")
+        >>> config
+        SmashConfig(
+         'fastercache': True,
+         'diffusers_int8': True,
```
simlang, line 589: similar comment to before, maybe also add example, which shows that you can also pass parameters with the dict

src/pruna/config/smash_config.py
```diff
@@ -541,39 +554,60 @@ def __setitem__(self, name: str, value: Any) -> None:
...
+        if isinstance(request, list):
+            if not all(isinstance(item, str) for item in request):
+                raise ValueError("Request must be a list of algorithm names.")
+            for item in request:
+                self._configuration[item] = True
+        # request wants to activate a dictionary of algorithms and their hyperparameters
+        if isinstance(request, dict):
+            for key, value in request.items():
+                # target modules are a special case, as they are a hyperparameter but their value is a dict
+                if isinstance(value, dict) and not "target_module" in key:
```
simlang: ruff is complaining here - it should probably be something like key.contains("target_module")

src/pruna/config/smash_config.py
```diff
@@ -541,39 +554,60 @@ def __setitem__(self, name: str, value: Any) -> None:
...
+                    self._configuration[key] = True
+                    for k, v in value.items():
+                        if not k.startswith(key):
+                            k = f"{key}_{k}"
+                        self._configuration[k] = v
+                else:
+                    self._configuration[key] = value
```
simlang, line 612: maybe an assertion here, that it has to be a boolean?

src/pruna/config/smash_space.py
```diff
@@ -79,161 +64,40 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
-        self.algorithm_buffer[algorithm_group].append(name)
-
-    def register_algorithm_arguments(self, name: str, hyperparameters: list, algorithm_group: str) -> None:
-        """
-        Register arguments conditional on active algorithm.
-
-        Parameters
-        ----------
-        name : str
+        algorithm_name : str
```
simlang, line 75: i see a lot deleted - i like!

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,98 @@
...
+                if obj is PrunaAlgorithmBase:
+                    continue
+
+                # Must inherit from PrunaAlgorithmBase
+                if PrunaAlgorithmBase not in obj.__mro__:
+                    continue
+
+                # Instantiate & register with the Smash Configuration Space
+                try:
+                    instance = obj()
```
gsprochette: Assuming `obj` has been renamed `algorithm_cls`, this becomes `algorithm = algorithm_cls()`

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,98 @@
...
+                    cls._registry[instance.algorithm_name] = instance
+                except Exception as e:
+                    logging.warning("Failed to instantiate %s from %s: %s", obj.__name__, modname, e)
```
gsprochette: fstring?

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,95 @@
...
+        ----------
+        tag : AlgorithmTag
+            The tag to get algorithms for.
+
+        Returns
+        -------
+        list[PrunaAlgorithmBase]
+            The algorithms that have the given tag.
+        """
+        return [alg.algorithm_name for alg in cls._registry.values() if tag in alg.group_tags]
```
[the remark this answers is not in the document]
gsprochette: very true, if both the name and algorithm are needed somewhere, this could output a subdict of cls._registry?

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        prefix = algorithms_pkg.__name__ + "."
+        for _finder, modname, _ispkg in pkgutil.walk_packages(algorithms_pkg.__path__, prefix):
+            try:
+                module = importlib.import_module(modname)
+            except Exception as e:
+                logging.warning("Skipping %s (import error): %s", modname, e)
+                continue
+
+            if "global_utils" in modname:
```
gsprochette, line 58: can we have a comment explaining this exception?

src/pruna/algorithms/base/registry.py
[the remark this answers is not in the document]
gsprochette: Also: do we want to return a set instead of a list? `get_incompatible_algorithms` transforms what is basically the output of this into a set, so maybe we should have sets from the start if the algorithms aren't ordered yet.

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -136,10 +116,20 @@ def dataset_required(self) -> bool:
...
     @property
     @abstractmethod
-    def algorithm_group(self) -> str:
-        """Return the config group (i.e. "quantizer", "pruner" or "compiler") of the algorithm."""
+    def group_tags(self) -> list[AlgorithmTag]:
+        """Subclasses need to provide a list of group tags for the algorithm."""
         pass
 
+    @property
+    def compatible_before(self) -> list[str]:
```
gsprochette: Probably for a future PR, but I could see a feature where we can give compatibility with tag logic. For example a new algorithm would be compatible with "(Quantizer and not Pruner) or (Compiler)" or something like that.

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -136,10 +116,20 @@ def dataset_required(self) -> bool:
 
     @property
     @abstractmethod
-    def algorithm_group(self) -> str:
-        """Return the config group (i.e. "quantizer", "pruner" or "compiler") of the algorithm."""
+    def group_tags(self) -> list[AlgorithmTag]:
```
gsprochette, line 123: why not just tags?

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -278,6 +268,59 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
...
+    def get_algorithms_to_run_before(self) -> list[str]:
+        """
+        Get algorithms required to run / be executed before the current algorithm.
+
+        Returns
+        -------
+        list[str]
+            The required algorithms.
+        """
+        return self._expand_tags_into_algorithm_names(self.compatible_before)
```
gsprochette: Either the typing of the compatiblity properties are wrong, or I don't understand why these `_expand_tags_into_algorithm_names` calls are needed, aren't those already lists of algorithm_names?

src/pruna/algorithms/base/tags.py
gsprochette: Me neither, was this copy pasted from somewhere?

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
-                    f"{algorithm} is not compatible with device {get_device(model)}, "
-                    f"compatible devices are {algorithm_dict[current_group][algorithm].runs_on}"
-                )
-
-
-def check_argument_compatibility(smash_config: SmashConfig, algorithm_name: str) -> None:
-    """
-    Check if the SmashConfig has the required arguments (tokenizer, processor, dataset) for an algorithm.
+    """
+    for algorithm in smash_config.get_active_algorithms():
```
gsprochette: can `algorithm` be renamed to `algorithm_name` for readability please?

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+        algorithm_class = AlgorithmRegistry[algorithm]
```
gsprochette: elements in the registry have been instanciated, so `algorithm = AlgorithmRegistry[algorithm_name]`

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -35,23 +36,8 @@ class PrunaAlgorithmBase(ABC):
...
-        # register allowed combinations
-        SMASH_SPACE.register_allowed_combinations(self.algorithm_name, self.compatible_algorithms)
-
-        # register argument requirements
-        SMASH_SPACE.model_requirements[self.algorithm_name] = dict(
-            dataset_required=self.dataset_required,
-            tokenizer_required=self.tokenizer_required,
-            processor_required=self.processor_required,
-        )
+        if self.algorithm_name not in SMASH_SPACE.get_all_algorithms():
```
gsprochette, line 39: shouldn't `algorithm_name` just be called `name`? The object will always be called `algorithm` or something like that, so you would already get `algorithm.name`.

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
-        raise ValueError(f"{algorithm_name} requires a tokenizer. Please provide it with smash_config.add_tokenizer().")
-    if algorithm_requirements["processor_required"] and smash_config.processor is None:
-        raise ValueError(f"{algorithm_name} requires a processor. Please provide it with smash_config.add_processor().")
-    if algorithm_requirements["dataset_required"] and smash_config.data is None:
-        raise ValueError(f"{algorithm_name} requires a dataset. Please provide it with smash_config.add_data().")
-    if smash_config._target_module is not None:
-        raise ValueError("Target module is only available in experimental mode. Please set experimental=True.")
+    for algorithm in smash_config.get_active_algorithms():
+        algorithm_class = AlgorithmRegistry[algorithm]
+
```
gsprochette: `algorithm` -> `algorithm_name` and `algorithm_class` -> `algorithm`, same in check_algorithm_availability, execute_algorithm_pre_smash_hooks

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+        The SmashConfig object containing the algorithm configuration.
     """
-    if algorithm_group not in algorithm_dict:
-        raise RuntimeError(f"Algorithm group {algorithm_group} is unavailable with pruna.smash")
-    if algorithm not in algorithm_dict[algorithm_group]:
-        raise RuntimeError(f"Algorithm {algorithm} is unavailable with pruna.smash")
+    for algorithm in smash_config.get_active_algorithms():
+        algorithm_class = AlgorithmRegistry[algorithm]
+        if "pruna_pro" in algorithm_class.__module__:
+            raise RuntimeError(f"Algorithm {algorithm} is unavailable with pruna.smash")
```
gsprochette: Is this truely needed?

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+    smash_config : SmashConfig
+        The SmashConfig object containing the algorithm configuration.
+
+    Returns
+    -------
+    list[str]
+        The order of the active algorithms to be applied.
+    """
+    if smash_config._algorithm_order is not None:
+        return smash_config._algorithm_order
```
gsprochette, line 239: I could see an order compatibility check and involving the experimental flag here. What I mean is the difference between these two situation:
- algorithms A and B are compatibile in any order and the user wants to apply them in a specific order: great!
- algorithms A and B are not explicitely compatible: should we raise an error unless the experimental flag is used or should we assume that manually setting _algorithm_order is enough? Should we at least raise a warning?

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+
+    for algorithm in active_algorithms:
+        graph.add_node(algorithm)
+
+    for alg_a, alg_b in algorithm_pairs:
+        if alg_a in AlgorithmRegistry[alg_b].get_algorithms_to_run_before():
+            graph.add_edge(alg_a, alg_b)
+        if alg_a in AlgorithmRegistry[alg_b].get_algorithms_to_run_after():
+            graph.add_edge(alg_b, alg_a)
+
```
gsprochette, line 276: A bit clearer to me, let me know if it's also the case for you

```suggestion
    for algorithm_name_a in active_algorithms:
        algorithm = AlgorithmRegistry[algorithm]
        for algorithm_name_b in algorithm.get_algorithms_to_run_before():
            graph.add_edge(algorithm_name_b, algorith_name_a)
        for algorithm_name_b in algorithm.get_algorithms_to_run_after():
            graph.add_edge(algorithm_name_a, algorith_name_b)

```

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+    Parameters
+    ----------
+    graph : nx.DiGraph
+        The directed graph to remove reciprocal edges from.
+    """
+    # Undirected view collapses mutual edges into one
+    reciprocal_pairs = {
+        (u, v) for u, v in graph.to_undirected().edges() if graph.has_edge(u, v) and graph.has_edge(v, u)
+    }
+
```
gsprochette:
```suggestion
    reciprocal_pairs = list(graph.to_undirected(reciprocal=True).edges())

```

src/pruna/config/pre_smash_routines.py
```diff
@@ -72,113 +75,184 @@ def ensure_device_consistency(model, smash_config):
...
+    graph.remove_edges_from((u, v) for u, v in reciprocal_pairs)
+    graph.remove_edges_from((v, u) for u, v in reciprocal_pairs)
```
gsprochette:
```suggestion
    reciprocal_subgraph = graph.to_undirected(reciprocal=True).to_directed()
    graph.remove_edges_from(reciprocal_subgraph.edges())
```

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,98 @@
...
+        algorithms_pkg : Any
+            The package to discover algorithms in.
+
+        Returns
+        -------
+        None
+            This function does not return anything.
+        """
+        prefix = algorithms_pkg.__name__ + "."
+        for _finder, modname, _ispkg in pkgutil.walk_packages(algorithms_pkg.__path__, prefix):
```
gsprochette:
```suggestion
        for __, modname, __ in pkgutil.walk_packages(algorithms_pkg.__path__, prefix):
```

src/pruna/evaluation/metrics/metric_model_architecture.py
```diff
@@ -166,12 +166,14 @@ def generate_dummy_inputs(
...
-        # rather than passing z as a keyword argument
+        """
+        Reorder arguments to match the module's forward method signature
+        Example: If forward(self, x, y, z=None) and we have:
+          dummy_args = [tensor1, tensor2] # ruff: noqa: ERA001
+          dummy_kwargs = {'z': tensor3} # ruff: noqa: ERA001
+        This ensures we return (tensor1, tensor2, tensor3) in the correct order
+        rather than passing z as a keyword argument
+        """
         for param in sig.parameters.values():
```
gsprochette: can we go back to # comments instead of a docstring? I don't think we should be doing this, unless there is some very specific reason of this being picked up by a documentation function, but even then there would be cleaner options I think.

src/pruna/algorithms/half.py
```diff
@@ -31,36 +32,26 @@ class HalfQuantizer(PrunaQuantizer):
...
+    compatible_after: list[str] = [
+        "deepcache",
+        "c_translate",
+        "c_generate",
+        "c_whisper",
+        "stable_fast",
+        "torch_compile",
+        "ifw",
+        "whisper_s2t",
+        "pab",
```
gsprochette: pab was not marked as compatible before the refactor, did you test this?

src/pruna/algorithms/pab.py
```diff
@@ -48,7 +52,8 @@ class PABCacher(PrunaCacher):
     processor_required: bool = False
     dataset_required: bool = False
     runs_on: list[str] = ["cpu", "cuda", "accelerate"]
-    compatible_algorithms: dict[str, list[str]] = dict(quantizer=["hqq_diffusers", "diffusers_int8"])
+    compatible_before: list[str] = ["hqq_diffusers", "diffusers_int8", "half"]
+    compatible_after: list[str] = ["torch_compile"]
```
gsprochette: half and torch_compile were not marked as compatible before the refactor. Did you test this?

src/pruna/algorithms/qkv_diffusers.py
```diff
@@ -30,6 +31,7 @@ class QKVDiffusers(PrunaFactorizer):
     """
 
     algorithm_name: str = "qkv_diffusers"
```
gsprochette, line 34: the name is now incoherent with the class name, are we ok with this?

src/pruna/algorithms/base/tags.py
johannaSommer: We are inheriting from an Enum which has these as class construction arguments, hence linting requires us to make them explicit here

src/pruna/algorithms/base/pruna_base.py
johannaSommer: yes I see what you mean... Didn't want to create another `utils.py` somewhere but I removed it from the class and made it a seperate utility function, hope that resolves it

src/pruna/config/smash_config.py
johannaSommer: Agree with you but added it to the documentation changes instead, otherwise the docstring might be quite long

src/pruna/config/smash_config.py
johannaSommer: same comment as above, will be shown in documentation

src/pruna/config/smash_config.py
johannaSommer: Hm not sure I understand, at least in the place that you highlighted it doesn't have to be boolean from what I can tell

src/pruna/config/smash_config.py
johannaSommer: Yeah I know what you mean and I agree, I guess we can just move tokebiner, data, etc. also to the Smash Config constructor because why not. I would prefer to leave this for a follow-up PR because its independent from the current refactoring

src/pruna/config/smash_config.py
johannaSommer: I guess that is up for preference, but when looking at the printed smash config I liked it a bit better to print a "-" to indicate that this algorithm doesnt have any hyperparameters

src/pruna/algorithms/base/registry.py
johannaSommer: yes because issubclass doesnt work with double inheritance, e.g. for recoverers etc

src/pruna/algorithms/base/pruna_base.py
johannaSommer: i think too general, we might introduce more tags in the future, e.g. compatibility tags or metric tags

src/pruna/algorithms/base/pruna_base.py
johannaSommer: fixed!

src/pruna/config/pre_smash_routines.py
johannaSommer: Not sure I 100% understand, my intentions were: if the user wants to set the order they can for sure always do that without warning, if two algorithms are not compatible the user _has_ to set experimental. Then we treat them as "no order specified" which might result in an error yes - but: in the documentation we clearly state that experimental can lead to errors. The user can attempt to then fix this with the order.

src/pruna/evaluation/metrics/metric_model_architecture.py
johannaSommer: No there is a specific reason, we do not allow commented-out-code for linting reasons, which this example hits. We can avoid it by putting it into docstrings though

src/pruna/algorithms/qkv_diffusers.py
johannaSommer: yes I know what you mean, I am not a big fan of putting "diffusers" into everything so I'd prefer it like this for conciseness but if you feel like the discrepancy is a problem we can definitely align it

src/pruna/config/smash_config.py
simlang: okay then i probably misunderstood something!

src/pruna/config/pre_smash_routines.py
johannaSommer: did this and then spent 2 hours debugging mysterious errors lol (super my fault though). The logic you suggested builds a complete dependency graph but we only want that for the currently active algorithms. Tried to extend yours and basically arrived at something not better or worse than what was originally there, so I went back to what I had. Thanks for your attention to this matter, JS MLRE at Pruna.

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,114 @@
...
+                    continue
+
+                # Must inherit from PrunaAlgorithmBase
+                if PrunaAlgorithmBase not in algorithm_cls.__mro__:
+                    continue
+
+                # Instantiate & register with the Smash Configuration Space
+                try:
+                    instance = algorithm_cls()
+                    cls._registry[instance.algorithm_name] = instance
```
gsprochette: I think it would make sense to call the `register_algorithm` method here instead to centralize the _registry access logic

src/pruna/algorithms/base/registry.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+class AlgorithmRegistry:
+    """
+    Registry for algorithms.
+
+    The registry is a dictionary that maps algorithm names to algorithm instances.
+    """
+
+    _registry: Dict[str, Callable[..., Any]] = {}
```
gsprochette, line 31: Shouldn't the type be `Dict[str, PrunaAlgorithmBase]`?

src/pruna/config/pre_smash_routines.py
gsprochette: What you propose is super legit but I think there is one case that slips through it. Imagine the following: the user wants to use algorithms A and B, which are compatible in the order A -> B, so the compatibility check in smash accepts without the need for experimental. However, user explicitely defines the order B -> A which is not marked as compatible.

Currently, to the best of my knowledge we only do an undirected compatibility check before determining the algorithm order, so this case would pass, but it should fail a directed compatibility check after determining the algorithm order if the user didn't use the experimental flag.

src/pruna/config/pre_smash_routines.py
```diff
@@ -176,23 +175,69 @@ def check_algorithm_cross_compatibility(smash_config: SmashConfig) -> None:
...
+    list[tuple[str, str]]
+        A list of violations, where each tuple contains (alg_before, alg_after) indicating
+        that alg_before appears before alg_after in the order but should actually run after it.
+    """
+    violations = []
+    for i, alg_before in enumerate(algorithm_order):
+        for j, alg_after in enumerate(algorithm_order):
+            # alg_before comes before alg_after in the provided order
+            # Check if there's an edge from alg_after to alg_before in the graph
+            if i < j and graph.has_edge(alg_after, alg_before):
```
gsprochette: This relies on the fact that bi-directional edges have been removed from the graph before this check, correct? I think `if i < j and not graph.has_edge(alg_before, alg_after):` would be more resilient to a change in the way the graph is defined

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -389,7 +389,7 @@ def _wrapper(self, *args, **kwargs):
     return _wrapper
 
 
-def _expand_tags_into_algorithm_names(items: Iterable[str | AlgorithmTag] | None) -> list[str]:
+def _expand_tags_into_algorithm_names(items: Iterable[str | AlgorithmTag] | None) -> set[str]:
     """Expand algorithms/tags -> concrete algorithm names."""
     if not items:
         return []
```
gsprochette, line 395: If we return a set: the empty here should be set() instead of []

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -403,4 +403,4 @@ def _expand_tags_into_algorithm_names(items: Iterable[str | AlgorithmTag] | None
             out.append(it)
         else:
             out.extend(AlgorithmRegistry.get_algorithms_by_tag(it))
-    return out
+    return set(out)
```
gsprochette: If we return a set: it feels like `AlgorithmRegistry.get_algorithms_by_tag` should also return a set so then out would be set from the start and we would return `out | {it}` or `out | AlgorithmRegistry.get_algorithms_by_tag(it)`

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -323,9 +323,9 @@ def get_incompatible_algorithms(self) -> list[str]:
         compatible_after = set(_expand_tags_into_algorithm_names(self.compatible_after))
         allowed = compatible_before | compatible_after
         all_algorithms = set(SMASH_SPACE.get_all_algorithms())
-        return sorted(all_algorithms - allowed)
+        return set(all_algorithms - allowed)
```
gsprochette: If we return a set: The last set is not necessary, and maybe `SMASH_SPACE.get_all_algorithms` should already return a set

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -310,7 +310,7 @@ def apply(self, model: Any, smash_config: SmashConfig) -> Any:
         wrapped_config = SmashConfigPrefixWrapper(smash_config, prefix)
         return self._apply(model, wrapped_config)
 
-    def get_incompatible_algorithms(self) -> list[str]:
+    def get_incompatible_algorithms(self) -> set[str]:
```
gsprochette: If we return a set: the docstring's return type is wrong, it is also the case fore.g. get_algorithms_to_run_after

gsprochette (approved):
Perfection

PR #406, opened 2025-10-18
Reduce Flakiness in CI by configuring HF Token and add caching for HF datasets and models

ParagEkbote:
## Description

Some test runs have been failing due to rate limits and usage of memory intensive datasets and models.:

https://github.com/PrunaAI/pruna/actions/runs/18522935006/job/52786867749

https://github.com/PrunaAI/pruna/actions/runs/18612146084/job/53071775373

https://github.com/PrunaAI/pruna/actions/runs/18194737280/job/52458561387

I have introduced caching for datasets and models, while adding the HF token as an env variable for better rate limits. I think we also need to add the HF token as a github secret and add retry attempts to fix the remaining failures.

Could you please review?

cc: @johannaSommer 

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

## Additional Notes

---

> [!NOTE]
> Configure Hugging Face auth and caching in CI and enable pytest reruns to reduce test flakiness.
> 
> - **CI/Workflows (`.github/workflows/tests.yaml`)**:
>   - **Hugging Face setup**: Add `HF_TOKEN`, `HF_HOME`, and dataset/model cache env vars; export `HUGGINGFACE_HUB_TOKEN` for subprocesses.
>   - **Caching**: Cache `huggingface` datasets, hub, and models; clean any `*.incomplete` cache dirs before tests.
>   - **Testing**: Run pytest with reruns (`--reruns 3 --reruns-delay 10 --maxfail=1`) while keeping xdist workers.
> - **Tooling (`pyproject.toml`)**:
>   - Add `pytest-rerunfailures` to `dev` dependencies.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 66ee369519d7de505adca8b651f75b7849aa3cfb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,7 +45,6 @@ jobs:
             **/*.py
           files_ignore: |
             tests/**
-            
       - name: Run ruff on changed Pruna code
         if: steps.changed-files.outputs.any_changed == 'true'
         uses: astral-sh/ruff-action@v3


.github/workflows/tests.yaml
```diff
@@ -71,7 +71,9 @@ jobs:
 
     env:
       HF_ENDPOINT: https://hf-mirror.com
-
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+      HF_HOME: ${{ github.workspace }}/.cache/huggingface
+      HF_DATASETS_CACHE: ${{ github.workspace }}/.cache/huggingface/datasets
```
davidberenstein1957, line 76: why is this explicitely needed?

.github/workflows/tests.yaml
```diff
@@ -71,7 +71,9 @@ jobs:
 
     env:
       HF_ENDPOINT: https://hf-mirror.com
-
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
davidberenstein1957, line 74: @johannaSommer I beleive were already using our HF_TOKEN, given that we push our models to our, organisation, right?

davidberenstein1957 (commented):
Thanks @ParagEkbote , two minor commenct. Thanks for the help :)

.github/workflows/tests.yaml
johannaSommer: I am also still trying to figure out these issues but from what I currently understand this is actually exactly the change we need! In this tests workflow the HF token isnt set so our rate limits are that of an "unregistered" user. Setting the HF token here should fix almost of our issues (afaik)

.github/workflows/tests.yaml
ParagEkbote: These env variables make the cache_dir set in a pre-defined path and will be simpler to debug in the future

If these are not set, we would lose the cache on every workflow run.

.github/workflows/tests.yaml
```diff
@@ -83,9 +86,45 @@ jobs:
...
+          fi
+
       - name: Set test worker count
         run: echo "PYTEST_WORKERS=$(( $(nproc) - 1 ))" >> "$GITHUB_ENV" # leave 1 for the system process
 
-      - name: Run tests
+ - name: Run tests with pytest-rerunfailures
         run: |
-          pytest -n ${{ env.PYTEST_WORKERS }} -m "cpu and not slow and not high_cpu"
+          echo "Running tests with up to 3 reruns on failure using $PYTEST_WORKERS workers..."
```
ParagEkbote, line 129: I've readjusted the approach and have decided to use `pytest-rerunfailures` to retry failing tests. Due to this only a single test is failing and it seems unrelated to the rate-limiting errors.

```
ERROR tests/algorithms/test_algorithms.py::test_full_integration[TestIFW_whisper_tiny_random-cpu] - IndexError: list index out of range
```

Could you please review?

cc: @johannaSommer

.github/workflows/tests.yaml
```diff
@@ -71,6 +72,9 @@ jobs:
 
     env:
       HF_ENDPOINT: https://hf-mirror.com
```
johannaSommer, line 73: Now that we are properly handling the requests to HF, do we still need this mirror or can we send our requests directly to HF?

ParagEkbote: I think that using the hf mirror is a good approach, sending requests directly to hf could result in 429 errors due to many gh runners accessing hf directly.

johannaSommer: makes a lot of sense, thanks a lot again @ParagEkbote

johannaSommer:
LGTM, let's wait for @davidberenstein1957 's final check the let's merge this ASAP so that other contributors are unblocked. 🚀🌻

.github/workflows/tests.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
             **/*.py
           files_ignore: |
             tests/**
-
+
```
davidberenstein1957:
```suggestion

```

.github/workflows/tests.yaml
```diff
@@ -71,6 +71,9 @@ jobs:
 
     env:
       HF_ENDPOINT: https://hf-mirror.com
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
Borda, line 74: Would this work with forked PRs, where GH does not share secrets?

ParagEkbote: As seen in this [github resource](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow), secrets are not passed to the runner when a workflow is triggered from a forked repository, but since public models are used, so there should not be any issues for breaking the CI and being within the rate limits of HF. 

Also, since HF_TOKEN is passed, we get a [better rate limit](https://huggingface.co/docs/hub/en/rate-limits#rate-limit-tiers) than anonymous users and CI caching prevents repeated downloads.

cc: @Borda

PR #368, opened 2025-10-01
Add TinyCIFAR10 dataset for lightweight experiments

kris70lesgo:
## Description

This PR adds a new TinyCIFAR10 dataset to Pruna's data module, providing a lightweight version of the CIFAR-10 dataset with fewer than 1,000 samples for quick experiments, testing, and proof-of-concepts.

## Related Issue

Fixes #358

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

- Added TinyCIFAR10 to the existing parametrized test suite in `test_datamodule.py`
- Verified the dataset follows the same pattern as existing CIFAR10 implementation
- Confirmed the dataset size constraint (<1,000 samples: 600 train + ~200 val + 200 test)
- Tested dataset loading and splitting functionality
- Verified proper integration with Pruna's collate function and SmashConfig

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Additional Notes

- The implementation reuses the existing `uoft-cs/cifar10` dataset and creates a subset rather than requiring a new dataset upload
- Maintains the same 32x32 image format and 10 classes as the full CIFAR10 dataset
- Uses the same `image_classification_collate` function for consistency
- Total sample count: ~800 samples (600 train + ~200 val + 200 test), staying well under the 1,000 sample requirement
- Can be used with `smash_config.add_data("TinyCIFAR10")` following the same API as other Pruna datasets

### Files Modified

- `src/pruna/data/datasets/image.py` - Added `setup_tiny_cifar10_dataset()` function
- `src/pruna/data/__init__.py` - Added import and registered TinyCIFAR10 in base_datasets
- `tests/data/test_datamodule.py` - Added TinyCIFAR10 test case

---

> [!NOTE]
> Adds TinyCIFAR10 dataset and updates CIFAR10 to use an `image` column for collate compatibility, with tests and registry wiring.
> 
> - **Datasets (image)**:
[...]

PR #422, opened 2025-10-29
feat: add tiny datasets for lightweight experiments

begumcig:
## Description

This PR introduces the Tiny CIFAR dataset.

The core implementation was contributed by [kris70lesgo](https://github.com/kris70lesgo) in PR https://github.com/PrunaAI/pruna/pull/368 and this branch brings their work into the main PrunaAI repository. I made minor adjustments (styling, tests) to align with our current codebase and standards.

Full credit for the original implementation goes to @kris70lesgo  💜💜💜 Thanks a lot for your amazing contribution!

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/data/__init__.py
+++ b/src/pruna/data/__init__.py
@@ -78,9 +78,24 @@ base_datasets: dict[str, Tuple[Callable, str, dict[str, Any]]] = {
         "image_classification_collate",
         {"img_size": 32},
     ),
-    "TinyCIFAR10": (partial(setup_cifar10_dataset, fraction=0.1), "image_classification_collate", {"img_size": 32}),
-    "TinyMNIST": (partial(setup_mnist_dataset, fraction=0.1), "image_classification_collate", {"img_size": 28}),
-    "TinyImageNet": (partial(setup_imagenet_dataset, fraction=0.1), "image_classification_collate", {"img_size": 224}),
+    # our full CIFAR10 has 50k train and 10k test
+    "TinyCIFAR10": (
+        partial(setup_cifar10_dataset, train_sample_size=800, test_sample_size=100),
+        "image_classification_collate",
+        {"img_size": 32},
+    ),
+    #  our full MNIST has 60k train and 10k test
+    "TinyMNIST": (
+        partial(setup_mnist_dataset, train_sample_size=800, test_sample_size=100),
+        "image_classification_collate",
+        {"img_size": 28},
+    ),
+    # our full ImageNet has 100k train and 10k val
+    "TinyImageNet": (
+        partial(setup_imagenet_dataset, train_sample_size=1000, test_sample_size=100),
+        "image_classification_collate",
+        {"img_size": 224},
+    ),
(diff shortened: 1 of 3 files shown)


src/pruna/data/datasets/image.py
```diff
@@ -76,8 +80,48 @@ def setup_cifar10_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+        Contains approximately 600 training samples, split validation, and 200 test samples.
+    """
+    train_ds, test_ds = load_dataset("uoft-cs/cifar10", split=["train", "test"])
+
+    # Rename 'img' column to 'image' to match collate function expectations
+    # This ensures compatibility with image_classification_collate function
+    train_ds = train_ds.rename_column("img", "image")
+    test_ds = test_ds.rename_column("img", "image")
+
+    tiny_train = train_ds.select(range(600))
```
davidberenstein1957: why are we just getting a this specific smaller subset? Can't we generalise this approach across all datasets and create general logic for getting tiny versions? perhaps to be tackled in a seperate PR?

src/pruna/data/__init__.py
```diff
@@ -77,6 +78,7 @@
         "image_classification_collate",
         {"img_size": 32},
     ),
+    "TinyCIFAR10": (setup_tiny_cifar10_dataset, "image_classification_collate", {"img_size": 32}),
```
davidberenstein1957: I can see us re-using somethign like `get_tiny(setup_cifar10_dataset)` or something.

src/pruna/data/datasets/image.py
begumcig: Yes this makes a lot  of sense actually!

src/pruna/data/__init__.py
```diff
@@ -77,6 +78,9 @@
         "image_classification_collate",
         {"img_size": 32},
     ),
+    "TinyCIFAR10": (partial(setup_cifar10_dataset, fraction=0.1), "image_classification_collate", {"img_size": 32}),
+    "TinyMNIST": (partial(setup_mnist_dataset, fraction=0.1), "image_classification_collate", {"img_size": 28}),
+    "TinyImageNet": (partial(setup_imagenet_dataset, fraction=0.1), "image_classification_collate", {"img_size": 224}),
```
davidberenstein1957: Awesome. Just not 100% sure if a fraction for each of these datasets is small enough, and it is clear how many samples we get now? We could also allow a range/number or something. Not sure if that would be better, but otherwise we can keep it like this.

begumcig: I completely see your point here, I only did fractions since we have `limit_datasets` in `PrunaDataModule` that allows us to give a number to limit the dataset. If you still think also having a number rather than a fraction here makes more sense I am happy to change it, what do you think?

davidberenstein1957: I think setting it to fixed numbers is nicer as we have more control and awareness surrounding the number.

begumcig: You are right, I have also added this feature 🧡🧡

PR #430, opened 2025-11-04
Feat: migrate pre-commit to prek

GreatBahram:
## Description
[prek](https://prek.j178.dev/) is a drop-in replacement for [pre-commit](https://pre-commit.com/).

Its homepage says it’s not yet production ready, but [several projects like Apache Airflow and PDM](https://prek.j178.dev/#who-is-using-prek) are already using it.

After merging this PR, if you have the repo, ideally you should run the following commands:
```
cd pruna
pre-commit uninstall       # remove the old hooks
prek install # ⚡
```
What's the benefit? prek is noticeably quicker at installing hooks [[link](https://prek.j178.dev/benchmark/)]

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
+    rev: v6.0.0
     hooks:
       - id: check-added-large-files
         exclude: uv.lock
@@ -15,7 +15,7 @@ repos:
         exclude: \.md|.svg|.ipynb$
       - id: no-commit-to-branch
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.12
+    rev: v0.14.3
     hooks:
       - id: ruff-check
         types_or: [ python, pyi ]


.pre-commit-config.yaml
```diff
@@ -3,6 +3,7 @@ repos:
     rev: v5.0.0
     hooks:
       - id: check-added-large-files
+        exclude: uv.lock
```
GreatBahram, line 6: `uv.lock` is a more than 500KB

.pre-commit-config.yaml
```diff
@@ -11,20 +12,22 @@ repos:
...
         args: [--fix]
       - id: ruff-format
+        types_or: [ python, pyi ]
 
   - repo: local
     hooks:
       - id: ty
         name: ty check
-        entry: uvx ty check .
+        entry: uvx ty check . --exclude docs/tutorials/*.ipynb
```
GreatBahram, line 30: > [!TIP]
> Here, since we used the `pass_filenames` it means it runs on every file that we have so later we can't use the `exclude` argument in the pre-commit configuration; therefore I need to exclude the `docs/` directory directly.

PR #435, opened 2025-11-11
Kid metric added

minettekaum:
## Description
Added [KID](https://lightning.ai/docs/torchmetrics/stable/image/kernel_inception_distance.html) (Kernel Inception Distance) metric

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
The KID metric implementation has been tested with the following test suite:

1. **Main KID Test** (`test_kid`):
   - Tests KID metric functionality on both CPU and CUDA devices
   - Uses LAION256 dataset with 4 batches of images
   - Validates that KID returns ~0 when comparing identical images (abs < 0.25)
   - Ensures proper handling of `subset_size` parameter (must be smaller than number of samples)
   - Checks for NaN values to ensure numerical stability
   - Test cases: `test_kid[LAION256-cpu]` and `test_kid[LAION256-cuda]` - both PASSED

2. **Call Type Tests** (`test_check_call_type`):
   - Validates KID works correctly in both "single" and "pairwise" evaluation modes
   - Ensures proper call type conversion (`gt_y` for single, `pairwise_gt_y` for pairwise)
   - Test cases: `test_check_call_type[single-kid]` and `test_check_call_type[pairwise-kid]` - both PASSED

3. **All Torch Metrics Tests**:
   - Ran full test suite: `uv run pytest tests/evaluation/test_torch_metrics.py -v`
   - Result: 41 passed, 7 skipped (CUDA tests on macOS), 0 failed
   - Confirmed no regressions in existing metrics

## Checklist

### DIFF
--- a/src/pruna/evaluation/metrics/metric_torch.py
+++ b/src/pruna/evaluation/metrics/metric_torch.py
@@ -107,7 +107,8 @@ def kid_compute(metric: KernelInceptionDistance) -> Any:
     """
     Compute handler for KID metric.
 
-    KID returns (mean, std) but we only need the mean value.
+    KID normally returns (mean, std) but we only need the mean value.
+    The defensive check handles edge cases (e.g., insufficient data, version differences).
 
     Parameters
     ----------
@@ -117,11 +118,11 @@ def kid_compute(metric: KernelInceptionDistance) -> Any:
     Returns
     -------
     Any
-        The computed metric value (mean from tuple).
+        The computed metric value (mean from tuple, or result as-is if not a tuple).
     """
     result = metric.compute()  # type: ignore
     if isinstance(result, tuple) and len(result) == 2:
-        return result[0]  # Extract mean from tuple (KID returns (mean, std))
+        return result[0]  # Extract mean from tuple (KID returns normally (mean, std))
     return result
 
 
--- a/tests/evaluation/test_torch_metrics.py
+++ b/tests/evaluation/test_torch_metrics.py
@@ -149,6 +149,7 @@ def test_torch_metrics(datamodule_fixture: PrunaDataModule, device: str, metric:
     metric.update(gt, gt, gt)
     assert metric.compute().result == 1.0
 
+
 @pytest.mark.cpu
 @pytest.mark.parametrize("datamodule_fixture", ["LAION256"], indirect=True)
 def test_arniqa(datamodule_fixture: PrunaDataModule) -> None:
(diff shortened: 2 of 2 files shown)


src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -341,6 +360,11 @@ def compute(self) -> Any:
         """
         result = self.metric.compute()
 
+        # Handle KID which returns a tuple (mean, std)
+        if self.metric_name == "kid" and isinstance(result, tuple) and len(result) == 2:
```
begumcig: I don't really have a suggestion for this yet but I think we should not do conditional statements in the compute function based on the metric name, as this can get messy quite easily, how do you feel?

minettekaum: Very good point, and I agree. Having conditionals in `compute()` based on the metric name can get messy fast. So I cleaned it up a bit:

- Took out the if `self.metric_name == "kid"` check from `compute()`.
- Added a `kid_compute` function to handle KID’s tuple return.
- Updated the TorchMetrics enum so KID has a 4th element for `kid_compute` (others stay as 3).
- Now `compute()` just uses `self.compute_fn` from the enum if it’s there, otherwise falls back to the default `self.metric.compute()`.

So now the `compute()` method is generic and clean, no metric-specific conditionals. Each metric can have its own compute logic in the enum, just like how we handle the update functions.

What do you think, does this look like a good direction?

src/pruna/evaluation/metrics/metric_torch.py
```diff
@@ -85,6 +86,45 @@ def fid_update(metric: FrechetInceptionDistance, reals: Any, fakes: Any) -> None
...
+    metric : KernelInceptionDistance
+        The KID metric instance.
+
+    Returns
+    -------
+    Any
+        The computed metric value (mean from tuple).
+    """
+    result = metric.compute()  # type: ignore
+    if isinstance(result, tuple) and len(result) == 2:
```
sharpenb, line 124: Are these checks always true? It sounds like it might be based on the comments.

tests/evaluation/test_torch_metrics.py
```diff
@@ -77,6 +112,7 @@ def test_clip_score(datamodule_fixture: PrunaDataModule, device: str) -> None:
     score = metric.compute()
     assert score.result > 0.0 and score.result < 100.0
 
+
```
begumcig, line 115: is this space intentional?

PR #444, opened 2025-11-21
feat: decoder algorithm

minettekaum:
## Description
The ZipAr algorithm was added to Pruna. 

## Type of Change

## Checklist

## Additional Notes
There aren't tests implemented for this algorithm yet. But it has been tested that the script runs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,7 +159,6 @@ full = [
     "xformers>=0.0.30",
     "stable-fast-pruna==1.0.8",
 ]
-
 dev = [
     "wget",
     "python-dotenv",


src/pruna/algorithms/zipar/utils.py
```diff
@@ -0,0 +1,694 @@
...
+        output_attentions: bool,
+        output_hidden_states: bool,
+        generation_config: Any,
+        **model_kwargs: Any,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, Any]:
+        """
+        Generate the image tokens using the ZipAR algorithm.
+
+        This implementation was inspired by the original implementation of ZipAR:
+        https://github.com/ThisisBillhe/ZipAR/blob/2a5695ca2525872ac6ee38c9d62be38f0c9e985b/LlamaGen-ZipAR/autoregressive/models/generate_zipar.py.
```
gsprochette, line 314: Because of the instability of Janus code, this `generate` implementation had to be based on one specific version of transformers, namely 4.54.0 .
I think this would deserve adding a "zipar" extra to pin the transformer version if the user wants to use this algorithm.
- adding a zipar extra in the pyproject.toml
- adding a dynamic transformers version check in the `model_check_fn` function
Ideally we would test different versions of transformer to see when this breaks, but for a first integration I would say pinning is fine and we can test & fix in a following PR

gsprochette (approved):
LGTM, I hope we can generalize this for other transformers versions in a future PR :)

PR #302, opened 2025-08-13
fix: make qkv compatible with torch.compile in next diffusers release

llcnt:
## Description
The current main branch in `diffusers` is undergoing a large refactorization for attention computation.
The attention processors still exist but Flux and Wan have now a local version of their own processor. Also qkv fusing was changed: it is now taking place in the AttentionMixin class (not anymore in the transformer class itself).
Sister PR in pruna_pro is [here](https://github.com/PrunaAI/prunatree/pull/197).

## Related Issue

The combination qkv_diffusers+torch_compile is taking forever to compute on the latest diffusers codebase

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
Tested on Flux-dev with diffusers==0.34.0 and diffusers==0.35.0dev0
On Flux-dev, the generation time goes from 10.44s (original) to 5.12s (with `qkv_diffusers`+`fp8`+`torch_compile`).
The warm-up time (first inference) takes 56.16s.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #452, opened 2025-12-04
feat: add target modules to quantizers

### LINKED ISSUE #386: [FEATURE] Add target modules to torchao quantizer
> ‼️ If you want to work on this issue: please comment below and wait until a maintainer assigns this issue to you before opening a PR to avoid several contributions on the same issue. Thanks! 😊

## **✨ What You’ll Do**

Right now, we have torchao quantizer in Pruna. However, it quantizes every Linear layers whose path in the model does not include "norm" or "embedding", without a way to change that behavior. You task will be to add a TargetModules hyperparameter and use it to select which modules are quantized.

## 🤖 Useful Resources

- [ ] TargetModules [source code](https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/target_modules.py) and [documentation](https://github.com/PrunaAI/pruna/blob/main/docs/user_manual/target_modules.rst) (unfortunately not in the current documentation build)
- [ ] [Bits and Bytes](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/quantization/huggingface_llm_int8.py) and [Quanto](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/quantization/quanto.py) implementations in Pruna which has TargetModules, for reference
- [ ] Add the TargetModules hyperparameter in `get_hyperparameters`
- [ ] Update the modules selection in _apply. In particular, consider using the functions already defined in `target_modules.py`.
- [ ] Define the default target modules in `get_model_dependent_hyperparameter_defaults` with the current behavior
- [ ] Prove that your feature is effective by comparing the list of quantized modules before and after with the default / non-default target modules.

---

## **✅ Acceptance Criteria**

- It follows the style guidelines.
- **Docs** new TargetModules hyperparameter is documented
- **Tests** a code snippet / notebook is provided to prove the change is effective

---

## **❓ Questions?**

Feel free to jump into the [#contributing Discord channel](https://discord.com/invite/JFQmtFKCjd) if you hit any roadblocks. Can’t wait to see your contribution! 🚀

---

## Share on Socials

gsprochette:
## Description
This PR adds target modules capabilities to torchao, awq, hqq and hqq-diffusers.
All apply methods were updated with the map_targeted_nn_roots structure with custom quantization function, similarly to what was done in llm-int8, diffusers-int8 and quanto.

A utility function `target_backbone` was added to provide a global default for target modules: automatically target the transformer, unet or model language.

The save and load functions for hqq were updated to maintain compatibility with this new feature and improve the pipeline saving and loading.

(documentation update because it adds new hyperparameter to the quantizers)

## Related Issue

Fixes #386 

## Type of Change

## How Has This Been Tested?
Ran each algorithm in a notebook to make sure the model can run after quantization, some Linears are quantized and some excluded linears are not.
Some tests don't pass, but they already didn't before this PR.

## Checklist

### DIFF
--- a/src/pruna/engine/utils.py
+++ b/src/pruna/engine/utils.py
@@ -691,6 +691,12 @@ def monkeypatch(obj, attr_name, value):
     value : Any
         The temporary value to set for the attribute.
 
+    Yields
+    ------
+    None
+        This context manager does not yield a value and is intended to be
+        used for its side effects only (temporary attribute patching).
+
     Examples
     --------
     with monkeypatch(some_object, "attribute_name", new_value):


src/pruna/algorithms/hqq.py
```diff
@@ -127,43 +167,107 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+                The subpaths of the component to quantize.
+
+            Returns
+            -------
+            torch.nn.Module
+                The quantized component.
+            """
+            if is_janus_llamagen_ar(module):
+                # dispatch to the language model because HQQ expects a LLM as input
+                lm_path = "model.language_model"
```
begumcig, line 206: Is this a very specific case for Janus, or are there any more models out there that would have this model.language_model structure? Maybe we can make the function name more general if that's the case

src/pruna/algorithms/torchao.py
```diff
@@ -151,6 +164,67 @@ def model_check_fn(self, model: Any) -> bool:
...
+            The configuration for the quantization.
+        """
+        if (
+            smash_config["torch_compile"]
+            and smash_config._base_config["torch_compile_mode"] != "max-autotune-no-cudagraphs"
+        ):
+            pruna_logger.warning(
+                "You are using torchao with torch.compile. "
+                "Please set `smash_config['torch_compile_mode']='max-autotune-no-cudagraphs'` for best results; "
+                "otherwise you may encounter undesirable outcomes."
```
begumcig, line 212: what's the undesirable outcomes here

src/pruna/config/target_modules.py
```diff
@@ -94,6 +102,36 @@ def legal_value(self, value: TARGET_MODULES_TYPE | None):  # type: ignore[overri
...
+    elif is_transformers_pipeline_with_causal_lm(model) or is_janus_llamagen_ar(model):
+        return {"include": ["model.*"], "exclude": ["model.lm_head"]}
+    elif isinstance(model, DiffusionPipeline):
+        denoiser_cls = tuple(get_diffusers_unet_models() + get_diffusers_transformer_models())
+        include = [
+            f"{attr_name}.*" for attr_name, component in model.components.items() if isinstance(component, denoiser_cls)
+        ]
+        return {"include": include, "exclude": []}
+    elif is_janus_llamagen_ar(model):
+        return {"include": ["model.language_model.*"], "exclude": []}
```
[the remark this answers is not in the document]
begumcig: Hmm I do agree with this we already have the janus check at line 121

src/pruna/engine/save.py
```diff
@@ -340,50 +342,55 @@ def save_model_hqq(model: Any, model_path: str | Path, smash_config: SmashConfig
...
-    else:
-        quantized_path = Path(model_path)
-
-    # save the quantized model only.
-    with ModelContext(model, read_only=True) as (_, working_model):
-        if isinstance(working_model, algorithm_packages["HQQModelForCausalLM"]):
-            working_model.save_quantized(quantized_path)
+        # pipeline calls model.save_pretrained, so we monkey patch it to skip the saving
+        original_save_pretrained = model.model.save_pretrained if hasattr(model.model, "save_pretrained") else None
+        model.model.save_pretrained = lambda *args, **kwargs: None
```
begumcig: Can the model here be a PrunaModel or is it the internal model inside?

src/pruna/engine/save.py
```diff
@@ -340,50 +342,55 @@ def save_model_hqq(model: Any, model_path: str | Path, smash_config: SmashConfig
...
+        model.save_pretrained(model_path)
+        if original_save_pretrained is None:
+            delattr(model.model, "save_pretrained")
         else:
-            algorithm_packages["AutoHQQHFModel"].save_quantized(working_model, str(quantized_path))
-
-    # save the rest of the model, if it is a janus like model,
-    # and add a config file to the quantized model path.
-    if hasattr(model, "model") and hasattr(model.model, "language_model"):
+            model.model.save_pretrained = original_save_pretrained
```
begumcig: shall we wrap this in a try finally block so that if even saving fails we revert the saving back to original_save_pretrained? It might be confusing for users if they try to save the model again.

src/pruna/config/target_modules.py
```diff
@@ -94,6 +102,36 @@ def legal_value(self, value: TARGET_MODULES_TYPE | None):  # type: ignore[overri
         return super().legal_value(value)
 
 
+def target_backbone(model: Any) -> TARGET_MODULES_TYPE:
```
begumcig, line 105: I was going to comment that `get_model_dependent_hyperparameter_defaults` is called by every algorithm and maybe we should create a default function for the shared lines across algorithms but of course you already handled that 🏅🏅🏅🏅

src/pruna/algorithms/hqq.py
```diff
@@ -219,3 +322,71 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+        non_saveable_attributes = [
+            path.split(".")[0] for path in targeted_paths if not (path == saveable or path.startswith(f"{saveable}."))
+        ]
+        if non_saveable_attributes:
+            pruna_logger.warning(
+                f"HQQ saving/loading is not implemented for the following attributes: {non_saveable_attributes}"
+            )
+
+
+def construct_base_class(imported_modules: Dict[str, Any], extra_ignore_modules: list[str]) -> Type[Any]:
```
llcnt, line 354: If I understand correctly, you are patching the `get_ignore_layers` function to also include the extra_ignore_modules, right? Maybe I would suggest adding this precision inside docstring to explain, if Im not wrong :)

src/pruna/algorithms/hqq_diffusers.py
```diff
@@ -133,69 +177,89 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
+            torch.nn.Module
+                The quantized component.
+            """
+            module.layers = find_module_layers_type(module, nn.Linear)
+
+            warn_model_specific_errors(module, subpaths)
+
+            ignored_leaf_modules = get_skipped_submodules(module, subpaths, filter_fn=is_leaf_module)
+            auto_hqq_hf_diffusers_model = construct_base_class(
+                imported_modules, extra_ignore_modules=ignored_leaf_modules
```
llcnt, line 211: Did you test this in combination with torch.compile ? I am asking because before we were using a single instance of the HQQ class for the whole model, while now we use an instance for each module. In the [past](https://github.com/PrunaAI/pruna/pull/302) we have seen that having several instance of the same class can increase the graph torch_compile computes and leads to OOM.

src/pruna/engine/load.py
```diff
@@ -480,47 +485,57 @@ def load_hqq_diffusers(path: str | Path, smash_config: SmashConfig, **kwargs) ->
...
-        # So we set it manually here.
+        # force dtype if specified, from_quantized does not set it properly
         if "torch_dtype" in kwargs:
             model.to(kwargs["torch_dtype"])
+    else:  # save directory is for a pipeline, of which some components were quantized
+        model_index = load_json_config(path, "model_index.json")
+
+        # first we load each quantized component separately
+        quantized_components: dict[str, Any] = {}
+        for quantized_path in [qpath for qpath in path.iterdir() if qpath.name.endswith("_quantized")]:
```
llcnt, line 511: here you have several path because you want to manage model like wan that has several transformers in it, right? Maybe a line of comment would help here

src/pruna/engine/utils.py
```diff
@@ -132,6 +132,8 @@ def move_to_device(
     """
     if safe_is_instance(model, Pipeline):
         move_to_device(model.model, device, raise_error, device_map)
+        if device != "accelerate":
```
llcnt, line 135: why do we need to add the device attribute to the model here?

src/pruna/algorithms/hqq.py
gsprochette: This case is very specific to Janus. Basically HQQ here only works on CausalLM, so we seek the CausalLM attribute within the network. The CausalLM and pipeline for causal LM cases are pretty straightforward, and here janus is a special case we decided to make compatible.
I would say making this behavior more general would require to search for causal LM, which is not for this PR. WDYT?

src/pruna/algorithms/hqq.py
gsprochette: I added a line in the docstring to explain that the point of this class is to ignore layers, and did small docs/comments improvements

src/pruna/engine/save.py
gsprochette: It's the internal model inside since this is being called through `PrunaModel.save_pretrained` which calls it on `self.model`.

src/pruna/engine/save.py
gsprochette: That's a super good point. I think a context is a clearer and cleaner way to factorize this code out of here for clarity in this already complicated saving function

src/pruna/algorithms/torchao.py
gsprochette: I don't know, the content of this function was already in the file I just isolated the warnings from the apply logic. Maybe @johannaSommer knows more about this undesirable outcome?

src/pruna/engine/utils.py
gsprochette: In the case of Pipelines apparently the pipeline moves the input ids to the device that is registered in `pipeline.device` instead of `pipeline.model.device`, at least in some versions of transformers. This led to wrong device errors because the pipeline was casting to cpu with a transformer on gpu. The device attribute is declared in Pipeline so I added this in move_to_device.

src/pruna/algorithms/hqq_diffusers.py
gsprochette: I just tested on Flux1dev and Wan-i2v and I don't get unreasonable VRAM usage. Here it's actually only using one HQQ class per component, meaning for flux you define a single class for model.transformer. In Wan-i2v for example you would have only 2 classes: one for model.transformer and one for model.transformer_2.

llcnt (approved):
Thanks for the fixes :)

src/pruna/engine/utils.py
```diff
@@ -132,6 +132,8 @@ def move_to_device(
...
+            model.device = torch.device(device)
```
begumcig, line 136: Could this possibly cause problems if the model doesn't have a device attr? or it's a property that cannot be set directly?

begumcig (approved):
Amazing Gaspar just left one comment about the device attribute, maybe we should talk about it, wdyt? Feel free to ignore it If I am making no sense! Everything looks great overall! Thank you!!

src/pruna/engine/utils.py
gsprochette: In this if block we can assume model is a transformers Pipeline, which relies on getting and setting its `device` attribute (I checked transformers 4.42, 4.57 and 5.0.0rc1 and they all set `self.device = ...` in their init. If you want I can add a `try..except` but it may be better to explicitely fail if there is unknown behavior so we can handle it. WDYT?

begumcig: Sounds great to me! Thank you so much for checking <3

PR #455, opened 2025-12-08
feat: Sage Attention Algorithm

Marius-Graml:
## Description
Integration of the Sage Attention algorithm into the Pruna framework. The current version applies the attention backend from Diffusers, choosing the Sage Attention kernel from the Kernel Hub. This is because the original sageattn function appears to be broken (its outputs were pure noise). Additionally, tests for the Sage Attention algorithm were implemented.

## Related Issue
No issues were fixed.

## Type of Change

## How Has This Been Tested?
Reuse of the tests for flashattn3 adapted to sage attention.

## Checklist

## Additional Notes
/

### DIFF
--- a/src/pruna/algorithms/sage_attn.py
+++ b/src/pruna/algorithms/sage_attn.py
@@ -172,7 +172,7 @@ class SageAttn(PrunaAlgorithmBase):
             A dictionary with keys "include" and "exclude" defining which modules should be
             targeted by default.
         """
-        # We include all transformer modules by default. 
+        # We include all transformer modules by default.
         # SageAttn might also be applicable to other modules but could significantly decrease model quality.
         include = ["transformer*"]
         exclude = []


src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,98 @@
...
+
+    SageAttention is a fast and memory-efficient attention mechanism. It applies the flash attention mechanism
+    in combination with quantization and smoothing to speed up attention computations.
+    """
+
+    algorithm_name: str = "sage_attn"
+    group_tags: list[str] = [tags.KERNEL]
+    save_fn = SAVE_FUNCTIONS.reapply
+    references: dict[str, str] = {
+        "GitHub": "https://github.com/thu-ml/SageAttention",
```
johannaSommer, line 43: Great references already, let's also add one of the papers here i'd say! :)

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,98 @@
...
+        "Kernel Hub": "https://huggingface.co/kernels-community/sage_attention",
+    }
+    tokenizer_required: bool = False
+    processor_required: bool = False
+    runs_on: list[str] = ["cuda", "accelerate"]
+    dataset_required: bool = False
+    compatible_before: Iterable[str] = []
+    compatible_after: Iterable[str] = ["torch_compile"]
```
johannaSommer: compatible after would also be tags.CACHERS and compatible before probably also tags.QUANTIZERS

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,98 @@
...
+        Returns
+        -------
+        bool
+            True if the model is a valid model for the algorithm, False otherwise.
+        """
+        if not isinstance(model, DiffusionPipeline) or not hasattr(model, "components"):
+            return False
+
+        return any(
+            hasattr(component, "set_attention_backend") and component.dtype in [torch.bfloat16, torch.float16]
```
johannaSommer: i recall this dtype check for the components from flash attention (because attention needs to be computed in this precision for FA3 to work), did we double check that that is the case also here?

src/pruna/algorithms/sage_attn.py
johannaSommer: then add this compatibility also in other algorithms

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,98 @@
...
+
+        Returns
+        -------
+        Any
+            The wrapped model.
+        """
+        # We simply apply the sage attention backend from diffusers
+        # Furthermore, we use the sage attention kernel from the hub as the default sageattn function
+        # is broken (at least at the moment)
+        for component in model.components.values():
```
johannaSommer: as discussed, let's add target modules also here :)

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,351 @@
...
+        Get the list of configurable hyperparameters for this algorithm.
+
+        Returns
+        -------
+        list
+            A list of hyperparameter objects (e.g., Boolean, TargetModules) used by the
+            configuration system.
+        """
+        return [
+            Boolean(
```
johannaSommer: this is actually not needed and we can remove it, as the user can specify this exactly through the target modules anyway (there is a smash config interface for this)

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,351 @@
...
+        smash_config : SmashConfigPrefixWrapper
+            The configuration for the application of the algorithm.
+
+        Returns
+        -------
+        Any
+            The wrapped model.
+        """
+        target_modules = smash_config["target_modules"]
+        exclude_first_and_last_transformer_blocks = smash_config["exclude_first_and_last_transformer_blocks"]
```
johannaSommer: for the target modules, let's please use the functionality we already have, otherwise we have a lot of duplicate code here

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,184 @@
...
+        Any
+            The wrapped model.
+        """
+        target_modules = smash_config["target_modules"]
+
+        if target_modules is None:
+            target_modules = self.get_model_dependent_hyperparameter_defaults(
+                model,
+                smash_config
+            )  # for consistency, not used yet
```
johannaSommer: i think we can remove this comment, this is used now right?

src/pruna/algorithms/sage_attn.py
```diff
@@ -0,0 +1,184 @@
...
+
+        Returns
+        -------
+        TARGET_MODULES_TYPE
+            A dictionary with keys "include" and "exclude" defining which modules should be
+            targeted by default.
+        """
+        # So far, everything is included and nothing is excluded
+        # Filtering is done in the _apply method by the set_attention_backend method
+        include = ["*"]
```
johannaSommer: the default should apply it on the transformer, I dont think this is the case right now, no?

src/pruna/algorithms/sage_attn.py
Marius-Graml: Yess

PR #467, opened 2025-12-16
build: unpin torch version

gsprochette:
## Description
This PR unpins the torch version from "torch==2.7.0" to "torch>=2.7.0".

The version was pinned mainly to correspond to the version the stable-fast wheels were built against, so merging this PR will cause instability in the stable-fast install (in stable-fast and full extras) until we upload the new wheels (currently in testing).

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
Ran the tests on torch 2.9 and made sure we are getting the same results as before

## Checklist

### DIFF
--- a/src/pruna/algorithms/torchao.py
+++ b/src/pruna/algorithms/torchao.py
@@ -150,6 +150,13 @@ class Torchao(PrunaAlgorithmBase):
         bool
             True if the model is suitable for torchao quantization, False otherwise.
         """
+        pruna_logger.warning(
+            "torchao has strict version compatibility requirements with torch. "
+            "If you encounter crashes when using torchao, ensure that your torch and torchao "
+            "versions are compatible, as documented in the torchao compatibility table: "
+            "https://github.com/pytorch/ao/issues/2919#issue-3375688762"
+        )
+
         transformer_models = get_diffusers_transformer_models()
         unet_models = get_diffusers_unet_models()
         if isinstance(model, tuple(transformer_models)):


gsprochette:
@ParagEkbote you looked at unpinning the torch version in an other PR, would you like to review this?
One flaw of this PR is that if torch isn't pinned we can't pin torchao either, but the pyproject.toml file doesn't really allow to pin torchao conditionnally on torch so we'll just have to warn the user and rely on them?

ParagEkbote:
> @ParagEkbote you looked at unpinning the torch version in an other PR, would you like to review this? One flaw of this PR is that if torch isn't pinned we can't pin torchao either, but the pyproject.toml file doesn't really allow to pin torchao conditionnally on torch so we'll just have to warn the user and rely on them?

In this PR, why are we aiming to unpin the `torch` version for future builds, since `torch` has platform-specific constraints? Is it simpler to specify the minimum version of `torchao>0.12.0` and update the torch version every minor release?

We could also define `pytorch extras` in optional dependencies for the users who are willing to try out `torch==2.8.0` or newer versions:

```toml
[project.optional-dependencies]
torch29 = [
  "torch==2.9.0",
  "torchao==0.14.1",
]

torch28 = [
  "torch==2.8.0",
  "torchao==0.13.0",
]
```

Another point to be considered is that if we unpin `torchao` itself, it will lead to breakages since the [upstream API](https://github.com/pytorch/ao/releases/tag/v0.14.1) has been updated as compared to the [quantizer](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/torchao.py#L312-L319) defined in `pruna`.

WDYT?

cc: @gsprochette

gsprochette:
@ParagEkbote I would like to keep the a broad range of torch versions so that pruna is compatible with as many setup as possible, so I don't think updating the torch version every minor update is the solution we want.

What I understand is that the torchao quantizer in pruna would crash with newer versions of torch requiring newer versions of torchao, but it is also the only part of pruna requiring the torchao dependency, correct? So maybe we should make a torchao extra instead, and pin the torch and torchao version until we make it compatible with newer versions?

The longer term solution I think would be to have version ranges  for both torch and torchao and warn the user about the version compatibilities, with a link to [this table](https://github.com/pytorch/ao/issues/2919#issue-3375688762).

gsprochette:
This PR is not stale

ParagEkbote:
> @ParagEkbote I would like to keep the a broad range of torch versions so that pruna is compatible with as many setup as possible, so I don't think updating the torch version every minor update is the solution we want.
> 

I understand, it is indeed a bit cumbersome for the long term.

> What I understand is that the torchao quantizer in pruna would crash with newer versions of torch requiring newer versions of torchao, but it is also the only part of pruna requiring the torchao dependency, correct? So maybe we should make a torchao extra instead, and pin the torch and torchao version until we make it compatible with newer versions?
> 

I think that creating `torchao` extras with `torch` and `torchao` versions matching with the compatibility table is a good idea. We could create the extras beyond the current supported version of `torch==2.7.0.` , correct? 

Also, #471 will expand the usage of `torchao` with an additional quantizer algorithm

> The longer term solution I think would be to have version ranges for both torch and torchao and warn the user about the version compatibilities, with a link to [this table](https://github.com/pytorch/ao/issues/2919#issue-3375688762).

I think this approach can work, but would we need to set the torch version at a specified min. version so that it supports torchao and the other libraries such as `llmcompressor`, `hqq`, `torch_pruning` etc. within pruna?

WDYT?

cc: @gsprochette

gsprochette:
@ParagEkbote my prefered option is to unpin torch and torchao and I'll add a warning in torchao (and send a message in PR #471 to add a warning in the new torchao quantizer) so the user can refer to the compatibility table. I like this option because it gives more freedom and we can also directly see what needs to be fixed.

ParagEkbote:
> @ParagEkbote my prefered option is to unpin torch and torchao and I'll add a warning in torchao (and send a message in PR #471 to add a warning in the new torchao quantizer) so the user can refer to the compatibility table. I like this option because it gives more freedom and we can also directly see what needs to be fixed.

Ok, that works for me 👍

src/pruna/algorithms/torchao.py
```diff
@@ -150,6 +150,13 @@ def model_check_fn(self, model: Any) -> bool:
         bool
             True if the model is suitable for torchao quantization, False otherwise.
         """
+        pruna_logger.warning(
```
begumcig, line 153: brilliant

PR #482, opened 2025-12-23
feat: moe kernel tuning

llcnt:
## Description
This PR is inspired from vLLM benchmarks (the benchmark_config fn is copied from [here](https://github.com/vllm-project/vllm/blob/main/benchmarks/kernels/benchmark_moe.py)) and enable one to tune the MoE (triton) kernel used in vllm.
This new algorithm `MoeKernelTuner` does not modify the model. It generates a tuned configuration that is saved in:
- the vllm configs folder (so that using the model on the same gpu afterward makes vllm use this optimized config);
- the RedhatAI kernel folder in the hf cache (so that using the moe kernels from the `kernels` lib will make use of the optimized config);
- a folder `moe_kernel_tuned_configs` in the model directory (to be later re-used without waiting for tuning, when loading the model with pruna).

The core modifications are in:
- the new moe_kernel_tuner.py file ((i) it does not modify the model, so it is compatible with every other algorithm before/after; (ii) the user can select dtypes but also size of parameters gridsearch; (iii) the kernel is tuned for batch sizes(ie the input dimension M) from 1 to 8192 using `ray` for parallelization; (iv) the best configurations are saved in hf and vllm caches (so that after smashing, hf cache and vllm cache are already populated with optimal configs that the user can use), and in the pruna cache (similar to what we do with `save_before_apply`);
- the save_artifacts.py file (we move the tuned config from the pruna cache to the saved path);
- the load_artifacts.py file (for re-saving the tuned config inside vllm/hf cache when loading a smashed model).

## Related Issue

Fixes #(issue number)

## Type of Change

## Checklist

## Additional Notes
Notebook for testing with vllm is available [here](https://drive.google.com/file/d/1XQGs2DuYSokKAZugvcti-bpk3wiH_RWf/view?usp=drive_link). On H100 for qwen3Coder-30B, latency goes from 6.43ms (before tuning) to 5.83 ms (after tuning) while using `vllm`.

### DIFF
--- a/tests/algorithms/testers/moe_kernel_tuner.py
+++ b/tests/algorithms/testers/moe_kernel_tuner.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import json
 from pathlib import Path
 
-from pruna import PrunaModel
+import torch
+
+from pruna import PrunaModel, SmashConfig
 from pruna.algorithms.moe_kernel_tuner import MoeKernelTuner
 
 from .base_tester import AlgorithmTesterBase
@@ -32,21 +34,18 @@ class TestMoeKernelTuner(AlgorithmTesterBase):
 
     def _resolve_hf_cache_config_path(self) -> Path:
         """Read the saved artifact and compute the expected HF cache config path."""
-        from pruna.algorithms.utils.moe_kernel_tuner import save_configs  # noqa: F401
 
-        tuner = MoeKernelTuner()
-        imported_packages = tuner.import_algorithm_packages()
+        imported_packages = MoeKernelTuner().import_algorithm_packages()
 
-        smash_config_path = self._saving_path / "smash_config.json"
-        with open(smash_config_path) as f:
-            smash_cfg = json.load(f)
+        smash_cfg = SmashConfig()
+        smash_cfg.load_from_json(self._saving_path)
 
-        artifact_path = self._saving_path / "moe_kernel_tuner.json"
-        with open(artifact_path) as f:
+        with open(self._saving_path / "moe_kernel_tuner.json") as f:
             artifact = json.load(f)
 
+        dtype = torch.bfloat16 if artifact["dtype"] == "bfloat16" else torch.float16
         dtype_str = imported_packages["_get_config_dtype_str"](
-            artifact["dtype"] == "bfloat16" and __import__("torch").bfloat16 or __import__("torch").float16,
+            dtype,
             use_int8_w8a16=artifact["use_int8_w8a16"],
             use_fp8_w8a8=artifact["use_fp8_w8a8"],
         )
@@ -57,7 +56,7 @@ class TestMoeKernelTuner(AlgorithmTesterBase):
             None,
         )
 
-        hf_cache_base = smash_cfg.get("moe_kernel_tuner_path_to_huggingface_hub_cache", "~")
+        hf_cache_base = smash_cfg["moe_kernel_tuner_path_to_huggingface_hub_cache"]
         return (
             Path(hf_cache_base)
             / ".cache/huggingface/hub/models--RedHatAI--moe/blobs/configs"


llcnt:
bugbot run

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        if model_config is None:
+            raise ValueError(f"Model {model.__class__.__name__} has no config.")
+        nb_experts = model_config.num_experts                # number of experts
+        # number of active experts per token
+        topk = (
+            model_config.num_experts_per_tok
+            if is_moe_lm(model)
+            else model_config.moe_topk[0]
+        )
+        # qwen_moe can use different intermediate size compared to mixtral.
```
sharpenb: What do we mean by intermediate size?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        dtype = smash_config["compute_dtype"]
+        dtype = torch.bfloat16 if dtype == "bfloat16" else torch.float16
+        use_fp8_w8a8 = smash_config["weight_dtype"] == "fp8_w8a8"
+        use_int8_w8a16 = smash_config["weight_dtype"] == "int8_w8a16"
+
+        # (iii) Tune the kernel over a range of batch sizes
+        batch_sizes = [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192]
+
+        # use ray to parallelize the tuning
+        ray.init(ignore_reinit_error=True)
```
sharpenb, line 207: The utilisation of ray makes me think of the optimization agent we worked on.

tests/algorithms/testers/moe_kernel_tuner.py
```diff
@@ -0,0 +1,13 @@
+from pruna.algorithms.moe_kernel_tuner import MoeKernelTuner
+
+from .base_tester import AlgorithmTesterBase
+
+
+class TestMoeKernelTuner(AlgorithmTesterBase):
```
gsprochette, line 14: I'm not sure that AlgorithmTesterBase is really able to test this algorithm, since it's not changing the model. Two options that I would be satisfied with:
- the minimal option would be to manually smash the model and check that the artifacts exist
- the more advanced options would be to implement post a smash hook checking for the artifacts and an other hook after loading to make sure the artifact_loader reloaded the missing config files (e.g. by deleting the artifacts not in the saved_model_dir and checking they are restored after loading).

pyproject.toml
```diff
@@ -138,6 +138,7 @@ dependencies = [
     "imageio-ffmpeg",
     "jaxtyping",
     "peft>=0.17.1",
+    "vllm>=0.11.0",
```
gsprochette: Should vllm be a main dependency? Could we add it into a vllm extra and add it to the list of extras in the CI?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+    group_tags: list[str] = [tags.KERNEL]
+    save_fn: None = None
+    references: dict[str, str] = {
+        "GitHub": "https://github.com/vllm-project/vllm/blob/main/benchmarks/kernels/benchmark_moe.py",
+    }
+    tokenizer_required: bool = False
+    processor_required: bool = False
+    runs_on: list[str] = ["cuda", "accelerate"]
+    dataset_required: bool = False
+    compatible_before: Iterable[str] = [
```
gsprochette, line 51: can we get this list of tags as the result of a function/call on `tags` so it's automatically expanded if we add new tags?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+from pruna.config.hyperparameters import UnconstrainedHyperparameter
+from pruna.config.smash_config import SmashConfigPrefixWrapper
+from pruna.engine.model_checks import is_moe_lm, is_transformers_pipeline_with_moe_lm
+from pruna.engine.save_artifacts import SAVE_ARTIFACTS_FUNCTIONS
+from pruna.logging.logger import pruna_logger
+
+
+class MoeKernelTuner(PrunaAlgorithmBase):
+    """
+    Tune the MoE Triton kernel for the model.
```
gsprochette: Maybe this is a good place to state that MoE stands for Mixture of Experts

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        if is_transformers_pipeline_with_moe_lm(model):
+            return self._apply_to_model_within_transformers_pipeline(model, smash_config)
+
+        imported_packages = self.import_algorithm_packages()
+
+        # (i) Get the MoE parameters
+        model_config = model.config
+        if model_config is None:
+            raise ValueError(f"Model {model.__class__.__name__} has no config.")
+        nb_experts = model_config.num_experts                # number of experts
```
gsprochette: the comment in uninformative and incorrectly formatted, can we remove it?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        Any
+            The untouched model.
+        """
+        if is_transformers_pipeline_with_moe_lm(model):
+            return self._apply_to_model_within_transformers_pipeline(model, smash_config)
+
+        imported_packages = self.import_algorithm_packages()
+
+        # (i) Get the MoE parameters
+        model_config = model.config
```
gsprochette: is `model.config` always defined or is this supposed to be `getattr(model, "config", None)` to check for Noneness later?

src/pruna/algorithms/moe_kernel_tuner.py
gsprochette: For typing reasons, there should be either a protection `nb_experts = getattr(model_config, "num_experts", None")` followed by a check `if nb_experts is None`, or better if available, the check above should check for the type of model_config `isinstance(model_config, BaseConfigClass)` where BaseConfigClass declares the num_experts attribute.

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        if model_config is None:
+            raise ValueError(f"Model {model.__class__.__name__} has no config.")
+        nb_experts = model_config.num_experts                # number of experts
+        # number of active experts per token
+        topk = (
+            model_config.num_experts_per_tok
```
gsprochette: this really makes me believe that there should be a type check for model_config, or a separate function that ensures it has all the expected attributes

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            if is_moe_lm(model)
```
gsprochette: If I understand correctly, at this point there are two options:
- the general case where model matches is_moe_lm
- the exception where model is a HunyuanImage3ForCausalMM

IMO it would be clearer to check for the exception and have the else be for the general case

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            The model to wrap.
+        smash_config : SmashConfigPrefixWrapper
+            The configuration for the application of the algorithm.
+
+        Returns
+        -------
+        Any
+            The untouched model.
+        """
+        if is_transformers_pipeline_with_moe_lm(model):
```
gsprochette, line 172: is there no pipeline for HunyuanImage3ForCausalMM?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+
+        start = time.time()
+        outputs = []
+        configs = []
+
+        # try/except to catch any exceptions that may stuck the ray workers.
+        try:
+            for batch_size in batch_sizes:
+                output = tune.remote(
+                    batch_size,                 # num_tokens
```
gsprochette: comment formatting problem, also
- the `# num_experts per block` doesn't bring information that isn't already in the name
- `# num_tokens` isn't verbose enough for me to understand the information it's trying to bring in
- is it normal that the random seed is fixed?

src/pruna/algorithms/moe_kernel_tuner.py
gsprochette: I agree that this is what the comment should explain instead

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            if is_moe_lm(model)
+            else model_config.moe_topk[0]
+        )
+        # qwen_moe can use different intermediate size compared to mixtral.
+        intermediate_size = (
+            getattr(model_config, "moe_intermediate_size", None)
+            if getattr(model_config, "moe_intermediate_size", None) is not None
+            else model_config.intermediate_size
+        )
+        hidden_size = model_config.hidden_size  # model hidden dim
```
gsprochette: comment is uninformative

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        assert intermediate_size % smash_config["tensor_parallel_size"] == 0, (
```
gsprochette: Should be a ValueError I think: `if intermediate_size % smash_config["tensor_parallel_size"] != 0: raise ValueError(...)`
and for the message can we rephrase it to something like f"Expected tensor_parallel_size to be a divisor of the model's intermediate size {intermediate_size}, but got {smash_config['tensor_parallel_size']}." (if intermediate size could be slightly more explicit it would be even better)

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        ray.init(ignore_reinit_error=True)
+
+        search_space = get_configs_compute_bound(smash_config)
+        pruna_logger.info(f"Start tuning over {len(search_space)} configurations...")
+
+        start = time.time()
+        outputs = []
+        configs = []
+
+        # try/except to catch any exceptions that may stuck the ray workers.
```
gsprochette: this try is not catching any exception, I think the `# Try to shutdown ray ...` is already doing a fantastic job explaining what the try block is there for. Could we remove this one and improve on the comment in the `finally` block? Some suggestion: "Shutdown ray processes on other devices and log any resulting exception as warnings"

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        use_fp8_w8a8: bool,
+        use_int8_w8a16: bool,
+        search_space: list[dict[str, int]],
+        block_quant_shape: list[int],
+        use_deep_gemm: bool,
+        imported_packages: Dict[str, Any],
+        seed: int,
+        num_iters: int,
+    ) -> dict[str, int]:
+    """
```
gsprochette: is the return type not `BenchmarkConfig`?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        # use ray to parallelize the tuning
+        ray.init(ignore_reinit_error=True)
+
+        search_space = get_configs_compute_bound(smash_config)
+        pruna_logger.info(f"Start tuning over {len(search_space)} configurations...")
+
+        start = time.time()
+        outputs = []
+        configs = []
+
```
gsprochette: it would be super helpful for readability to have the type hints for these two lists

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            configs = ray.get(outputs)
+        finally:
+            # Try to shutdown ray, and catch any exceptions.
+            try:
+                ray.shutdown()
+            except Exception as e:
+                pruna_logger.warning(f"Exception during ray.shutdown(): {e}")
+
+        # (iv) Sort the configs by batch size and save the best configs
+        best_configs = {
```
gsprochette: going from configs to best_configs implies that best_configs is a subset of configs which I don't think is the case. Can that be reworked?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            # Some configurations may be invalid and fail to compile.
+            continue
+
+        if kernel_time < best_time:
+            best_time, best_config = kernel_time, config
+
+    now = datetime.now()
+    pruna_logger.info(f"{now.ctime()}] Completed tuning for batch_size={num_tokens}")
+    if best_config is None:
+        raise RuntimeError(
```
gsprochette: Here an error would be raised, and the only way to fix it is to reduce the batch size. However the user does not have control over the batch size, it's hard-coded in apply. I think this should raise a custom NoValidConfigError and catch it in apply. This way only the smaller batch sizes (small enough to run on the available GPU) would run and the user would still get something, instead of needing batch size 8192 to fit.

gsprochette: Also batch_size here is a reference to the name given to that variable outside of this function. Could that get fixed?

src/pruna/algorithms/moe_kernel_tuner.py
gsprochette: if the point is to convert the list into a {batch_size: config} dict, can we have already that structure in outputs and call `configs = dict(zip(outputs.keys(), ray.get(list(outputs.values()))))` or something like that? I'm also try to prevent a case where some batch_size would fail resulting in no config, which could create a misalignment between batch_sizes and configs

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            f"No valid kernel configuration was found for batch_size={num_tokens}. "
+            "All configurations failed (e.g., due to OutOfResources). "
+            "This can happen on GPUs with limited resources. "
+            "Consider reducing your model size, batch size, or tuning search space."
+        )
+    return best_config
+
+
+def sort_config(config: BenchmarkConfig) -> BenchmarkConfig:
```
gsprochette: I don't understand what this function does, I'm suspecting it's a type conversion from the `dict[str, int]` output of tune, to a `BenchmarkConfig` containing the same information? If so the input type or tune's output type is wrong, and the name makes it unclear

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+    -------
+    float
+        The average latency of the kernel.
+    """
+    if not torch.cuda.is_available():
+        raise RuntimeError("CUDA is required for MoeKernelTuner.")
+    # Ray sets CUDA_VISIBLE_DEVICES per worker to the GPU it scheduled
+    torch.cuda.set_device(0)
+    device = torch.device("cuda")
+
```
gsprochette: Is it necessary to set_device then? It seems like this is using the information that the tuning is done in a single device, which is not declared in the docstring and is only availble outside this function

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+    for i in range(num_iters):
+        prepare(i)
+        torch.cuda.synchronize()
+
+        start_event.record()
+        graph.replay()
+        end_event.record()
+        end_event.synchronize()
+        latencies.append(start_event.elapsed_time(end_event))
+    avg = sum(latencies) / (num_iters * 10) * 1000  # us
```
gsprochette: I dont get the comment, is this supposed to be milliseconds or microseconds maybe?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+    path_to_kernel_configs = (
+        pathlib.Path(path_to_huggingface_hub_cache) /
+        ".cache/huggingface/hub/models--RedHatAI--moe/blobs/configs"
+    )
+    pathlib.Path(path_to_kernel_configs).mkdir(exist_ok=True, parents=True)
+    filename_hf = path_to_kernel_configs / filename
+    if not pathlib.Path(filename_hf).exists():
+        pruna_logger.info(f"Writing best config to {filename_hf}...")
+        with open(filename_hf, "w") as f:
+            json.dump({"triton_version": imported_packages["triton"].__version__, **configs}, f, indent=4)
```
gsprochette: this is overriding the possibly-already existing hf config, could there be anything in there that will be missing from this dict? If so maybe we can load the dict, edit it and save it instead of defining a new dict from scratch?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        else:
+            raise RuntimeError("Could not determine installation path for vllm.")
+        path_to_vllm_configs = pathlib.Path(path_where_vllm_is_installed).parent / path_to_vllm_cache
+    pathlib.Path(path_to_vllm_configs).mkdir(exist_ok=True, parents=True)
+    filename_vllm = path_to_vllm_configs / filename
+    if not pathlib.Path(filename_vllm).exists():
+        pruna_logger.info(f"Writing best config to {filename_vllm}...")
+        with open(filename_vllm, "w") as f:
+            json.dump({"triton_version": imported_packages["triton"].__version__, **configs}, f, indent=4)
+            f.write("\n")
```
gsprochette: This block feels redundant with the first one, can we factorize this and or is there specificities I'm not getting?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+        "GROUP_SIZE_M": group_m_range,
+        "num_warps": num_warps_range,
+        "num_stages": num_stage_range,
+    }
+
+    keys, values = zip(*param_ranges.items())
+    for config_values in product(*values):
+        config = dict(zip(keys, config_values))
+        configs.append(config)
+
```
gsprochette:
```suggestion
    configs = [
        dict(zip(param_ranges.keys(), config_values))
        for config_values in product(*param_ranges.values())
    ]
```

src/pruna/engine/load_artifacts.py
```diff
@@ -81,6 +82,59 @@ def load_torch_artifacts(model: Any, model_path: str | Path, smash_config: Smash
...
+    -------
+    Any
+        The loaded MoE model.
+    """
+    from pruna.algorithms.moe_kernel_tuner import MoeKernelTuner, save_configs
+
+    imported_packages = MoeKernelTuner().import_algorithm_packages()
+    save_dir = Path(model_path) / "moe_kernel_tuned_configs"
+    with open(save_dir / "moe_kernel_tuner.json") as f:
+        payload = json.load(f)
```
gsprochette: why payload? could the name be more explicit?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            shard_intermediate_size,
+            dtype,
+            use_fp8_w8a8,
+            use_int8_w8a16,
+            None,
+            smash_config["path_to_huggingface_hub_cache"],
+            smash_config["path_to_vllm_cache"],
+            imported_packages,
+        )
+        # results to be saved for later loading
```
gsprochette: Can the pyload name be improved?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,796 @@
...
+            num_experts=nb_experts,
+            shard_intermediate_size=shard_intermediate_size,
+            dtype="bfloat16" if dtype == torch.bfloat16 else "float16",
+            use_fp8_w8a8=use_fp8_w8a8,
+            use_int8_w8a16=use_int8_w8a16,
+        )
+        # store artifacts in pruna cache for later saving/loading
+        save_dir = smash_config.cache_dir / "moe_kernel_tuned_configs"
+        save_dir.mkdir(parents=True, exist_ok=True)
+        with open(save_dir / "moe_kernel_tuner.json", "w") as f:
```
gsprochette: Why not placing it directly in smash_config.cache_dir if only one file will be contained in moe_kernel_tuned_configs?

src/pruna/algorithms/moe_kernel_tuner.py
llcnt: In moe, the MLPs have a different dimension compared to the "traditional" hidden dimension that hidden states have elsewhere in the model. Hidden states are projected into the "intermediate size" in the MLP and then unprojected back at the output of the MLP to the hidden dimension. I have added a comment to makes this clearer :)

tests/algorithms/testers/moe_kernel_tuner.py
llcnt: I totally agree with you. But to be honest the majority of algos in pruna are not properly tested in the sense that you expect here: i.e. we test if the smash/save/load works without errors, but we don't check (expect for pruning where we always check if the number of weight has decreased, or in prunapro for higgs we check if the quantization was effective) if the algo has been effectively applied to the model itself.
I did some implementation of a post_load_hook(), and a custom execute_load() I can add if you really want :)

src/pruna/algorithms/moe_kernel_tuner.py
llcnt: To the best of my knowledge no. Because (this is very weird but we had this discussion few months ago) huyuan3 is an image gen model that contains itself a `.pipeline` (and this pipeline attribute refers back to the model itself!)

src/pruna/algorithms/moe_kernel_tuner.py
llcnt: yes because each ray worker sets its own CUDA_VISIBLE_DEVICES var ! ;)

tests/algorithms/testers/moe_kernel_tuner.py
gsprochette: I agree for the "we don't check if the algo has been effectively applied" part, but assuming it was we do run inference on the smashed model to test if everything is running.
Here, only artifacts are created, so we are essentially saving, loading and running the model for nothing after that. I would prefer having the hook to at least check something about the algorithm yes.

src/pruna/algorithms/moe_kernel_tuner.py
gsprochette: up

src/pruna/algorithms/moe_kernel_tuner.py
gsprochette: Sorry this was unclear. Now that I think more about it, it would resolve the fact that we need to manually add the reciprocal compatibility in every algorithm. It's not ideal but it's the way smash_config works right now.

Can you revert to the tags lists you had before (sorry) and add the reciprocal compatibility in the other algorithms please?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,441 @@
...
+                except no_valid_config_error:
+                    pruna_logger.warning(
+                        "No valid config for batch_size=%s; skipping (smaller batch sizes may still be used).",
+                        batch_size,
+                    )
+        finally:
+            try:
+                ray.shutdown()
+            except Exception as e:
+                pruna_logger.warning("Exception during ray.shutdown(): %s", e)
```
gsprochette: The agent got rid of fstring :P
```suggestion
                pruna_logger.warning(f"Exception during ray.shutdown(): {e}")
```

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,441 @@
...
+
+        ray.init(ignore_reinit_error=True)
+        search_space = get_configs_compute_bound(smash_config)
+        pruna_logger.info(f"Start tuning over {len(search_space)} configurations...")
+
+        start = time.time()
+        ray_outputs: list[Any] = []
+        tune_kernel = ray.remote(num_gpus=1)(tune_kernel)
+
+        for batch_size in batch_sizes:
```
gsprochette: This for loop used to be in the try block below, unless you're very sure this can't crash at launch, I think it should be back into the try so the finally can always shut ray down.
Thanks for doing the comment changes in the previous comment here :)

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,441 @@
...
+        tuned_config_by_batch_size: dict[int, Any] = {}
+        try:
+            for batch_size, output_ref in zip(batch_sizes, ray_outputs):
+                try:
+                    raw_config = ray.get(output_ref)
+                    tuned_config_by_batch_size[batch_size] = ensure_benchmark_config(raw_config)
+                except no_valid_config_error:
+                    pruna_logger.warning(
+                        "No valid config for batch_size=%s; skipping (smaller batch sizes may still be used).",
+                        batch_size,
```
gsprochette: fstring
```suggestion
                        f"No valid config for {batch_size=}; skipping (smaller batch sizes may still be used)."
```

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -0,0 +1,441 @@
...
+    model : Any
+        The model (used only for type context).
+    model_config : Any
+        The model's config object. Must have num_experts, moe_topk, intermediate_size, hidden_size.
+    tensor_parallel_size : int
+        Tensor parallel size (must divide intermediate_size).
+
+    Returns
+    -------
+    tuple[int, int, int, int]
```
gsprochette, line 372: super clear thanks a lot 🤩

src/pruna/algorithms/utils/moe_kernel_tuner.py
```diff
@@ -0,0 +1,539 @@
...
+
+    path_to_kernel_configs = (
+        pathlib.Path(path_to_huggingface_hub_cache) / ".cache/huggingface/hub/models--RedHatAI--moe/blobs/configs"
+    )
+    path_hf = path_to_kernel_configs / filename
+    write_config_file(
+        path_hf,
+        data,
+        merge_if_exists=True,
+        write_only_if_missing=False,
```
gsprochette: I don't know what I was thinking when I wrote a comment about overriding the already existing hf_config, you were explicitely checking if the file already existed to avoid replacing it. It was the same behavior between path_hf and path_vllm. Sorry.
Let's remove the merge_if_exists and write_only_if_missing args from write_config_file, and always behave as if `merge_if_exists=False, write_only_if_missing=True`

src/pruna/engine/load_artifacts.py
```diff
@@ -102,23 +102,35 @@ def load_moe_kernel_tuner_artifacts(model: Any, model_path: str | Path, smash_co
...
+        # check if the triton version is the same as the one used to tune the kernel
+        triton_version = best_configs_and_hyperparameters["triton_version"]
+        if triton_version != imported_packages["triton"].__version__:
+            msg = (
+                f"Triton version mismatch: {triton_version} != "
+                f"{imported_packages['triton'].__version__}. "
+                "Performance may be degraded or config may be invalid. "
+                "We recommend re-tuning the kernel in your environment."
+            )
+            pruna_logger.info(msg)
```
gsprochette, line 124: yes !

src/pruna/algorithms/moe_kernel_tuner.py
llcnt: oh yes, good catch :)

tests/algorithms/testers/moe_kernel_tuner.py
llcnt: I have added a `post_load_hook` function to check artifacts are correctly restored :)

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -294,11 +303,12 @@ def import_algorithm_packages(self) -> dict[str, Any]:
         import vllm.envs as envs
         import vllm.model_executor.layers.fused_moe.fused_moe as fused_moe
         import vllm.platforms as vllm_platforms
-        from vllm.model_executor.layers.fused_moe import override_config
+        from vllm.model_executor.layers.fused_moe import fused_topk, override_config
```
gsprochette, line 306: This fused_topk is the part you adapted for the last vllm versions right? Or is it something new?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -336,6 +348,16 @@ def extract_hunyuan_dimensions(
     """
     Extract MoE dimensions for HunyuanImage3ForCausalMM.
 
+    In MoE architectures each expert is an MLP that projects hidden states from
```
gsprochette, line 351: Super clear, thanks a lot 🤓

src/pruna/algorithms/reduce_noe.py
```diff
@@ -46,7 +46,7 @@ class ReduceNOE(PrunaAlgorithmBase):
     dataset_required: bool = False
     runs_on: list[str] = ["cuda", "accelerate"]
     save_fn: None = None
-    compatible_after: Iterable[str] = ["*"]
+    compatible_after: Iterable[str] = ["*", "moe_kernel_tuner"]
```
gsprochette, line 49: Ah right, you can only tune the experts after removing some of them. Good catch 👍

tests/algorithms/testers/base_tester.py
```diff
@@ -132,7 +136,7 @@ def execute_load(self) -> PrunaModel:
         """Load the smashed model."""
         model = PrunaModel.from_pretrained(str(self._saving_path))
         assert isinstance(model, PrunaModel)
-        self.post_smash_hook(model)
+        self.post_load_hook(model)
```
gsprochette, line 139: makes sense yes

tests/algorithms/testers/moe_kernel_tuner.py
```diff
@@ -11,3 +17,67 @@ class TestMoeKernelTuner(AlgorithmTesterBase):
...
+        assert artifact.exists(), f"Expected artifact at {artifact} after smash"
+
+    def _resolve_hf_cache_config_path(self) -> Path:
+        """Read the saved artifact and compute the expected HF cache config path."""
+        from pruna.algorithms.utils.moe_kernel_tuner import save_configs  # noqa: F401
+
+        tuner = MoeKernelTuner()
+        imported_packages = tuner.import_algorithm_packages()
+
+        smash_config_path = self._saving_path / "smash_config.json"
```
gsprochette: I think for these 3 lines you're just looking for `SmashConfig.load_from_json(self._saving_path)`, no?

tests/algorithms/testers/moe_kernel_tuner.py
```diff
@@ -11,3 +17,67 @@ class TestMoeKernelTuner(AlgorithmTesterBase):
...
+        with open(smash_config_path) as f:
+            smash_cfg = json.load(f)
+
+        artifact_path = self._saving_path / "moe_kernel_tuner.json"
+        with open(artifact_path) as f:
+            artifact = json.load(f)
+
+        dtype_str = imported_packages["_get_config_dtype_str"](
+            artifact["dtype"] == "bfloat16" and __import__("torch").bfloat16 or __import__("torch").float16,
```
gsprochette: why use `__import__("torch")` instead of simply importing torch and using it? It seems clearer...

tests/algorithms/testers/moe_kernel_tuner.py
```diff
@@ -11,3 +17,67 @@ class TestMoeKernelTuner(AlgorithmTesterBase):
...
     metrics = ["perplexity"]
+    # for faster testing
+    hyperparameters = {
+        "moe_kernel_tuner_num_iters": 1,
+        "moe_kernel_tuner_block_size_m_max": 4,
+        "moe_kernel_tuner_block_size_n_max": 5,
+        "moe_kernel_tuner_block_size_k_max": 6,
+    }
+
+    def post_smash_hook(self, model: PrunaModel) -> None:
```
gsprochette, line 30: Thanks a lot for adding these post_smash and post_load hooks :)

src/pruna/algorithms/moe_kernel_tuner.py
llcnt: Yep, exactly :)

tests/algorithms/testers/moe_kernel_tuner.py
llcnt: yes, makes sense !

PR #492, opened 2026-01-14
refactor: formalize artifact saving

gsprochette:
## Description
Formalize artifact saving and expose it to more algorithms, and add artifact hooks into Smash config
- Generalized artifact saving/loading: Extracted and formalized the artifact-saving logic so it is no longer tied specifically to the Torch Compile workflow. Other Pruna algorithms can now register and reuse the same standardized artifact save/load utilities.
- Config-driven hooks: Extended the Smash configuration to include save_artifacts_fns and load_artifacts_fns, working in a way similar to save_fns and load_fns
- Improved composability: This structure makes it easy to save artifacts (such as parameters estimated from a warmup, or checkpoints for finetuning or distillation).

## Type of Change

## Checklist

### DIFF
--- a/src/pruna/engine/load_artifacts.py
+++ b/src/pruna/engine/load_artifacts.py
@@ -105,6 +105,9 @@ class LOAD_ARTIFACTS_FUNCTIONS(Enum):  # noqa: N801
         The type of the enum.
     start : int
         The start index for auto-numbering enum values.
+    boundary : enum.FlagBoundary or None
+        Boundary handling mode used by the Enum functional API for Flag and
+        IntFlag enums.
 
     Examples
     --------
--- a/src/pruna/engine/save_artifacts.py
+++ b/src/pruna/engine/save_artifacts.py
@@ -110,6 +110,9 @@ class SAVE_ARTIFACTS_FUNCTIONS(Enum):  # noqa: N801
         The type of the enum.
     start : int
         The start index for auto-numbering enum values.
+    boundary : enum.FlagBoundary or None
+        Boundary handling mode used by the Enum functional API for Flag and
+        IntFlag enums.
 
     Examples
     --------


src/pruna/engine/save.py
```diff
@@ -61,9 +62,10 @@ def save_pruna_model(model: Any, model_path: str | Path, smash_config: SmashConf
     if not model_path.exists():
         model_path.mkdir(parents=True, exist_ok=True)
 
-    if SAVE_FUNCTIONS.torch_artifacts.name in smash_config.save_fns:
-        save_torch_artifacts(model, model_path, smash_config)
-        smash_config.save_fns.remove(SAVE_FUNCTIONS.torch_artifacts.name)
+    # Backward compatibility with torch artifacts save function
```
simlang: i see why this is necessary, but it would be nice to remove this as soon as possible again - should we add a deprecation warning or something, and then actually remove it again at some point?

src/pruna/engine/load.py
```diff
@@ -59,16 +60,16 @@ def load_pruna_model(model_path: str | Path, **kwargs) -> tuple[Any, SmashConfig
     # since the model was just loaded from a file, we do not need to prepare saving anymore
     smash_config._prepare_saving = False
 
+    # Backward compatibility with torch artifacts save function
```
llcnt: Cool to think about it, I would have completely forgotten ahah

src/pruna/engine/load_artifacts.py
```diff
@@ -0,0 +1,120 @@
...
+        # Only handle artifact loaders we explicitly know about here.
+        if fn_name not in LOAD_ARTIFACTS_FUNCTIONS.__members__:
+            continue
+
+        LOAD_ARTIFACTS_FUNCTIONS[fn_name](model, model_path, smash_config)
+
+
+def load_torch_artifacts(model: PrunaModel, model_path: str | Path, smash_config: SmashConfig) -> None:
+    """
+    Load a torch artifacts from the given model path.
```
simlang:
```suggestion
    Load torch artifacts from the given model path.
```

src/pruna/engine/load.py
```diff
@@ -56,6 +57,8 @@ def load_pruna_model(model_path: str | Path, **kwargs) -> tuple[Any, SmashConfig
     """
     smash_config = SmashConfig()
     smash_config.load_from_json(model_path)
+    if "torch_artifacts" in smash_config.load_fns:
+        _convert_to_artifact(smash_config, model_path)
```
gsprochette, line 61: @simlang in response to your comment I removed the logic in save.py and added this function:
- update both the save and load fns
- save the smash config so that it's fixed for when we remove this logic in a future version after some deprecation time
- log a warning saying it was updated and asking to warn any remote provider (because those would never receive the warning nor have the update automatically saved)

Let me know if that's to your liking

PR #502, opened 2026-01-22
feat: add benchmark support to PrunaDataModule and implement PartiPrompts

davidberenstein1957:
## Summary

Adds a benchmark system to PrunaDataModule and implements PartiPrompts plus 7 additional benchmarks for text-to-image and image editing evaluation.

## Benchmarks Included

| Benchmark | Task type | Description |
|-----------|-----------|-------------|
| **PartiPrompts** | text_to_image | 1,600+ prompts across 12 categories and 11 challenge aspects |
| **GenEval** | text_to_image | Fine-grained compositional evaluation (object co-occurrence, positioning, counting, color binding) |
| **HPS** | text_to_image | Human preference score (HPDv2) with 4 categories: anime, concept-art, paintings, photo |
| **LongTextBench** | text_to_image | Long, detail-rich prompts (~285 tokens) for compositional reasoning |
| **OneIG** | text_to_image | Text rendering and image-text alignment (anime, portrait, object) |
| **DPG** | text_to_image | Descriptive prompt generation (entity, attribute, relation, global, other) |
| **ImgEdit** | image_edit | 8 edit types: replace, add, remove, adjust, extract, style, background, compose |
| **GEditBench** | image_edit | 11 task types for fine-grained editing (background_change, color_alter, etc.) |

## Changes

- **Benchmark system**: `BenchmarkInfo` dataclass, `benchmark_info` registry, `BENCHMARK_CATEGORY_CONFIG` for category/subset filtering
- **Shared helper**: `_prepare_test_only_prompt_dataset` for test-only prompt datasets
- **Discovery**: `list_benchmarks()` and `get_benchmark_info()` for filtering by task type
- **Category filtering**: All benchmarks support `category` or `subset` via `PrunaDataModule.from_string(name, category=...)`
- **Auxiliaries**: `prompt_with_auxiliaries_collate` preserves metadata (Category, Challenge, questions, text_content, etc.) for evaluation

## Usage

```python
from pruna.data import PrunaDataModule

# Load PartiPrompts with category filter
dm = PrunaDataModule.from_string("PartiPrompts", category="Animals")

# Load HPS with anime category
dm = PrunaDataModule.from_string("HPS", category="anime")

# Load ImgEdit with specific edit type
dm = PrunaDataModule.from_string("ImgEdit", category="replace")

# List all text-to-image benchmarks
from pruna.data import list_benchmarks
list_benchmarks(task_type="text_to_image")
```

## Merged PRs

- #505 HPS
- #506 LongTextBench
- #507 GenEval
- #508 OneIG
- #509 DPG
- #517 ImgEdit
- #518 GEditBench

## Test Plan

- `test_benchmark_category_filter` – category filtering for all benchmarks
- `test_dm_from_string` – loading each dataset via `from_string`
- `test_long_text_bench_auxiliaries`, `test_oneig_text_rendering_auxiliaries` – auxiliaries preserved

### DIFF
--- a/src/pruna/data/utils.py
+++ b/src/pruna/data/utils.py
@@ -244,7 +244,8 @@ def stratify_dataset(
     fraction: float = 1.0,
     seed: int | None = None,
 ) -> Dataset:
-    """Stratify the dataset to a specific size via optional shuffled sampling.
+    """
+    Stratify the dataset to a specific size via optional shuffled sampling.
 
     Parameters
     ----------
--- a/src/pruna/evaluation/metrics/registry.py
+++ b/src/pruna/evaluation/metrics/registry.py
@@ -91,7 +91,19 @@ class MetricRegistry:
 
     @classmethod
     def has_metric(cls, name: str) -> bool:
-        """Return True if a metric with this name is registered."""
+        """
+        Return True if a metric with this name is registered.
+
+        Parameters
+        ----------
+        name : str
+            Name of the metric to look up.
+
+        Returns
+        -------
+        bool
+            True if the metric is registered, False otherwise.
+        """
         return name in cls._registry
 
     @classmethod


src/pruna/data/datasets/prompt.py
```diff
@@ -51,13 +55,30 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
     Tuple[Dataset, Dataset, Dataset]
-        The Parti Prompts dataset.
+        The Parti Prompts dataset (dummy train, dummy val, test).
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is not None:
+        ds = ds.filter(lambda x: x["Category"] == category or x["Challenge"] == category)
+
+    ds = ds.shuffle(seed=seed)
```
begumcig: since we are only creating test set how do you feel about not shuffling the data?

src/pruna/data/__init__.py
```diff
@@ -97,8 +98,164 @@
...
+benchmark_info: dict[str, BenchmarkInfo] = {
+    "PartiPrompts": BenchmarkInfo(
+        name="parti_prompts",
+        display_name="Parti Prompts",
+        description=(
+            "Over 1,600 diverse English prompts across 12 categories with 11 challenge aspects "
+            "ranging from basic to complex, enabling comprehensive assessment of model capabilities "
+            "across different domains and difficulty levels."
+        ),
+        metrics=["arniqa", "clip_score", "clipiqa", "sharpness"],
```
begumcig: I don't think we should tightly couple benchmarking datasets with metrics. I think benchmarks should have their datasets available as PrunaDataModules, and the metrics for the Benchmarks should be Pruna Metrics. This way we can give the user the flexibility to use whichever dataset with whichever metric they choose, how do you feel?

begumcig (changes requested):
Really nice David! I think we shouldn't tightly couple metrics with datasets under benchmark, but otherwise it all looks good to me! Thank you!

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,182 @@
...
+
+    name: str
+    description: str
+    metrics: list[str]
+    task_type: str
+    subsets: list[str] = field(default_factory=list)
+
+    def __post_init__(self) -> None:
+        """Populate subsets from setup function's Literal when a matching lookup key exists."""
+        lookup_key = self.name.replace(" ", "")
```
simlang: up to you, but since you also use this later, should we introduce a computed property lookup_key which does the replace? just to make sure that everywhere it's the same

src/pruna/data/datasets/prompt.py
```diff
@@ -51,13 +84,34 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
     Returns
     -------
     Tuple[Dataset, Dataset, Dataset]
-        The Parti Prompts dataset.
+        The Parti Prompts dataset (dummy train, dummy val, test).
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is not None:
+        if isinstance(category, list):
```
simlang: completely personal taste, but if category is not a list, just create [category], so below we only have one filter. to me this feels cleaner

src/pruna/data/utils.py
```diff
@@ -217,6 +262,35 @@ def stratify_dataset(dataset: Dataset, sample_size: int, seed: int = 42) -> Data
...
+        The seed for shuffling.
+    dataset_name : str
+        Name for logging.
+
+    Returns
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        Dummy train, dummy val, and test datasets.
+    """
+    ds = ds.shuffle(seed=seed)
```
simlang: afaik test datasets should not be shuffled, if you need it somewhere maybe add a flag if should be shuffled, but i would always expect an unshuffled test dataset

src/pruna/data/__init__.py
```diff
@@ -97,8 +105,223 @@
...
+    "GEditBench": (setup_gedit_dataset, "prompt_with_auxiliaries_collate", {}),
+    "OneIG": (setup_oneig_dataset, "prompt_with_auxiliaries_collate", {}),
+    "DPG": (setup_dpg_dataset, "prompt_with_auxiliaries_collate", {}),
     "TinyIMDB": (setup_tiny_imdb_dataset, "text_generation_collate", {}),
     "VBench": (setup_vbench_dataset, "prompt_with_auxiliaries_collate", {}),
 }
+
+
+@dataclass
+class Benchmark:
```
simlang: i think a leftover from the merging of PRs, but already defined in evaluation/benchmarks/__init__.py

src/pruna/data/datasets/prompt.py
```diff
@@ -51,16 +161,202 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    Returns
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The HPS dataset (dummy train, dummy val, test).
+    """
+    import json
+
+    from huggingface_hub import hf_hub_download
+
+    categories_to_load = (
```
simlang, line 322: personal opinion, but i would do a "proper" if else on category is None just for readability

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +379,400 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    import json
+
+    import requests
+
+    ds = load_dataset("OneIG-Bench/OneIG-Bench", "OneIG-Bench")["train"]  # type: ignore[index]
+
+    questions_by_id: dict[str, dict] = {}
+    url = "https://raw.githubusercontent.com/OneIG-Bench/OneIG-Benchmark/main/benchmark/alignment_questions.json"
+    response = requests.get(url)
+    if response.status_code == 200:
```
simlang: is there are reason that compared to the previous similar requests we have to check the status_code here?

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +379,400 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+
+def _load_oneig_generic(
+    seed: int,
+    category_filter: str | None = None,
+    class_filter: str | None = None,
+    config: str = "OneIG-Bench",
+) -> Dataset:
+    """Load OneIG data for Knowledge_Reasoning, Multilingualism, or any category without alignment questions."""
+    ds = load_dataset("OneIG-Bench/OneIG-Bench", config)[  # type: ignore[index]
+        "train"
```
simlang: all in one line, to stay consistent with the others before

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +379,400 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+                "class": row_class,
+                "questions": q_info.get("questions", []),
+                "dependencies": q_info.get("dependencies", []),
+            }
+        )
+
+    return Dataset.from_list(records).shuffle(seed=seed)
+
+
+ONEIG_DATASET_CATEGORIES = frozenset(
```
simlang: this feels somehow redundant as these are a subset of previously defined OneIG Categories

.github/actions/setup-uv-project/action.yml
```diff
@@ -11,5 +11,13 @@ runs:
         enable-cache: true
         github-token: ${{ github.token }}
 
+    - name: Cache uv virtual environment
```
simlang: imo the actions should be a separate PR as they have nothing to do with the benchmarking stuff
I also feel lik i'm the wrong reviewer for the action changes 👀

simlang (changes requested):

src/pruna/data/datasets/prompt.py
```diff
@@ -51,13 +55,30 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
     Returns
     -------
     Tuple[Dataset, Dataset, Dataset]
-        The Parti Prompts dataset.
+        The Parti Prompts dataset (dummy train, dummy val, test).
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is not None:
+        ds = ds.filter(lambda x: x["Category"] == category or x["Challenge"] == category)
```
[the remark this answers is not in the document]
davidberenstein1957: ✅ **FIXED**: Category filter now handles list correctly via `categories = [category] if not isinstance(category, list) else category`.

tests/data/test_datamodule.py
```diff
@@ -1,10 +1,10 @@
 from typing import Any, Callable
 
 import pytest
-from transformers import AutoTokenizer
-from datasets import Dataset
-from torch.utils.data import TensorDataset
 import torch
+from transformers import AutoTokenizer
+
+from pruna.data import BenchmarkInfo, benchmark_info
```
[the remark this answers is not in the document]
davidberenstein1957: ✅ **RESOLVED**: test_datamodule.py imports `base_datasets` and `get_literal_values_from_param` only—no BenchmarkInfo or benchmark_info.

src/pruna/data/__init__.py
```diff
@@ -97,8 +98,164 @@
...
+
+    Raises
+    ------
+    KeyError
+        If benchmark name is not found.
+    """
+    if name not in benchmark_info:
+        available = ", ".join(benchmark_info.keys())
+        raise KeyError(f"Benchmark '{name}' not found. Available: {available}")
+    return benchmark_info[name]
```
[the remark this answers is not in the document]
davidberenstein1957: ✅ **RESOLVED**: BenchmarkRegistry in evaluation/benchmarks provides list/get. No list_benchmarks or get_benchmark_info in current codebase.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **FIXED**: Removed shuffle from `_prepare_test_only_prompt_dataset`. Test datasets are no longer shuffled.

src/pruna/data/__init__.py
davidberenstein1957: ℹ️ **INTENTIONAL**: Architectural discussion—decoupling benchmarks from metrics would be a larger design change. Deferring to a follow-up PR.

src/pruna/evaluation/benchmarks/__init__.py
davidberenstein1957: ✅ **RESOLVED**: `BenchmarkRegistry.Benchmark` already has `lookup_key` computed property.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **RESOLVED**: Using `categories = [category] if not isinstance(category, list) else category` for single filter.

src/pruna/data/utils.py
davidberenstein1957: ✅ **FIXED**: Removed shuffle from `_prepare_test_only_prompt_dataset`.

src/pruna/data/__init__.py
davidberenstein1957: ✅ **FIXED**: Removed duplicate Benchmark/benchmark_info from data/__init__.py.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **FIXED**: Using explicit if/else for category is None.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **FIXED**: Added status_code check to ImgEdit for consistency with OneIG.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **FIXED**: Put load_dataset on one line in _load_oneig_generic.

src/pruna/data/datasets/prompt.py
davidberenstein1957: ✅ **FIXED**: alignment_cats now derived from ONEIG_DATASET_CATEGORIES.

.github/actions/setup-uv-project/action.yml
davidberenstein1957: I'll move them over to a seperate PR.

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+
+    _registry: list[Benchmark] = [
+        Benchmark(
+            name="Parti Prompts",
+            description=(
+                "Holistic benchmark from Google Research with over 1,600 English prompts across 12 categories "
+                "and 11 challenge aspects. Evaluates text-to-image models on abstract thinking, world knowledge, "
+                "perspectives, and symbol rendering from basic to complex compositions."
+            ),
+            metrics=[
```
begumcig: are these the official metrics defined for the PartiPrompts or something we defined? I think if the official documentation lists / defines some metrics and we are not supporting them we should leave this empty.

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+            ],
+            task_type="text_to_image",
+        ),
+        Benchmark(
+            name="DrawBench",
+            description=(
+                "Comprehensive benchmark from the Imagen team for rigorous evaluation of text-to-image models. "
+                "Enables side-by-side comparison on sample quality and image-text alignment with human raters."
+            ),
+            metrics=[
```
begumcig: same comment as above

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+        ),
+        Benchmark(
+            name="GenAI Bench",
+            description=(
+                "1,600 prompts from professional designers for compositional text-to-visual generation. "
+                "Covers basic skills (scene, attributes, spatial relationships) to advanced reasoning "
+                "(counting, comparison, logic/negation) with over 24k human ratings."
+            ),
+            metrics=[
+                "clip_score",
```
begumcig: same as above, just pinging so it's easier for you to find the occurences, sorry 🥹

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+            task_type="text_to_image",
+        ),
+        Benchmark(
+            name="VBench",
+            description=(
+                "Comprehensive benchmark suite for video generative models. Decomposes video quality into "
+                "16 disentangled dimensions: temporal flickering, motion smoothness, subject consistency, "
+                "spatial relationship, color, aesthetic quality, and more."
+            ),
+            metrics=["clip_score"],
```
begumcig: again the same comment

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+        missing = [m for m in benchmark.metrics if m not in MetricRegistry._registry]
+        if missing:
+            raise ValueError(
+                f"Benchmark '{name}' references metrics not in MetricRegistry: {missing}. "
+                f"Available metrics: {list(MetricRegistry._registry.keys())}"
+            )
+        if benchmark.lookup_key not in base_datasets:
+            available = ", ".join(base_datasets.keys())
+            raise ValueError(
+                f"Benchmark '{name}' name '{benchmark.name}' does not align with base_datasets. "
```
begumcig: I am a bit confused about this message, why do we print "name" here, isn't it the lookup key from passed to the `_validate` function? Perhaps something like `f"Benchmark '{benchmark.name}' (lookup key '{benchmark.lookup_key}') is not in base_datasets. Available: {available}"` would be clearer?

src/pruna/data/datasets/prompt.py
```diff
@@ -12,12 +12,132 @@
...
+from typing import Literal, Tuple, get_args
 
 from datasets import Dataset, load_dataset
 
+from pruna.data.utils import define_sample_size_for_dataset
 from pruna.logging.logger import pruna_logger
 
+GenEvalCategory = Literal["single_object", "two_object", "counting", "colors", "position", "color_attr"]
+HPSCategory = Literal["anime", "concept-art", "paintings", "photo"]
+OneIGCategory = Literal[
```
begumcig, line 24: I might be mixing things up, so please correct me, but in OneIG, I thought we had main categories and then subcategories within them (the classes). In this list, I’m seeing both categories and classes together, and I just want to make sure I’m understanding it correctly. How will this be handled when computing the metrics?

src/pruna/data/datasets/prompt.py
```diff
@@ -51,18 +177,210 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is None:
+        pass
+    else:
+        categories = [category] if not isinstance(category, list) else category
+        ds = ds.filter(lambda x: x["Category"] in categories or x["Challenge"] in categories)
+
+    test_sample_size = define_sample_size_for_dataset(ds, fraction, test_sample_size)
```
begumcig: This is completely optional, so please feel free to ignore it if it’s more trouble than it’s worth but I think we already have a utility function in pruna.data.utils called `stratify_dataset` that handles this sizing issue. I know it takes a seed as well, so maybe we could make the seed optional and still reuse the function?

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The OneIG dataset (dummy train, dummy val, test).
+    """
+    import json
+
+    import requests
+
+    resp = requests.get(
+        "https://raw.githubusercontent.com/OneIG-Bench/OneIG-Benchmark/main/benchmark/alignment_questions.json"
```
begumcig: Sorry this is probably dumb but are we sure that this exists? Using OneIG recently myself, can't recall this in the repo :D

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    ds_raw = load_dataset("OneIG-Bench/OneIG-Bench", "OneIG-Bench")["train"]  # type: ignore[index]
+    records = [_to_oneig_record(dict(row), questions_by_id) for row in ds_raw]
+    ds = Dataset.from_list(records).shuffle(seed=seed)
+
+    if category is None:
+        pass
+    else:
+        categories = [category] if not isinstance(category, list) else category
+        ds = ds.filter(
+            lambda x: (x.get("category") in categories or x.get("class") in categories or x.get("subset") in categories)
```
begumcig: I think I made a comment about this also above but what happens if someone gives a class and the category that has that class at the same time?

src/pruna/data/datasets/prompt.py
```diff
@@ -51,18 +177,210 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+        )
+
+    ds = Dataset.from_list(records)
+    test_sample_size = define_sample_size_for_dataset(ds, fraction, test_sample_size)
+    ds = ds.select(range(min(test_sample_size, len(ds))))
+    pruna_logger.info("GenEval is a test-only dataset. Do not use it for training or validation.")
+    return ds.select([0]), ds.select([0]), ds
+
+
+def setup_hps_dataset(
```
begumcig, line 288: this is me being super nitpicky but I believe HPS is the benchmarking model and HPD is the name of the dataset

src/pruna/data/datasets/prompt.py
```diff
@@ -41,7 +161,13 @@ def setup_drawbench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
     return ds.select([0]), ds.select([0]), ds
 
 
-def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
+def setup_parti_prompts_dataset(
+    seed: int,
```
begumcig, line 162: I believe all the load functions in this file have a seed parameter, even though we’re not actually using it. I assume that’s because it’s passed down from the PrunaDataModule?

I’m just a little concerned that users might pass different seeds expecting it to have an effect. Would it make sense to at least print a message clarifying that no shuffling is happening?

src/pruna/data/datasets/prompt.py
begumcig, line 288: this is me being super nitpicky but I believe HPS is the benchmarking model and HPD is the name of the dataset

src/pruna/data/datasets/prompt.py
```diff
@@ -51,18 +177,210 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
-        The Parti Prompts dataset.
+        The Parti Prompts dataset (dummy train, dummy val, test).
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is None:
+        pass
+    else:
+        categories = [category] if not isinstance(category, list) else category
+        ds = ds.filter(lambda x: x["Category"] in categories or x["Challenge"] in categories)
```
begumcig, line 195: Just flagging this because I know PartiPrompts have overlap of prompts btw Categories and Challenges. Incase if it's important for metric computation later on

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+
+def setup_dpg_dataset(
+    seed: int,
+    fraction: float = 1.0,
+    train_sample_size: int | None = None,
+    test_sample_size: int | None = None,
+    category: DPGCategory | list[DPGCategory] | None = None,
+) -> Tuple[Dataset, Dataset, Dataset]:
+    """
+    Setup the DPG (Descriptive Prompt Generation) benchmark dataset.
```
begumcig: Again this is me being nitpicky but I believe the Benchmark is called (Dense Prompt Graph) Benchmark

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+        """Key for base_datasets lookup (name with spaces removed)."""
+        return self.name.replace(" ", "")
+
+    def __post_init__(self) -> None:
+        """Populate category and categories from setup function's Literal."""
+        if self.lookup_key in base_datasets:
+            setup_fn = base_datasets[self.lookup_key][0]
+            literal_values = get_literal_values_from_param(setup_fn, "category")
+            self.categories = literal_values if literal_values is not None else []
+            self.category = self.categories[0] if self.categories else None
```
begumcig: Can you explain why we have categories and category at the same time?

src/pruna/evaluation/benchmarks/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+
+from dataclasses import dataclass, field
+
+from pruna.data import base_datasets
+from pruna.data.utils import get_literal_values_from_param
+from pruna.evaluation.metrics import MetricRegistry
+
+
+@dataclass
+class Benchmark:
```
begumcig: I really like this Benchmark class! I don't know if this is planned for future but maybe we could add something like EvaluationAgent.from_benchmark(name, model) (or similar) to automatically hook up the dataset + metrics?

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+
+        if categories is not None and row_category not in categories:
+            continue
+
+        key = (row.get("text", ""), row_category)
+        q = row.get("question_natural_language", "")
+        if q and q not in grouped[key]:
+            grouped[key].append(q)
+
+    records = [{"text": text, "category_broad": cat, "questions": qs} for (text, cat), qs in grouped.items()]
```
begumcig: maybe we should also use category here instead of category_broad for consistency?

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+
+    categories = [category] if category is not None and not isinstance(category, list) else category
+    if categories is not None:
+        hf_types = [task_type_map.get(c, c) for c in categories]
+        ds = ds.filter(lambda x: x["task_type"] in hf_types)
+
+    records = []
+    for row in ds:
+        task_type = row.get("task_type", "")
+        category_name = {v: k for k, v in task_type_map.items()}.get(task_type, task_type)
```
begumcig: nitpicky suggestion: since category_name doesn't depend on anything from row, perhaps we can compute it outside of the loop to prevent computing it again for each row?

src/pruna/data/datasets/prompt.py
```diff
@@ -51,18 +177,210 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    -------
+    Tuple[Dataset, Dataset, Dataset]
+        The GenEval dataset (dummy train, dummy val, test).
+    """
+    import json
+
+    import requests
+
+    url = "https://raw.githubusercontent.com/djghosh13/geneval/d927da8e42fde2b1b5cd743da4df5ff83c1654ff/prompts/evaluation_metadata.jsonl"
+    response = requests.get(url)
```
begumcig: perhaps we should add a timeout here, i am a little worried that it could stall indefinitely

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    Tuple[Dataset, Dataset, Dataset]
+        The ImgEdit dataset (dummy train, dummy val, test).
+    """
+    import json
+
+    import requests
+
+    instructions_url = "https://raw.githubusercontent.com/PKU-YuanGroup/ImgEdit/b3eb8e74d7cd1fd0ce5341eaf9254744a8ab4c0b/Benchmark/Basic/basic_edit.json"
+    judge_prompts_url = "https://raw.githubusercontent.com/PKU-YuanGroup/ImgEdit/c14480ac5e7b622e08cd8c46f96624a48eb9ab46/Benchmark/Basic/prompts.json"
+    response_instructions = requests.get(instructions_url)
```
begumcig: again suggestion to add timeout here

src/pruna/data/datasets/prompt.py
begumcig: Also, perhaps it's a better idea to use a certain commit hash and not the main, since it can change?

begumcig: Please don't hate me but also a timeout would be great 🥹

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+        The DPG dataset (dummy train, dummy val, test).
+    """
+    import csv
+    import io
+    from collections import defaultdict
+
+    import requests
+
+    url = "https://raw.githubusercontent.com/TencentQQGYLab/ELLA/main/dpg_bench/dpg_bench.csv"
+    response = requests.get(url)
```
begumcig: tiny comment also here for timeout 🥹

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    resp.raise_for_status()
+    questions_by_id = {q["id"]: q for q in json.loads(resp.text)}
+
+    ds_raw = load_dataset("OneIG-Bench/OneIG-Bench", "OneIG-Bench")["train"]  # type: ignore[index]
+    records = [_to_oneig_record(dict(row), questions_by_id) for row in ds_raw]
+    ds = Dataset.from_list(records).shuffle(seed=seed)
+
+    if category is None:
+        pass
+    else:
```
begumcig: super nitpicky comment, perhaps we could change lines 523-525 with `if category is not None`

src/pruna/data/datasets/prompt.py
```diff
@@ -51,18 +177,210 @@ def setup_parti_prompts_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
 
     Returns
     -------
     Tuple[Dataset, Dataset, Dataset]
-        The Parti Prompts dataset.
+        The Parti Prompts dataset (dummy train, dummy val, test).
     """
     ds = load_dataset("nateraw/parti-prompts")["train"]  # type: ignore[index]
+
+    if category is None:
```
begumcig: nitpicky comment: perhaps we could change lines 196-200 with `if category is not None:`

src/pruna/data/datasets/prompt.py
```diff
@@ -83,3 +401,268 @@ def setup_genai_bench_dataset(seed: int) -> Tuple[Dataset, Dataset, Dataset]:
...
+    import json
+
+    import requests
+
+    instructions_url = "https://raw.githubusercontent.com/PKU-YuanGroup/ImgEdit/b3eb8e74d7cd1fd0ce5341eaf9254744a8ab4c0b/Benchmark/Basic/basic_edit.json"
+    judge_prompts_url = "https://raw.githubusercontent.com/PKU-YuanGroup/ImgEdit/c14480ac5e7b622e08cd8c46f96624a48eb9ab46/Benchmark/Basic/prompts.json"
+    response_instructions = requests.get(instructions_url)
+    response_judge_prompts = requests.get(judge_prompts_url)
+    response_instructions.raise_for_status()
+    response_judge_prompts.raise_for_status()
```
begumcig, line 432: oh i really like this raise_for_status() guard actually, can we add this to other instances of get()?

src/pruna/data/datasets/prompt.py
davidberenstein1957: Wow, I did not know. How inconvenient haha

davidberenstein1957: how inconvenient, I'lll change it :)

src/pruna/evaluation/benchmarks/__init__.py
davidberenstein1957: I kept a single category value

src/pruna/data/datasets/prompt.py
davidberenstein1957: Yes, but I believe each prompt belongs to both a category and a challenge but I feel the filter is intended just to focus specific subsets, in case we want to dedicate a metric to, let's say categories :) 
image

src/pruna/data/datasets/prompt.py
davidberenstein1957: So, the idea is that you can pass either one, and that we will return the main or the sub-category, after filtering.

src/pruna/data/datasets/prompt.py
davidberenstein1957: It basically is an OR filter.

src/pruna/evaluation/benchmarks/__init__.py
davidberenstein1957: I've double checked them

src/pruna/data/utils.py
```diff
@@ -190,58 +238,49 @@ def recover_text_from_dataloader(dataloader: DataLoader, tokenizer: Any) -> list
     return texts
 
 
-def stratify_dataset(dataset: Dataset, sample_size: int, seed: int = 42) -> Dataset:
+def stratify_dataset(
```
simlang, line 241: since we use this function now in every test dataset, we have to make seeding (which leads to shuffling) optional, as test sets should never be shuffled, as already mentioned in other comments

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -43,12 +52,19 @@ class EvaluationAgent:
     task : Task, optional
         Configuration object that defines how to evaluate the model.
     request : str | List[str | BaseMetric | StatefulMetric], optional
-        The user request to evaluate. Required if task is not provided.
+        The user request to evaluate. Required if task/benchmark not provided.
     datamodule : PrunaDataModule, optional
-        The dataloader to use for the evaluation. Required if task is not provided.
+        The dataloader to use for the evaluation. Required if task/benchmark not provided.
+    benchmark : str, optional
```
simlang: i can't tell you exactly why, but i don't super-like adding benchmark (because now we have task, request and benchmark which are all kinda the same but not really), tokenizer (single use-case to the EvaluationAgent) and dataloader_args (only for benchmark)- it feels a bit cluttered for some reason.
But to be very honest, i don't have very constructive feedback how to improve this 🫠 

maybe one way would be, not to pass the benchmark as a string, but as a Task, which is being created outside of the agent and got there all of the information it needs - but this might require Benchmark inheriting the Task interface

what do you think?

simlang (changes requested):
so only two comments, but they kinda imply a lot 👀

src/pruna/evaluation/evaluation_agent.py
davidberenstein1957: I removed benchmrk from the `init` and creadted a seperate classmethod. I feel that this should be something we'll refactor during the 1.0 release, as you alraedy mentioned, there are a lot of single use argument that could have been enforced as part of the `Task`

src/pruna/data/datasets/prompt.py
simlang, line 229: I feel like i'm super annoying about this seed - since we stopped using the seed now, do you think it makes sense to add, that when the seed is not None (and while writing this we should make this optional as well) to info/warn the user that a seed wouldn't change anything

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -77,6 +77,53 @@ def __init__(
...
+
+        Returns
+        -------
+        EvaluationAgent
+            Agent after evaluation, with results accessible via the agent.
+
+        Examples
+        --------
+        >>> agent = EvaluationAgent.from_benchmark("Parti Prompts", model)
+        >>> agent = EvaluationAgent.from_benchmark("HPS", model, category="anime", fraction=0.1)
```
simlang, line 116:
```suggestion
        >>> agent = EvaluationAgent.from_benchmark("Parti Prompts")
        >>> agent = EvaluationAgent.from_benchmark("HPS", category="anime", fraction=0.1)
```

PR #581, opened 2026-03-19
feat: initial implementation for rapidata

begumcig:
## Description

This PR introduces a new stateful, asynchronous metric that submits generative model outputs (images, videos, etc.) to the [Rapidata](https://app.rapidata.ai/) platform for human evaluation. Raters compare outputs across models on configurable criteria (e.g. image quality, prompt alignment), and results are retrieved later once enough votes are collected.

### New Stuffs Implemented 
New `AsyncEvaluationMixin`: an abstract mixin defining the create_request() / retrieve_results() contract for metrics that delegate evaluation to external services asynchronously.
New` CompositeMetricResult`: a result type for metrics that return multiple labeled scores (e.g. one score per model), alongside a MetricResultProtocol to unify the interface between MetricResult and CompositeMetricResult.
EvaluationAgent updates: the agent now calls set_current_context(model_name=...) on all stateful metrics before each evaluation run, accepts a model_name parameter in evaluate(), and handles None returns from compute() (for async metrics that don't produce immediate results).

### Details
RapidataMetric lifecycle:

- Authenticate via client ID/secret or interactive browser login
- Create a benchmark from a prompt list or PrunaDataModule (or attach an existing one via from_benchmark() / from_benchmark_id())
- Create one or more leaderboards, each with a different evaluation instruction
- For each model: accumulate outputs via update(), submit via compute()
- Retrieve aggregated or per-leaderboard results once human evaluation completes
- Media handling supports str (URLs/paths), PIL.Image, and torch.Tensor: tensors and PIL images are saved to a temp directory for upload, then cleaned up.

Other changes:

Added set_current_context() as a no-op hook on StatefulMetric so the agent can uniformly notify all metrics of the current model without changing the structure of update() and compute()
Typed EvaluationAgent result lists as MetricResultProtocol instead of concrete MetricResult
Added rapidata as an optional dependency extra in pyproject.toml
Updated CI to install the rapidata extra

## Type of Change

## Checklist

### Usage

```
from pruna.evaluation.metrics.metric_rapiddata import RapidataMetric

# Initialization
rdm = RapidataMetric()
rdm.create_benchmark("my_bench_standalone", prompt_to_test)
rdm.create_request("Quality", instruction="Which video looks better?")

#  Use with Evaluation agent
agent = EvaluationAgent(
    request=[rapidata_metric],
    datamodule=datamodule,
)

results_a = agent.evaluate(model_a, model_name="model_a")

results_b = agent.evaluate(model_b, model_name="model_b")

# Use standalone
rdm.set_current_context("model_a")
rdm.update(prompts, model_a_gt, model_a_outputs)
rdm.compute()

rdm.set_current_context("model_b")
rdm.update(prompts, model_b_gt, model_b_outputs)
rdm.compute()

# Retrieve rankings
rdm.retrieve_results()
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -198,6 +198,9 @@ full = [
 vbench = [
     "vbench-pruna; sys_platform != 'darwin'",
 ]
+rapidata = [
+    "rapidata>=3.0.0"
+]
 dev = [
     "wget",
     "python-dotenv",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -232,6 +232,10 @@ cpu = []
 lmharness = [
     "lm-eval>=0.4.0"
 ]
+evaluation = [
+    "pruna[rapidata]",
+    "pruna[lmharness]"
+]
 
 # Intel extension is tightly coupled with the torch version
 intel = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,9 @@ possibly-missing-attribute = "ignore"
 missing-argument = "ignore"
 unused-type-ignore-comment = "ignore"
 
+[tool.bandit]
+exclude_dirs = ["tests", "docs"]
+
 
 [tool.coverage.run]
 source = ["src/pruna"]
--- a/src/pruna/evaluation/evaluation_agent.py
+++ b/src/pruna/evaluation/evaluation_agent.py
@@ -28,7 +28,7 @@ from pruna.engine.pruna_model import PrunaModel
 from pruna.engine.utils import get_device, move_to_device, safe_memory_cleanup, set_to_best_available_device
 from pruna.evaluation.metrics.metric_base import BaseMetric
 from pruna.evaluation.metrics.metric_stateful import StatefulMetric
-from pruna.evaluation.metrics.result import MetricResult
+from pruna.evaluation.metrics.result import MetricResult, MetricResultProtocol
 from pruna.evaluation.metrics.utils import ensure_device_consistency, get_device_map, group_metrics_by_inheritance
 from pruna.evaluation.task import Task
 from pruna.logging.logger import pruna_logger
@@ -71,8 +71,8 @@ class EvaluationAgent:
                 raise ValueError("When not using 'task' parameter, both 'request' and 'datamodule' must be provided.")
             self.task = Task(request=request, datamodule=datamodule, device=device)
 
-        self.first_model_results: List[MetricResult] = []
-        self.subsequent_model_results: List[MetricResult] = []
+        self.first_model_results: List[MetricResultProtocol] = []
+        self.subsequent_model_results: List[MetricResultProtocol] = []
         self.device = set_to_best_available_device(self.task.device)
         self.cache: List[Tensor] = []
         self.evaluation_for_first_model: bool = True
@@ -124,18 +124,20 @@ class EvaluationAgent:
         )
         return cls(task=task)
 
-    def evaluate(self, model: Any) -> List[MetricResult]:
+    def evaluate(self, model: Any, model_name: str | None = None) -> List[MetricResultProtocol]:
         """
         Evaluate models using different metric types.
 
         Parameters
         ----------
-        model : PrunaModel
+        model : Any
             The model to evaluate.
+        model_name : str | None, optional
+            The name of the model to evaluate. Required for rapidata benchmark submission.
 
         Returns
         -------
-        List[MetricResult]
+        List[MetricResultProtocol]
             The results of the model.
         """
         results = []
@@ -145,6 +147,9 @@ class EvaluationAgent:
         single_stateful_metrics = self.task.get_single_stateful_metrics()
         pairwise_metrics = self.task.get_pairwise_stateful_metrics()
         stateless_metrics = self.task.get_stateless_metrics()
+
+        for metric in single_stateful_metrics + pairwise_metrics:
+            metric.set_current_context(model_name=model_name)
 
         # Update and compute stateful metrics.
         pruna_logger.info("Evaluating stateful metrics.")
@@ -278,7 +283,7 @@ class EvaluationAgent:
 
     def compute_stateful_metrics(
         self, single_stateful_metrics: List[StatefulMetric], pairwise_metrics: List[StatefulMetric]
-    ) -> List[MetricResult]:
+    ) -> List[MetricResultProtocol]:
         """
         Compute stateful metrics.
 
@@ -296,16 +301,20 @@ class EvaluationAgent:
         """
         results = []
         for stateful_metric in single_stateful_metrics:
-            results.append(stateful_metric.compute())
+            result = stateful_metric.compute()
+            if result is not None:
+                results.append(result)
             stateful_metric.reset()
 
         if not self.evaluation_for_first_model and self.task.is_pairwise_evaluation():
             for pairwise_metric in pairwise_metrics:
-                results.append(pairwise_metric.compute())
+                result = pairwise_metric.compute()
+                if result is not None:
+                    results.append(result)
                 pairwise_metric.reset()
         return results
 
-    def compute_stateless_metrics(self, model: PrunaModel, stateless_metrics: List[Any]) -> List[MetricResult]:
+    def compute_stateless_metrics(self, model: PrunaModel, stateless_metrics: List[Any]) -> List[MetricResultProtocol]:
         """
         Compute stateless metrics.
 
--- a/src/pruna/evaluation/evaluation_agent.py
+++ b/src/pruna/evaluation/evaluation_agent.py
@@ -26,6 +26,7 @@ from pruna.data.pruna_datamodule import PrunaDataModule
 from pruna.data.utils import move_batch_to_device
 from pruna.engine.pruna_model import PrunaModel
 from pruna.engine.utils import get_device, move_to_device, safe_memory_cleanup, set_to_best_available_device
+from pruna.evaluation.metrics.context_mixin import EvaluationContextMixin
 from pruna.evaluation.metrics.metric_base import BaseMetric
 from pruna.evaluation.metrics.metric_stateful import StatefulMetric
 from pruna.evaluation.metrics.result import MetricResult, MetricResultProtocol
@@ -148,8 +149,9 @@ class EvaluationAgent:
         pairwise_metrics = self.task.get_pairwise_stateful_metrics()
         stateless_metrics = self.task.get_stateless_metrics()
 
-        for metric in single_stateful_metrics + pairwise_metrics:
-            metric.set_current_context(model_name=model_name)
+        for metric in single_stateful_metrics:
+            if isinstance(metric, EvaluationContextMixin):
+                metric.current_context = model_name
 
         # Update and compute stateful metrics.
         pruna_logger.info("Evaluating stateful metrics.")
--- a/src/pruna/evaluation/metrics/async_mixin.py
+++ b/src/pruna/evaluation/metrics/async_mixin.py
@@ -1,0 +1,53 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class AsyncEvaluationMixin(ABC):
+    """
+    Mixin for metrics that submit to external evaluation services and retrieve results asynchronously.
+
+    Subclasses implement create_request() to set up an evaluation
+    (e.g., create a leaderboard) and retrieve_results() to retrieve
+    outcomes (e.g., standings from human evaluators).
+    """
+
+    @abstractmethod
+    def create_request(self, *args, **kwargs) -> Any:
+        """
+        Create/configure an evaluation request on the external service.
+
+        Parameters
+        ----------
+        *args :
+            Variable length argument list.
+        **kwargs :
+            Arbitrary keyword arguments.
+        """
+
+    @abstractmethod
+    def retrieve_results(self, *args, **kwargs) -> Any:
+        """
+        Retrieve results from the external service.
+
+        Parameters
+        ----------
+        *args :
+            Variable length argument list.
+        **kwargs :
+            Arbitrary keyword arguments.
+        """
--- a/src/pruna/evaluation/metrics/async_mixin.py
+++ b/src/pruna/evaluation/metrics/async_mixin.py
@@ -21,8 +21,8 @@ class AsyncEvaluationMixin(ABC):
     """
     Mixin for metrics that submit to external evaluation services and retrieve results asynchronously.
 
-    Subclasses implement create_request() to set up an evaluation
-    (e.g., create a leaderboard) and retrieve_results() to retrieve
+    Subclasses implement create_async_request() to set up an evaluation
+    (e.g., create a leaderboard) and retrieve_async_results() to retrieve
     outcomes (e.g., standings from human evaluators).
     """
 
--- a/src/pruna/evaluation/metrics/async_mixin.py
+++ b/src/pruna/evaluation/metrics/async_mixin.py
@@ -27,7 +27,7 @@ class AsyncEvaluationMixin(ABC):
     """
 
     @abstractmethod
-    def create_request(self, *args, **kwargs) -> Any:
+    def create_async_request(self, *args, **kwargs) -> Any:
         """
         Create/configure an evaluation request on the external service.
 
@@ -40,7 +40,7 @@ class AsyncEvaluationMixin(ABC):
         """
 
     @abstractmethod
-    def retrieve_results(self, *args, **kwargs) -> Any:
+    def retrieve_async_results(self, *args, **kwargs) -> Any:
         """
         Retrieve results from the external service.
 
--- a/src/pruna/evaluation/metrics/context_mixin.py
+++ b/src/pruna/evaluation/metrics/context_mixin.py
@@ -55,7 +55,6 @@ class EvaluationContextMixin(ABC):
 
     def on_context_change(self) -> None:
         """Hook called when the context changes. Override to reset state."""
-        pass
 
     def _require_context(self) -> None:
         """Raise if no context has been set."""
--- a/src/pruna/evaluation/metrics/context_mixin.py
+++ b/src/pruna/evaluation/metrics/context_mixin.py
@@ -1,0 +1,62 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from abc import ABC
+
+
+class EvaluationContextMixin(ABC):
+    """
+    Mixin for metrics that evaluate multiple models sequentially.
+
+    Provides a current_context property that tracks which model is being
+    evaluated. Setting a new context triggers on_context_change(), which
+    subclasses can override to reset state between models.
+    """
+
+    _current_context: str | None = None
+
+    @property
+    def current_context(self) -> str | None:
+        """
+        Return the current context.
+
+        Returns
+        -------
+        str | None
+            The current context.
+        """
+        return self._current_context
+
+    @current_context.setter
+    def current_context(self, value: str | None) -> None:
+        """
+        Set the current context.
+
+        Parameters
+        ----------
+        value : str
+            The new context.
+        """
+        self._current_context = value
+        self.on_context_change()
+
+    def on_context_change(self) -> None:
+        """Hook called when the context changes. Override to reset state."""
+        pass
+
+    def _require_context(self) -> None:
+        """Raise if no context has been set."""
+        if self._current_context is None:
+            raise ValueError("No context set. Set current_context first.")
--- a/src/pruna/evaluation/metrics/context_mixin.py
+++ b/src/pruna/evaluation/metrics/context_mixin.py
@@ -49,8 +49,9 @@ class EvaluationContextMixin(ABC):
         value : str
             The new context.
         """
-        self._current_context = value
-        self.on_context_change()
+        if self._current_context != value:
+            self._current_context = value
+            self.on_context_change()
 
     def on_context_change(self) -> None:
         """Hook called when the context changes. Override to reset state."""
--- a/src/pruna/evaluation/metrics/metric_rapiddata.py
+++ b/src/pruna/evaluation/metrics/metric_rapiddata.py
@@ -16,8 +16,9 @@ from __future__ import annotations
 
 import shutil
 import tempfile
+import time
 from pathlib import Path
-from typing import Any, List, Literal
+from typing import Any, Callable, List, Literal
 
 import PIL.Image
 import torch
@@ -84,7 +85,10 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
 
     media_cache: List[torch.Tensor | PIL.Image.Image | str]
     prompt_cache: List[str]
-    higher_is_better: bool
+    # With every metric higher is actually better,
+    # Because for negative questions like "Which image has more errors?"
+    # We create the leaderboard with inverse_ranking=True, which reverses the ranking.
+    higher_is_better: bool = True
     default_call_type: str = "x_y"
     metric_name: str = METRIC_RAPIDATA
 
@@ -150,6 +154,7 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
         self,
         name: str,
         data: list[str] | PrunaDataModule,
+        data_assets: list[str] | None = None,
         split: Literal["test", "val", "train"] = "test",
         **kwargs,
     ) -> None:
@@ -159,12 +164,20 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
         The benchmark defines the prompt pool. Any data submitted to
         leaderboards later must be drawn from this pool.
 
+        Prompts can be provided as a list of strings or as a PrunaDataModule.
+        When using a list of strings, you can optionally pass data_assets as a list of file paths or URLs.
+        When using a PrunaDataModule, data assets are extracted automatically from the datamodule, if available.
+
         Parameters
         ----------
         name : str
             The name of the benchmark.
         data : list[str] | PrunaDataModule
             The prompts or dataset to benchmark against.
+        data_assets : list[str] | None
+            The assets to attach to the prompts.
+            For instance, if you wish to benchmark an image editing model,
+            you can pass the original images as data_assets.
         split : str, optional
             Which split to use when data is a PrunaDataModule. Default is "test".
         **kwargs : Any
@@ -181,6 +194,11 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
             # PrunaDataModule dataset loaders always renames the prompts column to "text"
             if hasattr(dataset, "column_names") and "text" in dataset.column_names:
                 data = list(dataset["text"])
+                data_assets = None  # When using a PrunaDataModule, we need to get the data assets from the datamodule.
+                if "image" in dataset.column_names:
+                    images = list(dataset["image"])  # Pruna text to image datasets always have an "image" column.
+                    #  Rapidata only accepts file paths or URLs, so we need to convert the images to file paths.
+                    data_assets = self._prepare_media_for_upload(images)
             else:
                 raise ValueError(
                     "Could not extract prompts from dataset.\n "
@@ -188,13 +206,14 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
                     or pass a list[str] directly instead."
                 )
 
-        self.benchmark = self.client.mri.create_new_benchmark(name, prompts=data, **kwargs)
+        self.benchmark = self.client.mri.create_new_benchmark(name, prompts=data, prompt_assets=data_assets, **kwargs)
 
     def create_async_request(
         self,
         name: str,
         instruction: str,
         show_prompt: bool = False,
+        show_prompt_assets: bool = False,
         **kwargs,
     ) -> None:
         """
@@ -216,11 +235,13 @@ class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
             The evaluation instruction shown to human raters.
         show_prompt : bool, optional
             Whether to show the prompt to raters. Default is False.
+        show_prompt_assets : bool, optional
+            Whether to show the prompt assets to raters. Default is False.
         **kwargs : Any
             Additional keyword arguments passed to the Rapidata API.
         """
         self._require_benchmark()
-        self.benchmark.create_leaderboard(name, instruction, show_prompt, **kwargs)
+        self.benchmark.create_leaderboard(name, instruction, show_prompt, show_prompt_assets, **kwargs)
 
     def set_current_context(self, model_name: str, **kwargs) -> None:
         """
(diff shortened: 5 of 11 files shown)


src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+# We don't use the MetricRegistry here
+# because we need to instantiate the Metric directly with benchmark and leaderboards.
+class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
+    """
+    Evaluate models with human feedback via the Rapidata platform.
+
+    Parameters
+    ----------
+    call_type : str
+        How to extract inputs from (x, gt, outputs). Default is "single".
```
davidberenstein1957: defeault mention here is different from the default below.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        metric.set_current_context("model_b")
+        metric.update(prompts, ground_truths, outputs_b)
+        metric.compute()
+
+        # wait for human votes
+        overall = metric.retrieve_results()
+    """
+
+    media_cache: List[torch.Tensor | PIL.Image.Image | str]
+    prompt_cache: List[str]
```
davidberenstein1957, line 102: we don't do image editing yet?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+    default_call_type: str = "x_y"
+    higher_is_better: bool = True
+    metric_name: str = METRIC_RAPIDATA
+    runs_on: List[str] = ["cpu", "cuda"]
```
davidberenstein1957: why do we need cuda?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        metric.compute()
+
+        # wait for human votes
+        overall = metric.retrieve_results()
+    """
+
+    media_cache: List[torch.Tensor | PIL.Image.Image | str]
+    prompt_cache: List[str]
+    default_call_type: str = "x_y"
+    higher_is_better: bool = True
```
davidberenstein1957, line 106: not sure if this always applies right, it depends on the question.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        """
+        metric = cls(
+            rapidata_client_id=rapidata_client_id,
+            rapidata_client_secret=rapidata_client_secret,
+        )
+        metric.benchmark = benchmark
+        return metric
+
+    @classmethod
+    def from_benchmark_id(
```
davidberenstein1957: Perhaps this can be handled with a single entry point `from rapidata_benchmark`?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+            The created metric.
+        """
+        metric = cls(
+            rapidata_client_id=rapidata_client_id,
+            rapidata_client_secret=rapidata_client_secret,
+        )
+        metric.benchmark = metric.client.mri.get_benchmark_by_id(benchmark_id)
+        return metric
+
+    def create_benchmark(
```
davidberenstein1957, line 174: perhaps we can retunr the benchmark class from this method too?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        metric.set_current_context("model_a")
+        metric.update(prompts, ground_truths, outputs_a)
+        metric.compute()
+
+        metric.set_current_context("model_b")
+        metric.update(prompts, ground_truths, outputs_b)
+        metric.compute()
+
+        # wait for human votes
+        overall = metric.retrieve_results()
```
davidberenstein1957: how long does this normally take? In my experience it was not super quick i.e. 15 minutes or an hour or so but perhaps I'm wrong.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        if self.benchmark is not None:
+            raise ValueError("Benchmark already created. Use from_benchmark() to attach to an existing one.")
+
+        # Rapidata benchmarks only accept a list of string,
+        # so we need to convert the PrunaDataModule to a list of strings.
+        if isinstance(data, PrunaDataModule):
+            split_map = {"test": data.test_dataset, "val": data.val_dataset, "train": data.train_dataset}
+            dataset = split_map[split]
+            # PrunaDataModule dataset loaders always renames the prompts column to "text"
+            if hasattr(dataset, "column_names") and "text" in dataset.column_names:
```
davidberenstein1957, line 227: is the column always `text`? 100% of the time?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+            else:
+                raise ValueError(
+                    "Could not extract prompts from dataset.\n "
+                    "Expected a 'text' column. Please use a suitable dataset from Pruna \
+                    or pass a list[str] directly instead."
+                )
+
+        self.benchmark = self.client.mri.create_new_benchmark(name, prompts=data, **kwargs)
+
+    def create_request(
```
davidberenstein1957: why create_request and not create_leaderboard or create_evaluation_instruction.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        #  Ignoring the type error because _require_model() has already been called, but ty can't see it.
+        self.benchmark.evaluate_model(
+            self.current_benchmarked_model,  # type: ignore[arg-type]
+            media=media,
+            prompts=self.prompt_cache,
+        )
+
+        self._cleanup_temp_media()
+
+        pruna_logger.info(
```
davidberenstein1957: perhaps this should we a warning

tests/evaluation/test_rapidata.py
```diff
@@ -0,0 +1,267 @@
...
+    with pytest.raises(ValueError, match="No model set"):
+        metric_with_benchmark.update(["p"], [None], [torch.rand(3, 32, 32)])
+
+
+def test_prepare_media_string_passthrough(metric_ready):
+    """Test that string URLs/paths are passed through as-is."""
+    metric_ready.media_cache = ["https://example.com/img.png", "/tmp/local.png"]
+    paths = metric_ready._prepare_media_for_upload()
+    assert paths == ["https://example.com/img.png", "/tmp/local.png"]
+    metric_ready._cleanup_temp_media()
```
davidberenstein1957: can't we just do this with an overal tempdir fixture?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+        -------
+        CompositeMetricResult | None
+            The overall standings, or None if not enough votes yet.
+        """
+        self._require_benchmark()
+
+        try:
+            standings = self.benchmark.get_overall_standings(*args, **kwargs)
+        except Exception as e:
+            if "ValidationError" in type(e).__name__:
```
davidberenstein1957: Aren't there other stages? There are three stages, right? Not done, non-significant, and significant?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+                return None
+            raise
+
+        scores = dict(zip(standings["name"], standings["score"]))
+        return CompositeMetricResult(
+            name=self.metric_name,
+            params={},
+            result=scores,
+            higher_is_better=self.higher_is_better,
+        )
```
davidberenstein1957:
```suggestion
        return CompositeMetricResult(
            name=self.metric_name,
            params={},
            result=dict(zip(standings["name"], standings["score"])),
            higher_is_better=self.higher_is_better,
        )
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+            standings = self.benchmark.get_overall_standings(*args, **kwargs)
+        except Exception as e:
+            if "ValidationError" in type(e).__name__:
+                pruna_logger.warning(
+                    "The benchmark hasn't finished yet.\n "
+                    "Please wait for more votes and try again."
+                    "Skipping."
+                )
+                return None
+            raise
```
davidberenstein1957: maybe this captures edge cases already as highlighted above.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+
+        scores = dict(zip(standings["name"], standings["score"]))
+        return CompositeMetricResult(
+            name=self.metric_name,
+            params={},
+            result=scores,
+            higher_is_better=self.higher_is_better,
+        )
+
+    def retrieve_granular_results(self, **kwargs) -> List[CompositeMetricResult]:
```
davidberenstein1957: Also, this method might be combined with something like `granular=False` or, by default, just return a complete dict with all scores, `overall`, and for the different leaderboards.

src/pruna/evaluation/metrics/metric_stateful.py
```diff
@@ -121,8 +121,27 @@ def update(self, *args, **kwargs) -> None:
             The keyword arguments to pass to the metric.
         """
 
+    def set_current_context(self, *args, **kwargs) -> None:
```
davidberenstein1957: does this need to be in this class? I feel it is rather specific to rapidata, right?

src/pruna/evaluation/metrics/result.py
```diff
@@ -14,13 +14,52 @@
...
+
+@runtime_checkable
+class MetricResultProtocol(Protocol):
+    """
+    Protocol defining the shared interface for all metric results.
+
+    Any metric result class should implement these attributes and methods
+    to be compatible with the evaluation pipeline.
+
+    # Have to include this to prevent ty errors.
```
davidberenstein1957, line 28: ouch

src/pruna/evaluation/evaluation_agent.py
```diff
@@ -28,7 +28,7 @@
 from pruna.engine.utils import get_device, move_to_device, safe_memory_cleanup, set_to_best_available_device
 from pruna.evaluation.metrics.metric_base import BaseMetric
 from pruna.evaluation.metrics.metric_stateful import StatefulMetric
-from pruna.evaluation.metrics.result import MetricResult
+from pruna.evaluation.metrics.result import MetricResult, MetricResultProtocol
```
davidberenstein1957, line 32: where are we still required to use MetricResult then?

pyproject.toml
```diff
@@ -191,6 +191,9 @@ full = [
 vbench = [
     "vbench-pruna; sys_platform != 'darwin'",
 ]
+rapidata = [
```
davidberenstein1957, line 204: alright, so we already start with the extra seperation, nice!
@begumcig I know that we can also do something like.
```bash
evaluation = [
    rapidata,
    vbench
]
could be nice to already start structuring like this, right?

File: src/pruna/evaluation/metrics/metric_rapiddata.py [hunk 20]
```diff
@@ -0,0 +1,465 @@
...
+from pruna.logging.logger import pruna_logger
+
+METRIC_RAPIDATA = "rapidata"
+
+
+# We don't use the MetricRegistry here
+# because we need to instantiate the Metric directly with benchmark and leaderboards.
+class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
+    """
+    Evaluate models with human feedback via the Rapidata platform.
```
davidberenstein1957: perhaps a URL or attribution to the website could be nice?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,465 @@
...
+            "Sent evaluation request for model '%s' to Rapidata.\n "
+            "It may take a while to collect votes from human raters.\n "
+            "Use retrieve_results() to check scores later, "
+            "or monitor progress at: "
+            "https://app.rapidata.ai/mri/benchmarks/%s",
+            self.current_benchmarked_model,
+            self.benchmark.id,
+        )
+
+    def retrieve_results(self, *args, **kwargs) -> CompositeMetricResult | None:
```
davidberenstein1957: It could be nice to add the functionality to wait till done, WDYT? It feels a bit harsh to fail outright without giving the option to do so.

davidberenstein1957 (commented):
Thanks for the PR, left some comments. I feel that some documentation is required for this too right, or is this generated automatically?

begumcig:
@cursor push fc62a066b6

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: hihi yeah it's a bit confusing, we have both call_type and a default_call_type parameter. and the default for the call_type is "single" which is different than default_call_type

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: yea it's not super quick

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: hmm maybe I don't understand the question, but rapidata doesn't support image editing models, so the input can only be "prompts", which makes it not possible to support image editing!

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Oh yeah, you are absolutely correct!!! Removing that!

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: We don't need it but we support it! do you think it should be only supported in cpu?

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: yes great idea!

src/pruna/evaluation/metrics/metric_rapiddata.py
davidberenstein1957: It can be both in this case, or should it require CUDA? I assumed it was a recommendation, but was not sure.

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Because this is the method from the async_mixin class that we overload here!

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: I would argue that it is more informative :D

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Yes that makes a lot of sense!

src/pruna/evaluation/metrics/metric_stateful.py
simlang: yes agree, maybe also with a mixin, as you did for the async metric. some kind of MulitStateMetricMixin or something

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,466 @@
...
+# because we need to instantiate the Metric directly with benchmark and leaderboards.
+class RapidataMetric(StatefulMetric, AsyncEvaluationMixin):
+    """
+    Evaluate models with human feedback via the Rapidata platform.
+
+    Parameters
+    ----------
+    call_type : str
+        How to extract inputs from (x, gt, outputs). Default is "single".
+    client : RapidataClient | None
```
simlang, line 65: similar to what we spoke about with benchmark and benchmark_id
maybe a single rapidata client, which can be RapidataClient | str | None? and then based on the type do different things

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Hmm, do you think it makes sense to wait 15 mins - 1 hour? 
If we do not have any results yet I am catching the error that rapidata is throwing, supressing it, and returning None, as a result. I thought it to be a middle ground between failing, and waiting indefinitely, I think the user also gets some sort of notification from the platform when the benchmark is finished anyway, does it make sense to wait for a long time?

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: does it make sense to add something like runs_on "any"

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: I am not sure I follow, can you explain a bit 🥹

src/pruna/evaluation/evaluation_agent.py
begumcig: we need it for the stateless metrics!

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: agree that the naming is not optimal, as we already have requests which are something different - maybe something which includes async?

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: For our prompt datasets yes! Because the prompt collate function also depends on this column to be named "text"

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: I feel like that would take the RapidataBenchmark outside of pruna, and we want the Benchmark to be used through pruna.

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,466 @@
...
+            The evaluation instruction shown to human raters.
+        show_prompt : bool, optional
+            Whether to show the prompt to raters. Default is False.
+        **kwargs : Any
+            Additional keyword arguments passed to the Rapidata API.
+        """
+        self._require_benchmark()
+        self.benchmark.create_leaderboard(name, instruction, show_prompt, **kwargs)
+
+    def set_current_context(self, model_name: str, **kwargs) -> None:
```
simlang: i guess rapidate only supports comparing two models, right?
i had a mulitstate (better name maybe multimodel idk) mixin comment before. does it make sense to have a max_number of models in there? so we keep all contexts and if a user tries to add a third model we say nope?

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: i think i disagree, but just from the naming.
it is an info, but it probably should be printed by default - so like a warning, hahahaha

src/pruna/evaluation/metrics/metric_stateful.py
begumcig: I also am really on the fence about this function, and am not a huge fan of it I have to say. 
In pruna, we expect our update() function to collect the scores, and compute is just an aggregation and doesn't need any parameters. But rapidata definitely needs the name of the model and I didn't want the change the function signature. I thought if we add this function to stateful metric, we can also re-use it in future metrics that would require extra parameters. But I am super open to suggestions, I agree it's not super pretty

tests/evaluation/test_rapidata.py
begumcig: Yess, makes a lot of sense!

simlang (commented):
just a first quick pass, david already covered most of my comments, so waiting for the second iteration!
but already looks super amazing 💅

pyproject.toml
begumcig: Hmm, I don't think vbench and rapidata have a lot of shared dependencies, so doesn't really make sense to me to group them together

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: oh you are so correct, because by default the runs_on in StatefulMetric is configured to run on everything. So I could remove this parameter from this all together actually!

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: You can actually compare as many models as you like!

src/pruna/evaluation/metrics/metric_rapiddata.py
davidberenstein1957: I'm not sure, but currently, you have to check, then get an error, then wait 15 minutes, and then manually check again.

pyproject.toml
davidberenstein1957: I assumed it could be a shared dependency group for all evaluation metrics. You can add extras to extras, but perhaps you'd like to keep them seperate?

src/pruna/evaluation/metrics/metric_rapiddata.py
davidberenstein1957: So, currently it seems we can pass prompts and media as generations, but I understood they also support image-editing tasks where we'd pass custom formatting, but this is not something we want to support for now?

begumcig: Okay, I looked into the SDK architecture for this, and it seems Rapidata enables image editing via a 'prompt asset' pattern.

Essentially, the unedited image is linked as a prompt_asset when the benchmark is first created. When we call evaluate_model(), we only send the prompts, and model's output. The Rapidata backend then automatically pairs the prompt and the unedited imag in the UI if show_prompt_asset=True is set on the leaderboard.

Since these assets are static context already tied to the prompts on their server, we don't need to cache or pass them through our update() or compute() functions. I am updating the create logic to allow passing these assets during the initial setup instead. Do you think that makes sense for our current wrapper?

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: For RapidataClient we both need the rapidata_client_id and the rapidata_client_secret. I can make it a RapidataClient | dict [str, str] | None but I feel like this hides the the names of the variables needed for the function a little bit. I will implement it still if you think the benefit from a single function we get is more than what we are losing here!

src/pruna/evaluation/metrics/async_mixin.py
```diff
@@ -0,0 +1,53 @@
...
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+
+class AsyncEvaluationMixin(ABC):
+    """
+    Mixin for metrics that submit to external evaluation services and retrieve results asynchronously.
+
+    Subclasses implement create_request() to set up an evaluation
```
simlang:
```suggestion
    Subclasses implement create_async_request() to set up an evaluation
```

src/pruna/evaluation/metrics/async_mixin.py
```diff
@@ -0,0 +1,53 @@
...
+    (e.g., create a leaderboard) and retrieve_results() to retrieve
```
simlang:
```suggestion
    (e.g., create a leaderboard) and retrieve_async_results() to retrieve
```

src/pruna/evaluation/metrics/context_mixin.py
```diff
@@ -0,0 +1,62 @@
...
+        """
+        Set the current context.
+
+        Parameters
+        ----------
+        value : str
+            The new context.
+        """
+        self._current_context = value
+        self.on_context_change()
```
simlang: what do you think about checking if the current_context value _actually_ changed an only triggering then?
back in my frontend days, it was always done like this, but completely up to you

```suggestion
		if self._current_context != value:
        	self._current_context = value
        	self.on_context_change()
```

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: yes I agree with you, didn't see that!

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+
+    Parameters
+    ----------
+    call_type : str
+        How to extract inputs from (x, gt, outputs). Only "single" is supported.
+    client : RapidataClient | None
+        The Rapidata client to use. If None, a new one is created.
+    rapidata_client_id : str | None
+        The client ID of the Rapidata client.
+        If none, the credentials are read from the environment variable RAPIDATA_CLIENT_ID.
```
simlang:
```suggestion
        If None, the credentials are read from the environment variable RAPIDATA_CLIENT_ID.
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        If credentials are not found in the environment variable, you will be prompted to login via browser.
+    rapidata_client_secret : str | None
+        The client secret of the Rapidata client.
+        If none, the credentials are read from the environment variable RAPIDATA_CLIENT_SECRET.
```
simlang:
```suggestion
        If None, the credentials are read from the environment variable RAPIDATA_CLIENT_SECRET.
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        self,
+        call_type: str = SINGLE,
+        client: RapidataClient | None = None,
+        rapidata_client_id: str | None = None,
+        rapidata_client_secret: str | None = None,
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        self.client = client or RapidataClient(
```
simlang, line 125: is the RapidataClient throwing something if there something going wrong, or de have to check if client is None and throw ourselves?

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        self.add_state("prompt_cache", default=[])
+        self.benchmark: RapidataBenchmark | None = None
+        self.current_context: str | None = None
+
+    @classmethod
+    def from_rapidata_benchmark(
+        cls,
+        benchmark: RapidataBenchmark | str,
+        rapidata_client_id: str | None = None,
+        rapidata_client_secret: str | None = None
```
simlang, line 140: does it makes sense to also pass an already created client?

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: first did not get what you mean Begüm, but now I do - still, does it make sense to maybe return the benchmark id? so it can still be used in multiple Metrics

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        When using a PrunaDataModule, data assets are extracted automatically from the datamodule, if available.
+
+        Parameters
+        ----------
+        name : str
+            The name of the benchmark.
+        data : list[str] | PrunaDataModule
+            The prompts or dataset to benchmark against.
+        data_assets : list[str] | None
+            The assets to attach to the prompts.
```
simlang: I feel like i would swap the data_assets explanation here with the one in the overall description

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+            For instance, if you wish to benchmark an image editing model,
+            you can pass the original images as data_assets.
+        split : str, optional
+            Which split to use when data is a PrunaDataModule. Default is "test".
+        **kwargs : Any
+            Additional keyword arguments passed to the Rapidata API.
+        """
+        if self.benchmark is not None:
+            raise ValueError("Benchmark already created. Use from_benchmark() to attach to an existing one.")
```
simlang: i think the message is a bit misleading, i proposed something but whatever you like ofc
```suggestion
            raise ValueError("Benchmark already created. Use from_benchmark() to create a new metric from an existing one.")
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        # Rapidata benchmarks only accept a list of string,
+        # so we need to convert the PrunaDataModule to a list of strings.
+        if isinstance(data, PrunaDataModule):
+            split_map = {"test": data.test_dataset, "val": data.val_dataset, "train": data.train_dataset}
+            dataset = split_map[split]
+            # PrunaDataModule dataset loaders always renames the prompts column to "text"
+            if hasattr(dataset, "column_names") and "text" in dataset.column_names:
+                data = list(dataset["text"])
+                data_assets = None  # When using a PrunaDataModule, we need to get the data assets from the datamodule.
+                if "image" in dataset.column_names:
```
simlang, line 230: this is actually so cool. maybe we need a flag though, if we actually want to use the images from the dataset?
idk - i can see myself using some dataset e.g. LAION for prompts which happens to have images, and then i would automatically assume an editing task

src/pruna/evaluation/metrics/metric_rapiddata.py
simlang: nvm then!

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+
+        Parameters
+        ----------
+        model_name : str
+            The name of the model to evaluate.
+        **kwargs : Any
+            Additional keyword arguments.
+        """
+        self.current_context = model_name
+        self.reset()  # Clear the cache for the new model.
```
simlang: this line should now be handled by the context mixin as you have overridden it now later with the reset thingy

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        Parameters
+        ----------
+        *args : Any
+            Additional arguments passed to the Rapidata API.
+        **kwargs : Any
+            Additional keyword arguments passed to the Rapidata API.
+
+        Returns
+        -------
+        List[CompositeMetricResult] | None
```
simlang: update this to include the tuple

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        self,
+        fetch_fn: Callable,
+        is_blocking: bool,
+        timeout: float,
+        poll_interval: float,
+        *args,
+        **kwargs,
+    ) -> CompositeMetricResult | List[CompositeMetricResult] | None:
+        """
+        Wait for  the results or return whatever we have as is from the benchmark.
```
simlang:
```suggestion
        Wait for the results or return whatever we have as is from the benchmark.
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        metric.set_current_context("model_a")
+        metric.update(prompts, ground_truths, outputs_a)
+        metric.compute()
+
+        metric.set_current_context("model_b")
+        metric.update(prompts, ground_truths, outputs_b)
+        metric.compute()
+
+        # wait for human votes
+        overall = metric.retrieve_results()
```
simlang:
```suggestion
        overall = metric.retrieve_async_results()
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+    def compute(self) -> None:
+        """
+        Submit the accumulated outputs for the current model to Rapidata.
+
+        Converts cached media to uploadable file paths if necessary (saving tensors and
+        PIL images to a temporary directory), submits them to the benchmark,
+        and cleans up temporary files.
+
+        This method does **not** return a result — human evaluation is
+        asynchronous. Use :meth:`retrieve_results` or
```
simlang:
```suggestion
        asynchronous. Use :meth:`retrieve_async_results` or
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+            media=media,
+            prompts=self.prompt_cache,
+        )
+
+        self._cleanup_temp_media()
+
+        pruna_logger.warning(
+            "Sent evaluation request for model '%s' to Rapidata.\n "
+            "It may take a while to collect votes from human raters.\n "
+            "Use retrieve_results() to check scores later, "
```
simlang:
```suggestion
            "Use retrieve_async_results() to check scores later, "
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+        Additional keyword arguments passed to StatefulMetric.
+
+    Examples
+    --------
+    Standalone usage::
+        metric = RapidataMetric()
+        # OR metric = RapidataMetric.from_benchmark_id("69bc528fa858d3fbc1ea1475")
+
+        metric.create_benchmark("my_bench", prompts)
+        metric.create_request("Quality", instruction="Which image looks better?")
```
simlang:
```suggestion
        metric.create_async_request("Quality", instruction="Which image looks better?")
```

tests/evaluation/test_rapidata.py
```diff
@@ -0,0 +1,351 @@
...
+    mock_client.mri.create_new_benchmark.assert_called_once_with("my-bench", prompts=["prompt1", "prompt2"], prompt_assets=None)
+
+
+def test_create_benchmark_raises_if_already_exists(metric_with_benchmark):
+    """Test that creating a benchmark twice raises."""
+    with pytest.raises(ValueError, match="Benchmark already created"):
+        metric_with_benchmark.create_benchmark("dup", data=["x"])
+
+def test_create_request_raises_without_benchmark(metric):
+    """Test that create_request raises without a benchmark."""
```
simlang:
```suggestion
def test_create_async_request_raises_without_benchmark(metric):
    """Test that create_async_request raises without a benchmark."""
```

tests/evaluation/test_rapidata.py
```diff
@@ -0,0 +1,351 @@
...
+    with pytest.raises(ValueError, match="No benchmark configured"):
+        metric.create_async_request("quality", "Rate image quality")
+
+
+def test_create_request_delegates_to_leaderboard(metric_with_benchmark):
+    """Test that create_request delegates to the benchmark."""
```
simlang:
```suggestion
def test_create_async_request_delegates_to_leaderboard(metric_with_benchmark):
    """Test that create_async_request delegates to the benchmark."""
```

tests/evaluation/test_rapidata.py
```diff
@@ -0,0 +1,351 @@
...
+    """When using a PrunaDataModule, explicit data_assets are overridden."""
+    ds = Dataset.from_dict({"text": ["p1"]})
+    dm = PrunaDataModule(train_ds=ds, val_ds=ds, test_ds=ds, collate_fn=lambda x: x, dataloader_args={})
+    metric.create_benchmark("bench", data=dm, data_assets=["/should/be/ignored.png"])
+    mock_client.mri.create_new_benchmark.assert_called_once_with(
+        "bench", prompts=["p1"], prompt_assets=None,
+    )
+
+
+def test_create_request_forwards_show_prompt_assets_true(metric_with_benchmark):
```
simlang:
```suggestion
def test_create_async_request_forwards_show_prompt_assets_true(metric_with_benchmark):
```

src/pruna/evaluation/metrics/metric_rapiddata.py
```diff
@@ -0,0 +1,573 @@
...
+
+        metric.set_current_context("model_a")
```
simlang, line 89: just a question about how this works. (have not used rapidata, so maybe it's a stupid question)

So this looks like we upload per model, the model outputs and a ground truth (?). I thought you would upload the outputs of different models and then these compete against each other?

src/pruna/evaluation/metrics/metric_stateful.py
```diff
@@ -122,7 +122,7 @@ def update(self, *args, **kwargs) -> None:
         """
 
     @abstractmethod
-    def compute(self) -> Any:
+    def compute(self,) -> Any:
```
simlang: stupid question but was this by accident, haha?

tests/algorithms/testers/moe_kernel_tuner.py
```diff
@@ -34,7 +34,6 @@ def post_smash_hook(self, model: PrunaModel) -> None:
 
     def _resolve_hf_cache_config_path(self) -> Path:
         """Read the saved artifact and compute the expected HF cache config path."""
-
```
simlang: not connected to this PR, but I'll let it slide 👀

simlang (changes requested):
insane amount of comments, but everything is extremely minor. amazing work so far!!! 🚀

davidberenstein1957 (approved):
should be good after the final comments from @simlang have been adressed.

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Oh insanely good point, I looked into it and we throw a ValueError if we cannot create a client (coming from the OpenAI API) within 5 mins. So we cannot have a metric without a client!

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: yupp, updated

src/pruna/evaluation/metrics/metric_stateful.py
begumcig: Umm definitely not, it has a super secret special meaning, but will remove for your sake

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: So the thing is a single benchmark can contain multiple leaderboards, and each leaderboard has its own `show_prompt_assets` flag. This flag is set to false by default, meaning ground truth images are hidden unless explicitly enabled when the leaderboard is created. (so you would have to explicitly say if a task is an editing task)

If we instead move this setting to the benchmark level here, it would apply to all leaderboards under that benchmark. As a result, every leaderboard would have to follow the same configuration, either all showing prompt assets (edit tasks) or all hiding them (non-edit tasks). So I would rather keep the flag at the leaderboard level, do you agree?

src/pruna/evaluation/metrics/metric_rapiddata.py
begumcig: Let me try to explain my understanding of the system:

First, you start by creating a benchmark. In that benchmark, you define:
- A set of prompts (this is standard)
- And optionally prompt assets (In our case, these are the ground truth / unedited images incase we wanna do an edit task)

Then, from that benchmark, you create one or more leaderboards. Each leaderboard is tied to that same benchmark, which means:

- You can only use the prompts you originally defined
- You can’t send new or random prompts later

Now, when you evaluate a model, you send:

- The prompt
- The model output (inference result)

When `metric.update(prompts, ground_truths, outputs)` is called, we drop the ground truths (since the call type is `x_y`)

Then how do we do image editing?
When creating the leaderboard, you set:  `show_prompt_assets = True`

If you do that:

Rapidata will automatically match:
- The prompt
- Its prompt asset (original image from the benchmark)
- The model output

So you don’t have to send the original image yourself; it’s already stored from when you created the benchmark.

So what you actually see depends on the task:

If you’re doing just image comparison: You’ll only see the model outputs being compared
If you’re doing an edit task: You’ll see:
- The original (unedited) image on top
- The model outputs below

simlang (approved):
amazing!

PR #605, opened 2026-03-31
feat: extend moe model check to multimodal ones and add block quantization parameters

llcnt:
## Description
Just extend the model check, and the logic inside the `moe_kernel_tuner.py`, so that one can tune multimodel MoE (eg. model that contains a vision tower and a text model nested, as qwen3.5-32b).
And make the block quant shape (ie the size of the block on which the scales are computed for re-normalization before quantization) usable from the user.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

---

### DIFF
--- a/src/pruna/algorithms/moe_kernel_tuner.py
+++ b/src/pruna/algorithms/moe_kernel_tuner.py
@@ -178,6 +178,11 @@ class MoeKernelTuner(PrunaAlgorithmBase):
         model_config = getattr(model, "config", None)
         if model_config is None:
             raise ValueError(f"Model {model.__class__.__name__} has no config.")
+        # Multimodal MoE (e.g. Qwen3_5MoeForConditionalGeneration): MoE parameters live on text_config.
+        if getattr(model_config, "num_experts", None) is None:
+            text_cfg = getattr(model_config, "text_config", None)
+            if text_cfg is not None and getattr(text_cfg, "num_experts", None) is not None:
+                model_config = text_cfg
 
         tensor_parallel_size = int(smash_config["tensor_parallel_size"])
         if model.__class__.__name__ == "HunyuanImage3ForCausalMM":
--- a/src/pruna/algorithms/moe_kernel_tuner.py
+++ b/src/pruna/algorithms/moe_kernel_tuner.py
@@ -130,6 +130,18 @@ class MoeKernelTuner(PrunaAlgorithmBase):
                 default_value=8,
                 meta={"desc": "Maximum (log) block size for tiling through intermediate dimension."},
             ),
+            OrdinalHyperparameter(
+                "block_quant_shape_n",
+                sequence=[32, 64, 128, 256, 512, 1024, 2048, 4096, None],
+                default_value=None,
+                meta={"desc": "Block size for quantization through input dimension."},
+            ),
+            OrdinalHyperparameter(
+                "block_quant_shape_k",
+                sequence=[32, 64, 128, 256, 512, 1024, 2048, 4096, None],
+                default_value=None,
+                meta={"desc": "Block size for quantization through intermediate dimension."},
+            ),
         ]
 
     def model_check_fn(self, model: Any) -> bool:
@@ -199,13 +211,10 @@ class MoeKernelTuner(PrunaAlgorithmBase):
         dtype = torch.bfloat16 if dtype == "bfloat16" else torch.float16
         use_fp8_w8a8 = smash_config["weight_dtype"] == "fp8_w8a8"
         use_int8_w8a16 = smash_config["weight_dtype"] == "int8_w8a16"
-
-        # (ii.b) Extract block quantization shape so the tuned config file
-        # name matches what vLLM looks for at serving time (e.g.
-        # ``E=256,N=512,...,block_shape=[128,128].json``).
-        block_quant_shape = _get_block_quant_shape(getattr(model, "config", None))
-        if block_quant_shape is None:
-            block_quant_shape = _get_block_quant_shape(model_config)
+        block_quant_shape = [
+            smash_config["block_quant_shape_n"],
+            smash_config["block_quant_shape_k"],
+        ]
 
         # (iii) Tune the kernel over a range of batch sizes (single GPU per Ray worker).
         batch_sizes = [1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192]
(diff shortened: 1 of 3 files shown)


src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -130,6 +130,18 @@ def get_hyperparameters(self) -> list:
                 default_value=8,
                 meta={"desc": "Maximum (log) block size for tiling through intermediate dimension."},
             ),
+            OrdinalHyperparameter(
+                "block_quant_shape_n",
+                sequence=[32, 64, 128, 256, 512, 1024, 2048, 4096, None],
+                default_value=None,
+                meta={"desc": "Block size for quantization through input dimension."},
```
gsprochette: Can we make the default behavior explicit? Is it something like "all block sizes compatible with the available kernels"?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -130,6 +130,18 @@ def get_hyperparameters(self) -> list:
...
+            ),
+            OrdinalHyperparameter(
+                "block_quant_shape_k",
+                sequence=[32, 64, 128, 256, 512, 1024, 2048, 4096, None],
+                default_value=None,
+                meta={"desc": "Block size for quantization through intermediate dimension."},
```
gsprochette: Explicit default behavior?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -178,6 +190,11 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         model_config = getattr(model, "config", None)
         if model_config is None:
             raise ValueError(f"Model {model.__class__.__name__} has no config.")
+        # Multimodal MoE (e.g. Qwen3_5MoeForConditionalGeneration): MoE parameters live on text_config.
+        if getattr(model_config, "num_experts", None) is None:
+            text_cfg = getattr(model_config, "text_config", None)
+            if text_cfg is not None and getattr(text_cfg, "num_experts", None) is not None:
+                model_config = text_cfg
```
gsprochette, line 222: should we raise an error in the else case to make the error more explicit in the eventuality of a custom or future model?

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -194,6 +211,15 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
         dtype = torch.bfloat16 if dtype == "bfloat16" else torch.float16
         use_fp8_w8a8 = smash_config["weight_dtype"] == "fp8_w8a8"
         use_int8_w8a16 = smash_config["weight_dtype"] == "int8_w8a16"
+        block_quant_shape = None
+        if (
```
gsprochette: the case where one is manually set by the user but not the other (e.g. `block_quant_shape_n=32` and `block_quant_shape_k=None`) is not handled, can you take care of it?
- handle the two cases where one is set and not the other, or if not possible,
- raise an error or log a warning stating that you must set both if you want to manually set one, then this should also appear in the hyperparameter descriptions

gsprochette (changes requested):
Cool PR, there a missing behavior when setting one block_quant_shape and not the other, after this and a quick documentation it will be ready to merge :)

src/pruna/algorithms/moe_kernel_tuner.py
```diff
@@ -130,6 +130,32 @@ def get_hyperparameters(self) -> list:
                 default_value=8,
                 meta={"desc": "Maximum (log) block size for tiling through intermediate dimension."},
             ),
+            OrdinalHyperparameter(
+                "block_quant_shape_n",
+                sequence=[32, 64, 128, 256, 512, 1024, 2048, 4096, None],
+                default_value=None,
+                meta={
+                    "desc": (
+                        "Block size for quantization along the N dimension when weight_dtype is "
```
sharpenb: What is the N dimension? What is the Block size in this case? This is not a paramter automatically tuned? We can probably clarify the doc here :)

sharpenb (approved):
Thanks! It looks good. Left a comment to clarify the docs and it is clear for people what to expect :)

src/pruna/engine/model_checks.py
```diff
@@ -121,7 +123,13 @@ def is_moe_lm(model: Any) -> bool:
...
         True if the model is a MoE LM, False otherwise.
     """
-    return hasattr(getattr(model, "config", None), "num_experts")
+    config = getattr(model, "config", None)
+    if config is None:
+        return False
+    if getattr(config, "num_experts", None) is not None:
+        return True
+    text_cfg = getattr(config, "text_config", None)
+    return text_cfg is not None and getattr(text_cfg, "num_experts", None) is not None
```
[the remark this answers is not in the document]
llcnt: good catch! fixed

src/pruna/algorithms/reduce_noe.py
```diff
@@ -126,9 +126,11 @@ def _apply(self, model: Any, smash_config: SmashConfigPrefixWrapper) -> Any:
...
                 target_name = smash_config["target_name"]
-                if target_name not in config_json:
+                # if VLM MoE, the config that contains moe is the text_config one.
+                text_config_json = config_json.get("text_config", config_json)
+                if target_name not in text_config_json:
                     raise KeyError(f"Target name '{target_name}' not found in config file at {config_path}")
-                config_json[target_name] = smash_config["num_experts_per_token"]
+                text_config_json[target_name] = smash_config["num_experts_per_token"]
                 with config_path.open("w", encoding="utf-8") as f:
                     json.dump(config_json, f, indent=2)
```
gsprochette, line 135: Should this be `text_config_json` instead? I think text_config_json is not being used at all here, have you tested it?

gsprochette (changes requested):
Thanks a lot about the updates in moe_kernel_tuner, much clearer. I think there's a problem in reduce_noe that was introduced so I'll let you address it before I can approve

src/pruna/algorithms/reduce_noe.py
gsprochette: my bad

gsprochette (approved):
All good on my side ! Thanks a lot for the clarifications :)

PR #622, opened 2026-04-08
ci: separate extra installs

begumcig:
## Description

- Isolated whisper dependencies into a dedicated [whisper] extra. Moved ctranslate2 and whisper-s2t out of the default full install group into [whisper] to prevent side-effects from whisper_s2t imports breaking unrelated tests at collection time. The previously hard-skipped TestWhisperS2T is now gated behind requires_whisper instead.

- Pin torch/torchvision bounds on the [intel] extra. intel-extension-for-pytorch is tightly coupled to the torch version; added explicit torch>=2.7.0,<2.9.0 and torchvision>=0.22.0,<0.24.0 constraints so installs don't silently break.
- Also switch TestIPEXLLM to use opt_125m model and drop latency metric. The algorithm didn't work on a tiny model due to latent size being small; we also need a specific inference handler for it for the metrics, which can be done later, so removed the evaluation part for now. 
- IPEX is strictly coupled with the torch version and is not maintained. So makes sense to run it in the nightlies but not on the CI, as it requires its own setup.

- Declared packaging conflicts between intel/awq and gptq/awq. These extras pull in incompatible transitive deps; the new [tool.uv.conflicts] entries surface the error at resolve time instead of runtime.

- Added requires_* markers for optional-extra tests. New markers requires_awq, requires_gptq, requires_stable_fast, requires_vllm, requires_intel, requires_lmharness, and requires_whisper let CI select/deselect tests based on what's installed rather than relying on implicit skips or collection errors.

- Removed the old HIGH_RESOURCE_FIXTURES list and runtime CUDA/multi-GPU detection. Tests without an explicit hardware mark are now auto-tagged cpu. Incompatible device_parametrized variants (e.g. a CUDA-only algorithm's CPU test) are deselected at collection time instead of skipping after expensive fixture setup. Added explicit pytestmark = pytest.mark.cpu to pure-CPU test modules (test_compatibility_symmetry, test_evalharness_metrics, test_registry, test_metrics) so they're correctly selected by -m "cpu".

- Updated CI workflow. Dropped --extra vllm from the default uv sync in the setup action. Changed the pytest filter from -m "not (slow or style or high_cpu or cuda or distributed)" to the positive-selection form -m "cpu and not slow and not requires_intel".

## Related Issue

Fixes #(issue number)

## Type of Change

- [x] Refactor (no functional change)
- [ ] Documentation update

## Testing

## Checklist

---

> **Note:** 

---

## First Prune (1-year OSS anniversary)

**First Prune** marks one year of Pruna’s open-source work. During the initiative window, qualifying merged contributions count toward **First Prune**. You can earn credits for our **performance models** via our **API**.

If you’d like your contribution to count toward **First Prune**, here’s how it works:

- **Initiative window:** First Prune starts on **March 31**.
- **Issue assignment:** For your PR to count toward First Prune, the related issue must be assigned to the contributor opening the PR. Issues are labeled with **first-prune**.
- **Open for review:** Please **open your PR and mark it ready for review by April 30** (end of April).
- **Review priority:** We’ll make our best effort to review quickly any PR that is open and has a review request before April 30.
- **Credits:** Each qualifying merged PR earns **30 credits**. We’ll be in touch after all qualifying PRs for **First Prune** have been merged.
- **To get started:** Have a look at [all models](https://www.pruna.ai/all-models). You’ll need to **sign up** on the [dashboard](https://dashboard.pruna.ai/login) before you can redeem your credits.

### DIFF
--- a/.github/actions/setup-uv-project/action.yml
+++ b/.github/actions/setup-uv-project/action.yml
@@ -19,4 +19,4 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - shell: bash
-      run: uv sync --extra dev --extra lmharness
+      run: uv sync --extra dev --extra lmharness --extra intel
--- a/.github/actions/setup-uv-project/action.yml
+++ b/.github/actions/setup-uv-project/action.yml
@@ -19,4 +19,4 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - shell: bash
-      run: uv sync --extra dev --extra lmharness
+      run: uv sync --extra dev
--- a/.github/actions/setup-uv-project/action.yml
+++ b/.github/actions/setup-uv-project/action.yml
@@ -19,4 +19,4 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - shell: bash
-      run: uv sync --extra dev --extra lmharness --extra intel
+      run: uv sync --extra dev --extra lmharness
--- a/.github/actions/setup-uv-project/action.yml
+++ b/.github/actions/setup-uv-project/action.yml
@@ -19,4 +19,4 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - shell: bash
-      run: uv sync --extra dev --extra lmharness --extra vllm
+      run: uv sync --extra dev --extra lmharness
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -66,6 +66,14 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.11"]
+        name: ["base", 'lmharness']
+        include:
+        - name: base
+          extras: ""
+          mark_filter: "cpu and not slow and not style and not requires_intel and not requires_lmharness"
+        - name: lmharness
+          extras: "--extra lmharness"
+          mark_filter: "requires_lmharness"
 
     env:
       HF_TOKEN: ${{ secrets.HF_TOKEN }}
@@ -80,6 +88,10 @@ jobs:
       - uses: ./.github/actions/setup-uv-project
         with:
           python-version: ${{ matrix.python-version }}
+      
+      - name: Install additional extras
+        if: matrix.extras != ''
+        run: uv sync --extra dev ${{ matrix.extras }}
 
       - name: Cache Hugging Face datasets and models
         uses: actions/cache@v5
@@ -123,4 +135,4 @@ jobs:
       - name: Run tests with pytest-rerunfailures
         run: |
           echo "Running tests with up to 3 reruns on failure using $PYTEST_WORKERS workers..."
-          uv run pytest -n $PYTEST_WORKERS -m "cpu and not slow and not requires_intel" --reruns 3 --reruns-delay 10 --maxfail=1
+          uv run pytest -n $PYTEST_WORKERS -m "${{ matrix.mark_filter }}" --reruns 3 --reruns-delay 10 --maxfail=1
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -123,4 +123,4 @@ jobs:
       - name: Run tests with pytest-rerunfailures
         run: |
           echo "Running tests with up to 3 reruns on failure using $PYTEST_WORKERS workers..."
-          uv run pytest -n $PYTEST_WORKERS -m "not (slow or style or high_cpu or cuda or distributed)" --reruns 3 --reruns-delay 10 --maxfail=1
+          uv run pytest -n $PYTEST_WORKERS -m "cpu and not slow" --reruns 3 --reruns-delay 10 --maxfail=1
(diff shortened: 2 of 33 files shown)


.github/actions/setup-uv-project/action.yml
```diff
@@ -19,4 +19,4 @@ runs:
         python-version: ${{ inputs.python-version }}
 
     - shell: bash
-      run: uv sync --extra dev --extra lmharness --extra vllm
+      run: uv sync --extra dev --extra lmharness
```
gsprochette: why are we still installing lmharness? Shouldn't it be isolated as well?

.github/workflows/tests.yaml
```diff
@@ -123,4 +123,4 @@ jobs:
       - name: Run tests with pytest-rerunfailures
         run: |
           echo "Running tests with up to 3 reruns on failure using $PYTEST_WORKERS workers..."
-          uv run pytest -n $PYTEST_WORKERS -m "not (slow or style or high_cpu or cuda or distributed)" --reruns 3 --reruns-delay 10 --maxfail=1
+          uv run pytest -n $PYTEST_WORKERS -m "cpu and not slow and not requires_intel" --reruns 3 --reruns-delay 10 --maxfail=1
```
gsprochette: I think this will run the `style` tests a second time, since they are already running during the linting job line 56

tests/algorithms/testers/awq.py
```diff
@@ -5,7 +5,7 @@
 from .base_tester import AlgorithmTesterBase
 
 
-@pytest.mark.slow
+@pytest.mark.requires_awq
```
gsprochette, line 8: is it not slow anymore?

tests/algorithms/testers/upscale.py
```diff
@@ -5,7 +5,8 @@
 from .base_tester import AlgorithmTesterBase
 
 
-@pytest.mark.cuda
+# Takes too long to run on CPU, so we mark it as slow
+@pytest.mark.slow
```
gsprochette: You could also do something like
```
    @classmethod
    def compatible_devices(cls) -> list[str]:  # type: ignore[override]
        """Get the devices compatible with this algorithm, except CPU (too slow)."""
        return [d for d in super().compatible_devices() if d != "cpu"]
```
if the intent is to not run this test on cpu.
I do agree that the cuda mark is not the correct tool because it will just create a cpu test eitherway, and it will have both the cpu and cuda mark...

tests/evaluation/test_registry.py
```diff
@@ -5,6 +5,8 @@
 from pruna.evaluation.metrics.metric_base import BaseMetric
 from pruna.evaluation.metrics.registry import MetricRegistry
 
+pytestmark = pytest.mark.cpu
```
gsprochette: this is redundant with the cpu default in conftest.py and not very explicit

tests/telemetry/test_metrics.py
```diff
@@ -26,6 +26,8 @@
     set_opentelemetry_log_level,
 )
 
+pytestmark = pytest.mark.cpu
```
gsprochette: This is redundant with the cpu default in conftest.py, and not very explicit

pyproject.toml
```diff
@@ -230,8 +227,12 @@ cpu = []
 lmharness = [
     "lm-eval>=0.4.0"
 ]
+
+# Intel extension is tightly coupled with the torch version
 intel = [
     "intel-extension-for-pytorch>=2.7.0",
+    "torch>=2.7.0,<2.9.0",
```
gsprochette, line 234: I get that it's tightly coupled with torch, but if intel-extension already declares it's torch dependency we shouldn't need to declare it here no? Same goes for torchvision

gsprochette (changes requested):
Thanks for this absolute gold PR. Already super good, I left some comments for little improvements here and there, almost ready already

tests/algorithms/testers/awq.py
begumcig: i thought slow was a wrong tag here, as awq is not time intensive like gptq for instance, but how do you feel?

pyproject.toml
begumcig: I see what you mean and I softly disagree. yes the version already defines which torch version it's supposed to use, but it doesn't define it in its dependencies. same for the torchvision version. So someone could install ipex with the absolutely wrong torch stack and get really cryptic errors. I think it's better to be explicit here.

gsprochette: ah, if the torch version is not in the package dependency then absolutely let's pin torch so that the intel install works

tests/algorithms/testers/upscale.py
gsprochette: I don't think we need the slow thing now do we? maybe we still want it or maybe it should be a comment stating why we ignore the cpu case... I'll let you decide

gsprochette (commented):
Looks great, let's merge it ! I just left one comment for you to decide about the "slow" mark on upscale :) 💅

gsprochette (approved):
Meant to approve, please refer to previous comment 🙃

PR #623, opened 2026-04-10
feat: integrate KVPress for KV cache compression (#366)

### LINKED ISSUE #366: [FEATURE] Integrate KVPress
## **✨ What You’ll Do**

Making the models faster, smaller, and greener is essential in Pruna, which is why we continually seek new ways to increase efficiency.

KV Press is a package prepared by NVIDIA, which  "packs the latest KV cache compression techniques, enabling memory-efficient long-context LLMs".

Integrate KV Press now in Pruna to make all these algorithms available.
 
## **🛠️ Dependency**

Add `kvpress` to the project dependencies.

## 🤖 Useful Resources

- https://github.com/NVIDIA/kvpress
- https://huggingface.co/blog/nvidia/kvpress

- [ ] Explore KV Press. Explore the pruna codebase for inspiration.
- [ ] Design the integration.
- [ ] Test Compatibility: Ensure it works with Pruna.
- [ ] Write the unit tests and test the integration.
- [ ] Update the documentation
---

## **✅ Acceptance Criteria**

- It follows the style guidelines.
- **Tests & Docs:** All existing and new unit tests pass, and the documentation is updated

---

## **❓ Questions?**

Feel free to jump into the [#contributing Discord channel](https://discord.com/invite/JFQmtFKCjd) if you hit any roadblocks. Can’t wait to see your contribution! 🚀

---

## Share on Socials

kschwethelm:
## Description
Integrate [KVPress](https://github.com/NVIDIA/kvpress) into Pruna, making 20 KV cache compression strategies available for causal language models. KVPress compresses the key-value cache during the prefill phase, reducing memory usage for long-context inference. 

Key implementation details:
  - New `kvpress` algorithm module following the `PrunaAlgorithmBase`
  pattern                                                              
  - Supports 20 scorer presses (ExpectedAttention, SnapKV, StreamingLLM, TOVA, KVzip, etc.)                                     
  - Configurable `compression_ratio` and `press_kwargs` for press-specific parameters                                            
  - New `KV_CACHER` algorithm tag for the cache compression category 
  - Compatibility defined with quantization algorithms (before) and torch_compile (after)                                                
  - Uses `reapply` save strategy — press is re-applied on model load   
                                                                       
**Excluded press types:** Wrapper presses (ChunkPress, AdaKVPress, PerLayerCompressionPress, DMSPress, etc.) are not included in this initial integration. These require a nested `ScorerPress` instance as a constructor argument, which doesn't fit the current single-class design. Similarly, ThinKPress is excluded as it compresses along the channel dimension with a different parameter interface. These could be added in a follow-up if needed.

Some downstream evaluation results are available in repo [kschwethelm/pruna-kvpress-eval](https://github.com/kschwethelm/pruna-kvpress-eval).

## Related Issue
Fixes #366

## Type of Change
- [ ] Documentation update

## Testing

Unit tests added in `tests/algorithms/test_kvpress.py` with a dedicated tester in `tests/algorithms/testers/kvpress.py`. Integration evaluated in a separate repo -> see [evaluation report](https://github.com/kschwethelm/pruna-kvpress-eval/blob/main/eval_results.pdf).  

## Checklist


minettekaum:
Hi @kschwethelm! Thanks for the contribution. This branch has conflicts with `main`. Could you rebase onto `main`, resolve the conflicts, and push the updated branch? Could you also review the test results and make any necessary changes? Let @sdiazlor or me know if you need any help :)

kschwethelm:
Hi @minettekaum, thanks! I've rebased onto main and resolved the conflicts. The pytest import error should also be resolved now. The test job failed with `ModuleNotFoundError: No module named 'kvpress'` because the base job doesn't install the `kvpress` extra

sdiazlor:
Thank you for the contribution @kschwethelm! The tests have passed now. We will review soon!

src/pruna/algorithms/kvpress.py
```diff
@@ -0,0 +1,181 @@
...
+    scores and prunes KV pairs after the prefill phase according to a chosen importance criterion.
+
+    This integration covers all scorer and standalone presses. Wrapper presses (e.g., ChunkPress,
+    AdaKVPress, PerLayerCompressionPress) that require a nested scorer press as input are not
+    included, as well as ThinKPress which compresses along the channel dimension with a different
+    parameter interface.
+    """
+
+    algorithm_name: str = "kvpress"
+    group_tags: list[tags] = [tags.KV_COMPRESSOR]
```
simlang: I was a bit on the line if we should introduce a new tag for this algorithm, but after you also called it pruning in the description, now i'm sure :D
i would include in the tags.PRUNER tag - this triggers a bunch of other changes in other places of the code
```suggestion
    group_tags: list[tags] = [tags.PRUNER]
```

src/pruna/algorithms/kvpress.py
```diff
@@ -0,0 +1,181 @@
...
+        press = press_cls(compression_ratio=compression_ratio, **press_kwargs)
+
+        original_generate = model.generate
+
+        @functools.wraps(original_generate)
+        def generate_with_press(*args, **kwargs):
+            with press(model):
+                return original_generate(*args, **kwargs)
+
+        model.generate = generate_with_press
```
simlang, line 167: [reading the original repo](https://github.com/NVIDIA/kvpress#why-not-using-modelgenerate-) they recommend wrapping the call with the context manager instead of the generate.
Is there a specific reason why you chose this way of implementing it?

tests/algorithms/test_kvpress.py
simlang, line 1: this file can be removed completely, since you are just testing if the kvpress library works, which if it wouldn't we couldn't change anyway

simlang: this should also a trigger few other changes where things were introduced for this :)

simlang (changes requested):
Thank you so much for tackling this integration. First iteration already looks great! 🚀

src/pruna/algorithms/kvpress.py
kschwethelm: Good point, agreed. I thought that tags are frequently used as groups in `compatible_before/compatible_after` lists. But since `PRUNER` is not used this way, it's ok. Fixed!

tests/algorithms/test_kvpress.py
kschwethelm: Agreed! I was a bit too defensive with my pytests here. Fixed!

src/pruna/algorithms/kvpress.py
kschwethelm: Thanks for the comment. I will investigate this.

kschwethelm: Could you clarify what you mean by "wrapping the call"? Do you mean 'KVPressTextGenerationPipeline'? That would allow separating context from question during prefill, but I decided against it because pruna’s 'generate()' doesn’t expose that split.

pyproject.toml
```diff
@@ -236,6 +237,9 @@ intel = [
     "torch>=2.7.0,<2.9.0",
     "torchvision>=0.22.0,<0.24.0",
 ]
+kvpress = [
```
gsprochette, line 252: Hi @kschwethelm thanks a lot for this PR and specifically for adding the dependency in an extra which is precisely where it should go :) We are currently in the process of documenting how these extras work, it's not yet in the contribution markdown so I'll just explain here what to do.

In order for your TestKVPress and TestKVPressSnapKV to run in the CI, you'll need to run them in a environment where the extra is actually installed. The following steps will create a job in the CI for this and make it run your tests:
1. first we need to mark your tests: create the `requires_kvpress` mark in `pruna/tests/conftest.py`'s `pytest_configure` function (you'll see similar marks that you can get inspiration from)
2. mark your tests with the mark you created: `@pytest.mark.requires_kvpress` for `TestKVPress` and `TestKVPressSnapKV`
3. in `pruna/.github/workflows/tests.yaml`, in the `test` job matrix, add a `'kvpress'` to the list, so it should be `["base", 'lmharness', 'rapidata', 'kvpress']`
4. in the include block right below that, tell the CI that the name `kvpress` should lead to installing the `--extra kvpress` and filtering for `requires_kvpress`: this installs an environment with the dependencies you need and runs your tests only
5. finally, we can skip the tests in the base environment which doesn't have the required dependencies: in the `'base'` part of the include block, add `not requires_kvpress` at the end of the `mark_filter` field

As I said, we're in the process of documenting this so please ping me if you encounter any difficulty regarding this or if there is an unclear step. This will help me improve the documentation :)
(pinging @begumcig also working on this)

src/pruna/algorithms/kvpress.py
```diff
@@ -0,0 +1,184 @@
...
+    scores and prunes KV pairs after the prefill phase according to a chosen importance criterion.
+
+    This integration covers all scorer and standalone presses. Wrapper presses (e.g., ChunkPress,
+    AdaKVPress, PerLayerCompressionPress) that require a nested scorer press as input are not
+    included, as well as ThinKPress which compresses along the channel dimension with a different
+    parameter interface.
+    """
+
+    algorithm_name: str = "kvpress"
+    group_tags: list[tags] = [tags.PRUNER]
```
llcnt, line 69: why not tagging as `KV_CACHER ` here ? :)

llcnt (changes requested):
I tested a combo `hqq+kvpress(snapkv)+torch_compile`, and it works well!
Thanks for the contribution, this is very cool :):)
Could you address [this](https://github.com/PrunaAI/pruna/pull/623#discussion_r3160069151) comment? After that we will be good to go ;)

pyproject.toml
kschwethelm: Thanks for the detailed guide! I actually had this exact setup in an earlier commit but removed it along with the test files per @simlang's review.

src/pruna/algorithms/kvpress.py
kschwethelm: I originally introduced a `KV_CACHER` tag (later renamed to `KV_COMPRESSOR`), but switched to `PRUNER` per @simlang's review, which fits better I think.

pyproject.toml
gsprochette: Hello @kschwethelm , yes this change is very recent so the instruction might have differed if they came a couple of weeks ago. While documenting the process I described above we found it slightly redundant, so we made it simpler in #653 so there are now fewer and less redundant steps. You can find the instructions for what you need to do in #654: the line you have here corresponds to the first sentence "Add a new extra [...]", so you should still
- "Define the dependency group [...]"
- "[...] set ``required_install`` on the algorithm class [...]"
- "Finally, register a matching ``requires_<extra>``"

Thanks in advance, and please let us know in #654 if these steps are unclear so we can improve the contribution guidelines :)

src/pruna/algorithms/kvpress.py
simlang: Sorry for being absent for such a long time. With wrapping the call, i meant wrapping the `__call__`/`forward` method of a Pipeline/Model.
I work basically exclusively with diffusers and not transformers, so the logic might be completely different in transformers, i just saw the paragraph in the kvpress repo, about not wrapping the `generate`.

Could it be, that if only the `generate` of the Pipeline backbone is wrapped, that there is a way to use the model which directly calls the backbone (via `__call__` or `forward`), which wouldn't be pressed then?

kschwethelm: Thanks for clarification! In transformers, the KV cache and next token prediction (NTP) loop is managed by `generate()`. A plain `forward()` call is just a single pass through the model. If someone manually loops `forward()` they need to manage their own KV cache. This case is outside pruna's scope and maybe not even supported by kvpress library.

The kvpress repo was just mentioning that they have their own special wrapper of `generate()`, where you can split the input prompt into "context" and "question" and only press KVs of context part. Later in the readme they mention that simply wrapping `generate()` is also fine but loses this special feature.

simlang: thanks for the clarification!

simlang (approved):
Looks good to me! Thank you so much for introducing KVPress to pruna!

LGTM! 🚀

llcnt (approved):
Thank you for the work :)

kschwethelm:
Thank you so much! It was fun to contribute and discuss with you :))

PR #630, opened 2026-04-15
fix: ping `peft >= 0.18.0, < 0.19.0` for torchao compatability issues

davidberenstein1957:
## Summary
- `peft>=0.18.1` added a strict `is_torchao_available()` check that **raises** `ImportError` (not returns `False`) when `torchao<0.16.0` is installed
- This broke any LoRA loading through peft (e.g. the `Hyper` algorithm on `stable_diffusion_v1_4`) — failing CI job: PrunaAI/pruna#545
- Removes the `<0.16.0` upper cap and raises the floor to `>=0.16.0`

## Root cause
Call chain: `Hyper._apply` → `model.load_lora_weights` → peft `dispatch_torchao` → `is_torchao_available()` → `ImportError: Found version 0.15.0, but only versions above 0.16.0 are supported`

The original cap was added for a torchao+torch_compile+diffusers interaction. That issue is CPU/torch_compile-specific and does not affect the base test suite. torchao 0.17.0 Python API is compatible with `torch>=2.8.0` (our CI uses torch 2.9.1).

## Test Plan
- [ ] CI `test (3.11, base)` passes — specifically `TestHyper_stable_diffusion_v1_4-cpu`
- [ ] `TestTorchao_sd_tiny_random-cpu` and `TestTorchao_flux_tiny_random-cpu` still pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,7 +148,7 @@ dependencies = [
     "aenum",
     "imageio-ffmpeg",
     "jaxtyping",
-    "peft>=0.18.0",
+    "peft>=0.18.0,<0.19.0",
     "trl<=0.21.0",
     "termcolor==2.3.0",
     "realesrgan"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,6 +148,8 @@ dependencies = [
     "aenum",
     "imageio-ffmpeg",
     "jaxtyping",
+    # peft 0.19+ enforces torchao>=0.16 in PEFT (is_torchao_available); keep <0.19 while core torchao stays <0.16.
+    # https://github.com/huggingface/peft/blob/6d5a6f4f2f902dbf13d21d2661d57c3c05df1dae/src/peft/import_utils.py#L131
     "peft>=0.18.0,<0.19.0",
     "trl<=0.21.0",
     "termcolor==2.3.0",


pyproject.toml
```diff
@@ -140,7 +140,7 @@ dependencies = [
     "bitsandbytes; sys_platform != 'darwin' or platform_machine != 'arm64'",
     "optimum-quanto>=0.2.5",
     "hqq==0.2.7.post1",
-    "torchao>=0.12.0,<0.16.0", # 0.16.0 breaks diffusers 0.36.0, torch+torch: https://github.com/pytorch/ao/issues/2919#issue-3375688762
+    "torchao>=0.16.0", # peft>=0.18.1 requires torchao>=0.16.0 (raises ImportError otherwise); Python API supports torch>=2.8.0
```
davidberenstein1957:
```suggestion
    "torchao>=0.16.0", # peft>=0.18.1 requires torchao>=0.16.0
```

begumcig (changes requested):
Thanks for the PR @davidberenstein1957 

I think we currently don’t support torchao > 0.16. When we tested newer versions, our nightlies started failing (@gsprochette would know better about this than me), so we still need to properly investigate compatibility. Also even if we supported the newer torchao, pinning it >=0.16.0 would break the torch versions we support, because they are also tightly coupled.

From my perspective, the next steps would be:

- Investigate whether we can support torchao > 0.16.
- Introduce a new extra (aligned with our dependency separation strategy in Pruna), e.g. recovery, that includes the required versions for peft, trl, termcolor, and torchao. We should also guard all module-level imports of these packages across the codebase (tests and recovery).

Since the nightlies are currently failing and we need a quick fix, I think we can temporarily pin peft to:

peft >= 0.18.0, < 0.19.0

In the PR description, you mentioned that the issue comes from peft>=0.18.1, but after checking the source code, it looks like the breaking change was actually introduced in 0.19.0:
https://github.com/huggingface/peft/blob/6d5a6f4f2f902dbf13d21d2661d57c3c05df1dae/src/peft/import_utils.py#L131
Please correct me if I’m mistaken here.

But anyway, I think overall, pinning peft to a version range above 0.18.0 and below the first version that introduces the torchao incompatibility should be a reasonable short-term fix, until we can properly support newer torchao versions. What do you think?

davidberenstein1957:
Hi @begumcig,

I have not investigated the nightlies. @gsprochette is still on holiday, right? Let's, for now, go for the easy fix and pin peft to `peft >= 0.18.0, < 0.19.0`.

begumcig (approved):
Looks super good to me !! Thank you so much David

PR #645, opened 2026-04-28
feat(text-metrics): split qa_accuracy

davidberenstein1957:
## Summary
Splits `qa_accuracy` into its own stacked PR, adds `QAAccuracyMetric`, and wires GenEval to `qa_accuracy + clip_score`.

## Stack Position
- Base: PR #638 (`feat/vlm-pr-2-infrastructure`)
- Next: PR #646 (`feat/vlm-pr-3b-oneig-alignment`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/metric_qa_accuracy.py`
- `src/pruna/evaluation/benchmarks.py`

## Test Plan
```bash
uv run pytest tests/evaluation/test_text_metrics.py -k qa_accuracy
```

## Review Focus
- Aggregation behavior (`all_or_nothing`)
- GenEval benchmark wiring

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (3/10)
- Review after PR #638.
- Next PR to review: #646.
- Confirm this PR's tests and scope before continuing.

PR #641, opened 2026-04-25
feat(e2e-tests): stacked e2e after split metrics

davidberenstein1957:
## Summary
Final integration/e2e PR for the split VLM metrics stack.

Keeps e2e and integration-focused coverage isolated from per-metric implementation PRs.

## Stack Position
- Base: PR #651 (`feat/vlm-pr-4c-img-edit-score`)
- This is the stack tip / final integration PR
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Full Stack Order
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests (this PR)

## Files
- `tests/evaluation/test_vlm_e2e.py`
- `tests/evaluation/test_task.py`
- `tests/data/test_datamodule.py`
- `tests/evaluation/_vlm_batch_snapshot_helpers.py`

## Test Plan
```bash
uv run pytest tests/evaluation/test_vlm_e2e.py tests/evaluation/test_task.py tests/data/test_datamodule.py
```

## Review Focus
- End-to-end benchmark execution coverage
- Task/datamodule integration paths
- Split-stack parity with umbrella behavior

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (10/10)
- Review after PR #651.
- This is the final PR in the stack.
- Confirm this PR's tests and scope before continuing.

PR #638, opened 2026-04-25
feat(infrastructure): add VLM base classes and utilities

davidberenstein1957:
## Summary
Adds the VLM inference infrastructure used by all downstream VLM judge metrics:
- `BaseVLM`
- `LitellmVLM`
- `TransformersVLM`
- `StatefulVLMMeanScoresMetric`
- shared batch/device helpers

## Stack Position
- Base: PR #637 (`feat/vlm-pr-1-vendor`)
- Next: PR #645 (`feat/vlm-pr-3a-qa-accuracy`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/vlm_base.py`
- `src/pruna/evaluation/metrics/vlm_utils.py`
- `tests/evaluation/test_vlm_base_infrastructure.py`
- `src/pruna/evaluation/metrics/utils.py`
- `src/pruna/evaluation/metrics/__init__.py`
- `pyproject.toml`

## Alignment Notes
This PR is intentionally based on `feat/vlm-pr-1-vendor` so reviewers only see infrastructure delta.

## Test Plan
```bash
uv run pytest tests/evaluation/test_vlm_base_infrastructure.py -v
```

## Review Focus
- API/local VLM abstraction boundaries
- Device handling and batching behavior
- Stateful aggregation correctness

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (2/10)
- Review after PR #637.
- Next PR to review: #645.
- Confirm this PR's tests and scope before continuing.

PR #637, opened 2026-04-25
feat(vendor): add LLM2Vec embedding model

davidberenstein1957:
## Summary
Vendors OneIG's `LLM2Vec` implementation used by downstream OneIG-based VLM metrics.

This PR includes two minimal vendor compatibility fixes:
- honor constructor-provided `doc_max_length`
- honor `convert_to_numpy=True` in `encode()`

## Stack Position
- Base: `main`
- Next: PR #638 (`feat/vlm-pr-2-infrastructure`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/vendor/NOTICE.oneig_llm2vec`
- `src/pruna/evaluation/metrics/vendor/oneig_llm2vec/llm2vec.py`
- `src/pruna/evaluation/metrics/vendor/oneig_llm2vec/modeling_llama_encoder.py`
- `src/pruna/evaluation/metrics/vendor/oneig_llm2vec/models/bidirectional_llama.py`

## Test Plan
```bash
python -c "from pruna.evaluation.metrics.vendor.oneig_llm2vec.llm2vec import LLM2Vec; print('OK')"
```

## Review Focus
- License and attribution completeness
- Correctness of the two vendor compatibility fixes
- No coupling to Pruna internals

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (1/10)
- Start here (base of the stack).
- Next PR to review: #638.
- Confirm this PR's tests and scope before continuing.

PR #646, opened 2026-04-28
feat(text-metrics): split oneig_alignment

davidberenstein1957:
## Summary
Splits `oneig_alignment` into its own stacked PR, adds `OneIGAlignmentMetric`, and wires OneIG alignment subset benchmarks.

## Stack Position
- Base: PR #645 (`feat/vlm-pr-3a-qa-accuracy`)
- Next: PR #647 (`feat/vlm-pr-3c-text-score-pair`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/metric_oneig_alignment.py`
- `src/pruna/evaluation/benchmarks.py`
- `tests/evaluation/test_text_metrics.py`

## Test Plan
```bash
uv run pytest tests/evaluation/test_text_metrics.py -k oneig_alignment
```

## Review Focus
- Dependency-mask semantics
- OneIG subset benchmark mapping

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (4/10)
- Review after PR #645.
- Next PR to review: #647.
- Confirm this PR's tests and scope before continuing.


src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
...
+            Unused batch metadata (kept for metric interface).
+        gt : torch.Tensor
+            Ground-truth slot holding per-sample aux dicts with ``questions`` and
+            optionally ``dependencies``.
+        outputs : torch.Tensor
+            Model outputs (images) evaluated against the questions.
+        """
+        inputs = metric_data_processor(x, gt, outputs, self.call_type)
+        images = _process_images(inputs[0])
+        aux_list = inputs[1] if len(inputs) > 1 else []
```
begumcig: if the call type is y_gt i don't see this ever not having len>1

src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
...
+        if isinstance(aux_list, torch.Tensor):
```
begumcig: i am a bit confused by this, if the aux_list is a torch.Tensor we turn it into a list, but below we are trying to index it, and we are expecting it to be a list of dicts. How does a tensor turn into a list of dicts? Maybe if it's a tensor, it's already not what we are looking for?

src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
...
+            qids = _active_oneig_question_ids(qmap)
+            if not qids:
+                self.scores.append(0.0)
+                continue
+            question_texts = [str(qmap[qi]) for qi in qids]
+            deps = _normalize_dependencies(aux.get("dependencies", {}))
+            raw_scores_list = self.vlm.score(
+                [image] * len(question_texts),
+                question_texts,
+                ["Yes"] * len(question_texts),
```
begumcig: I am asking this just out of curiosity, I also don't know the answer: Can the expected answer here be "No" instead of "yes", or perhaps a number, etc?

src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
...
+            model_name="HuggingFaceTB/SmolVLM-256M-Instruct",
+            device="cpu",
+            vlm_kwargs={"model_load_kwargs": {"torch_dtype": torch.float32}},
+        )
+    """
+
+    metric_name: str = "oneig_alignment"
+    metric_units: str = "alignment"
+
+    def update(self, x: list[Any] | torch.Tensor, gt: torch.Tensor, outputs: torch.Tensor) -> None:
```
begumcig, line 260: This is more of a general comment but I noticed the base class for this `QAAccuracymetric` also implements many of the lines in its update function (processing the input, checking if questions exist in the data, calling the vlm.score function). If these are genuienly overlapping for many metrics and only the middle "computation" part is differing from metric to metric, maybe we should make these functions standalone in the base class and resuse them in the child classes?

src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
```
begumcig, line 1: Thank you so much for the refactoring David! I think this is much closer to the original implementation for OneIG Alignment! A couple of things I have noticed that perhaps we can discuss if we want to address them or not: 

- In the original Alignment score, I believe for each prompt we generate a grid of images (default being 2x2) So for each prompt, the VLM sees multiple images, factoring for problems that can come from the seed. I believe here we process images one by one, which is fine by me, but perhaps we should address this in the docstring or in the documentation? 
- Also related to the above topic a bit, I think in the original implementation, the questions for each prompt is passed to the VLM one by one (so if a prompt has 10 questions for that generation, each is one request). Please correct me if I am wrong (maybe the VLM PR handles these requests one by one) but here we seem to pass these questions all together. 

So in the original implementation we evaluate multiple images of the same prompt multiple times, and in our implementation we are evaluating one image at the same time with multiple questions. I think this may create some discrepancies in how the scores are shaped quite a lot, maybe at least we should also evaluate the questions one by one?

src/pruna/evaluation/metrics/metric_oneig_alignment.py
```diff
@@ -0,0 +1,234 @@
...
+#
```
begumcig, line 2: This is not really an issue at all, but I think the original implementation uses a Qwen model for scoring and I think our VLM implementation usually defaults to GPT 4-o? This is totally fine by me, but maybe we should again note it somewhere?

src/pruna/evaluation/metrics/metric_qa_accuracy.py
begumcig, line 79: if kwargs only support `aggegation` right now, maybe we should also make it a keyword argument?

begumcig (changes requested):
Thank you so much David this is really super close to the original implementation right now! Thank you so much for implementing the dependency masks and per prompt averaging! Just added a few nitpicky details and some general questions about the metric logic but this already looks super good to me!! Also I noticed the tests haven't run for this PR, do you have any idea why? (This is probably something I should know as the package team lol) Thanks again :)

begumcig (changes requested):
Something is wrong with the reviewing so i have to make this placeholder comment

davidberenstein1957:
## Review follow-up (@begumcig)

Implemented the paper-faithful alignment path:

- **2×2 grid** via `split_mxn_grid` (default `grid_size=(2, 2)`), mean across cells
- **One question per VLM call** (all grid cells scored for that question)
- **Default VLM:** `Qwen/Qwen2.5-VL-7B-Instruct` (`transformers`); GPT-4o still available via `vlm_type=litellm`
- **Aux:** removed tensor→tolist; explicit `list[dict]` validation
- **P(Yes) scoring:** documented in docstring + inline comment at the `["Yes"]` call
- **Shared refactor:** good idea for a follow-up refactor but for now I decided to split this up in separate PRs for the overview. Perhaps we can merge them later in a separate PR.
- **CI:** enabled on feature-branch PR targets in #638 (remove before merge to main)

Dataset loaders / Multilingualism Q_D remain in #641.

Please re-review when convenient.

PR #647, opened 2026-04-28
feat(text-metrics): split text_score pair

davidberenstein1957:
## Summary
Splits the `text_score` pair into a dedicated stacked PR, adds OCR-based text metrics and shared utilities, and wires Long Text Bench + OneIG Text Rendering.

## Stack Position
- Base: PR #646 (`feat/vlm-pr-3b-oneig-alignment`)
- Next: PR #648 (`feat/vlm-pr-3d-oneig-reasoning`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/metric_text_score.py`
- `src/pruna/evaluation/metrics/metric_text_score_utils.py`
- `src/pruna/evaluation/benchmarks.py`

## Test Plan
```bash
uv run pytest tests/evaluation/test_text_metrics.py -k text_score
```

## Review Focus
- OCR scoring behavior
- Long Text Bench and OneIG Text Rendering wiring

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (5/10)
- Review after PR #646.
- Next PR to review: #648.
- Confirm this PR's tests and scope before continuing.

PR #648, opened 2026-04-28
feat(text-metrics): split oneig_reasoning

davidberenstein1957:
## Summary
Splits `oneig_reasoning` into its own stacked PR, adds `OneIGReasoningMetric`, and introduces lazy `MetricRegistry` loading required by this metric family.

## Stack Position
- Base: PR #647 (`feat/vlm-pr-3c-text-score-pair`)
- Next: PR #649 (`feat/vlm-pr-4a-vqa`)
- Final integration: PR #641 (`feat/vlm-pr-5-e2e-tests`)
- Canonical umbrella reference: PR #545 (`feat/metrics-vlm-support`)

## Files
- `src/pruna/evaluation/metrics/metric_oneig_reasoning.py`
- `src/pruna/evaluation/metrics/registry.py`
- `src/pruna/evaluation/benchmarks.py`
- `tests/evaluation/test_text_metrics.py`

## Test Plan
```bash
uv run pytest tests/evaluation/test_text_metrics.py -k oneig_reasoning
```

## Review Focus
- Reasoning score extraction and averaging
- Lazy metric loading behavior in registry

## Review Flow (Order)
Review the stack in this exact order:
1. #637 vendor
2. #638 infrastructure
3. #645 qa_accuracy
4. #646 oneig_alignment
5. #647 text_score pair
6. #648 oneig_reasoning
7. #649 vqa
8. #650 vie_score
9. #651 img_edit_score
10. #641 e2e tests

### This PR in the flow (6/10)
- Review after PR #647.
- Next PR to review: #649.
- Confirm this PR's tests and scope before continuing.


tests/evaluation/test_text_metrics.py
```diff
@@ -134,3 +139,90 @@ def test_to_oneig_record_strips_null_questions_and_dependencies() -> None:
...
+        metric.compute()
+
+
+@pytest.mark.cpu
+def test_oneig_reasoning_has_metric_registered() -> None:
+    """oneig_reasoning is available via MetricRegistry (lazy)."""
+    assert MetricRegistry.has_metric("oneig_reasoning")
+
+
+@pytest.mark.cpu
```
begumcig: This is probably already handled in another PR, but if CLIPLLM stricly needs transformers <5.0.0 maybe we should define it as an extra in the toml file with its required dependencies?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
```
begumcig, line 1: This is more of a general question from my side, especially because I previously reviewed #646.

It seems to me that the two OneIG-related metrics are currently taking slightly different approaches. In #646, we use a generic VLM/LiteLLM wrapper, which I assumed was mainly to make model loading and model calls easier for users. However, by using a more generic setup and not hardcoding the benchmark model itself, we are missing some OneIG-Alignment specific requirements, such as using the Qwen-VL model.

This PR, on the other hand, seems much closer to the original reasoning setup because it uses the 8B LLM2CLIP model. I really really like that it follows the benchmark more faithfully. At the same time, I wonder if the original benchmark model should be the default rather than the only option. For example, we could expose the judge/model as a parameter, so users running on consumer hardware could choose a simpler CLIP model if needed.

Maybe this is also something I should bring up again in #646, but if we are moving toward a more faithful OneIG implementation here, then perhaps the alignment metric should also use the model from the original benchmark by default.

Overall, I am wondering whether we should align the two implementations around the same principle: default to the original benchmark setup for faithfulness, but allow lighter alternatives through configuration for accessibility (So we can have the best of both worlds in a consistent way)

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+        set_verbosity_debug()
+        enable_progress_bars()
+
+    if not _env_truthy(os.environ.get("PRUNA_ONEIG_HF_FAST_DOWNLOAD")):
+        return
+
+    import hf_transfer  # noqa: F401  # type: ignore[import-not-found]
+    import huggingface_hub.constants as hf_constants
+
+    hf_constants.HF_HUB_ENABLE_HF_TRANSFER = True
```
begumcig, line 110: I have never done this before so I don't really know, but is it chill to change user's env variables without asking them? Perhaps we should  keep the original state in a parameter and implement a reset function that reverts this at least at the end of execution?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+    pruna_logger.info("oneig_reasoning: enabled hf_transfer downloads (PRUNA_ONEIG_HF_FAST_DOWNLOAD=1).")
+
+
+def _to_pil_list(images: list) -> list:
```
begumcig, line 114: This function is super useful! I can see it used it in many places, so perhaps we should move it to a more general utilities function?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+    """Convert images to list of PIL.Image (RGB)."""
+    import numpy as np
+    from PIL import Image
+
+    out: list = []
+    for img in images:
+        if isinstance(img, Image.Image):
+            out.append(img.convert("RGB"))
+        elif isinstance(img, torch.Tensor):
+            if img.ndim == 4:
```
begumcig, line 124: sorry super dumb question but i am a bit confused by this, why do we only use the first image if it's a batch? Or is that not what it's doing?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+
+    Accepts PIL images and a single answer string; returns per-image scores.
+    Best-effort alignment with OneIG-Benchmark scripts (CUDA + bfloat16).
+    """
+
+    def __init__(
+        self,
+        processor_model: str = "openai/clip-vit-large-patch14-336",
+        model_name: str = "microsoft/LLM2CLIP-Openai-L-14-336",
+        llm_model_name: str = "microsoft/LLM2CLIP-Llama-3-8B-Instruct-CC-Finetuned",
```
begumcig, line 151: I also mentioned it in the general comment above but perhaps we should also support smaller models, or perhaps a quantized version of this model for consumer devices?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+        self._clip_model = AutoModel.from_pretrained(
+            self.model_name,
+            dtype=dtype,
+            trust_remote_code=True,
+        ).to(self.device)
+        self._clip_model.train(mode=False)
+
+        config = AutoConfig.from_pretrained(self.llm_model_name, trust_remote_code=True)
+        dev_str = str(self.device)
+        attn_impl = "sdpa" if dev_str == "cuda" or dev_str.startswith("cuda:") else "eager"
```
begumcig: what exactly is happening here, are we trying to switch to a more efficient attention implementation if we are on a gpu?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
begumcig, line 151: I also mentioned it in the general comment above but perhaps we should also support smaller models, or perhaps a quantized version of this model for consumer devices?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
begumcig: what exactly is happening here, are we trying to switch to a more efficient attention implementation if we are on a gpu?

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+        from pruna.evaluation.metrics.vendor.oneig_llm2vec.modeling_llama_encoder import LlamaEncoderModel
+
+        pruna_logger.info(
+            "oneig_reasoning: downloading or loading LLM2CLIP checkpoints "
+            "(%s, %s). First run can take many minutes and several gigabytes; "
+            "Hugging Face download progress may look idle when logs are piped.",
+            self.model_name,
+            self.llm_model_name,
+        )
+        dtype = torch.bfloat16 if self.device == "cuda" else torch.float32
```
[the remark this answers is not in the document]
begumcig: agreed, perhaps it should check 
```suggestion
        dtype = torch.bfloat16 if str(self.device).startswith("cuda") else torch.float32
```

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+
+    def _get_scorer(self) -> _LLM2CLIPScorer:
+        if self._scorer is not None:
+            return self._scorer
+        return _LLM2CLIPScorer(
+            processor_model=self.processor_model,
+            model_name=self.model_name,
+            llm_model_name=self.llm_model_name,
+            device=self.device,
+        )
```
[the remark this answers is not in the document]
begumcig: Yes, I completely agree with this one, If we reload the model to the memory for every processed input, we would quickly get OOM. I think we should also add deleting the model to the reset() and free up this model once the calculations are finished.

begumcig (changes requested):
Super cool David! Some minor comments only :)

src/pruna/evaluation/metrics/metric_oneig_reasoning.py
```diff
@@ -0,0 +1,357 @@
...
+        with torch.no_grad():
+            if self.device == "cuda":
+                with torch.amp.autocast(device_type="cuda"):
+                    image_features = self._clip_model.get_image_features(input_pixels)
+            else:
+                image_features = self._clip_model.get_image_features(input_pixels.float())
+
+        image_features = image_features.float()
+        text_features = text_features.float()
+        image_features /= image_features.norm(dim=-1, keepdim=True)
```
begumcig: should we add an alpha here in the denominator to prevent division by 0

davidberenstein1957:
## Review follow-up (@begumcig)

- **Grid split:** same `split_mxn_grid` path as alignment (default 2×2), mean cell scores
- **Scorer cache + `reset()`:** LLM2CLIP scorer cached; cleared on reset
- **`cuda:0` / dtype:** `str(device).startswith("cuda")` for autocast/dtype
- **`torch.no_grad()`:** text + image feature paths
- **Epsilon:** `1e-8` on feature norms
- **Env restore:** HF download env wrapped with save/restore
- **`attn_implementation`:** optional kwarg (`None` = HF default; Pruna defaults to `sdpa` on CUDA when unset — upstream OneIG does not set this)
- **Model IDs:** all three checkpoints configurable via constructor args
- **Alignment / reasoning defaults:** both paper-faithful (Qwen2.5-VL + LLM2CLIP); API overrides documented

Pushed on rebased branch.

PR #660, opened 2026-05-12
build: add dependencies cooldown

gsprochette:
## Description
This PR adds a 1 week cooldown for dependencies to protect against compromised packages. This gives us a 1 week delay to get any news about compromised packages and
- confirm that they have removed the compromised version, or
- restrict the version until we have found a better solution.

## Related Issue

Fixes #(issue number)

## Type of Change

## How Has This Been Tested?
I checked the uv pip install behavior for torch 2.11 dating back from march 23 (<2 months ago) with a no filter (is the version selected), 1 week filter (no change), 8 weeks filter (torch 2.10 is selected instead).

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,8 @@ explicit = true
 [tool.uv]
 index-strategy = "first-index"
 exclude-newer = "1 week"  # protection against compromised dependencies
+# trusted dev wheels that are missing an upload date
+exclude-newer-package = { gptqmodel = false, "stable-fast-pruna" = false }
 
 conflicts = [
     [{ extra = "awq" }, { extra = "vbench" }],
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ explicit = true
 
 [tool.uv]
 index-strategy = "first-index"
+exclude-newer = "1 week"  # protection against compromised dependencies
 
 conflicts = [
     [{ extra = "awq" }, { extra = "vbench" }],


SaboniAmine (approved):

pyproject.toml
```diff
@@ -77,6 +77,7 @@ explicit = true
 
 [tool.uv]
 index-strategy = "first-index"
+exclude-newer = "1 week"  # protection against compromised dependencies
```
begumcig, line 76: what do we mean by compromised here? Are we getting the all newer packages with a delay regardless?

begumcig (commented):
Not deny not approve but a secret third thing (a question)

pyproject.toml
SaboniAmine: Yes, this is to avoid getting versions where the compromission isn't already identified.
For those which are identified as poisonned, if the patch isn't shipped in 1 week we can pin the max versions that is safe to use.

gsprochette: Exactly as Amine said, "compromised" here means that there may be a security issue with the latest version of one of the dependencies. If this is the case, the cooldown introduces a 1 week delay giving the community time to spot it and patch it, or giving us the possibility to pin the version to a non-compromised one if the patch is not available after this one week period

begumcig: Okay that makes sense, thank you!

begumcig (approved):
LGTM, thanks a lot <3

PR #673, opened 2026-06-05
fix: torch compile default to inductor on cpu

gsprochette:
## Description
This PR makes inductor the default for torch compile on CPU. Doing this moves away from openvino which fails on some architectures due to some known shape checks in openvino.

With a very quick benchmark of `torchvision.models.vit_b_16` on CPU with both torch 2.11 and 2.12, the compiled model with inductor backend also runs slightly faster than the openvino one (41ms with 0.3ms uncertainty for openvino vs 39ms with 0.2ms uncertainty for inductor, on torch 2.12). Hence this change should not affect performance, or even improve them.

## Related Issue

Fixes tests/algorithms/test_algorithms.py::test_full_integration[TestTorchCompile_shufflenet-cpu]

## Type of Change

- [ ] Documentation update

## Testing

The tests/algorithms/test_algorithms.py::test_full_integration[TestTorchCompile_shufflenet-cpu] usually only passes because of the retry, which for some reason uses cached states on the second run avoiding the faulty shape assumptions. This can be reproduced with the following bash code that needs to run at the root of the pruna project, with your venv activated:
```
set -euo pipefail

# Should be run from the root of the pruna repository.
if [[ "$(basename "$PWD")" != "pruna" ]]; then
  echo "Error: This script must be run from the 'pruna' directory." >&2
  exit 1
fi
if [[ ! -d "./src/pruna" ]]; then
  echo "Error: './src/pruna' directory does not exist. Are you in the correct directory?" >&2
  exit 1
fi

# Clean up compilation artifacts
USER_TAG="$(whoami | tr '/:*?"<>|\\' '_')"
INDUCTOR_CACHE="${TORCHINDUCTOR_CACHE_DIR:-/tmp/torchinductor_${USER_TAG}}"
rm -rf "$INDUCTOR_CACHE"

# This test fails with a backend error on openvino (something about list index out of range, symbol table,
# or s7 symbol, depending on the torch version).
pytest -o addopts="" tests/algorithms/test_algorithms.py::test_full_integration[TestTorchCompile_shufflenet-cpu]
```

This test reliably fails on main for me with different versions of torch (+cu126 versions), and passes with the torch inductor switch.

## Checklist

---

> **Note:** 

### DIFF
--- a/docs/tutorials/cv_cpu.ipynb
+++ b/docs/tutorials/cv_cpu.ipynb
@@ -1,177 +1,177 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "# Smash your model with a CPU only"
-   ]
-  },
-  {
-   "cell_type": "raw",
-   "metadata": {
-    "vscode": {
-     "languageId": "raw"
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Smash your model with a CPU only"
+      ]
+    },
+    {
+      "cell_type": "raw",
+      "metadata": {
+        "vscode": {
+          "languageId": "raw"
+        }
+      },
+      "source": [
+        "<a target=\"_blank\" href=\"https://colab.research.google.com/github/PrunaAI/pruna/blob/v|version|/docs/tutorials/cv_cpu.ipynb\">\n",
+        "    <img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/>\n",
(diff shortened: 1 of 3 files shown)


begumcig:
Why were we using openvino in the first place? isn't it only optimal for intel?

begumcig (approved):
Well it makes sense to me that you even got a slightly better performance (assuming you ran this on your company laptop or your workstation, neither uses intel cpus), I don't think our default should be openvino at all and thank you for noticing and changing this!

begumcig:
Rerunning the tests that failed due to an internet error, let's wait for them to pass then merge away!!! thanks again <3

PR #674, opened 2026-06-06
build: switch back to unsafe-best-match for easy torch install

gsprochette:
## Description
This PR re-introduces the "unsafe-best-match" index resolution strategy that was replaced by "first-index" in #613

The reason for this is that torch used to ship some packages on its index. With first-index, uv only see those outdated versions, the first-index, and doesn't look on pypi for newer versions, leading to an unresolvable environment.

The reason for moving away in the first place was to avoid pinging the pythonanywhere index on every public package. However, this was already accomplished by the line "explicit = true" in the same PR. With explicit=true, the python-anywhere index is only checked when explicitely stated, as in `stable-fast-pruna = { index = "pruna_internal", extra = "stable-fast-extraindex" }`.

With the change in this PR, we should be able to run `uv pip install -e .[dev] torch==2.12 torchvision --extra-index-url https://download.pytorch.org/whl/cu126` (currently fails), without pinging the python-anywhere index on every package causing the HTTP errors.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

running `uv pip install -e .[dev] torch==2.12 torchvision --extra-index-url https://download.pytorch.org/whl/cu126` currently fails, but passes after this change.

## Checklist

---

> **Note:** 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ url = "https://prunaai.pythonanywhere.com/simple/"
 explicit = true
 
 [tool.uv]
-index-strategy = "first-index"
+index-strategy = "unsafe-best-match"
 
 conflicts = [
     [{ extra = "awq" }, { extra = "vbench" }],


begumcig (commented):
I thought the reason we switched to `first-index` was to prioritize PyPI when installing packages. With `unsafe-best-match`, we were searching the entire PythonAnywhere index, which was causing issues.

Now, my understanding is that `explicit = true` already prevents uv from looking at those custom indexes unless a package is explicitly assigned to them. Also, `first-index` makes index lookup greedy, which is not necessarily what we want. Is that correct? If that's the case APPROVED

begumcig (approved):
forgot to approve

gsprochette:
This is exactly correct ! Here's a recap:
- index strategy:
  - unsafe-best-match: search every index for any version, starting with the default pypi. This is also the one that reproduces pip's behavior.
  - first-index: same, but whenever at least one version is found, don't bother looking in the remaining indices
- explicit: if true, this index is a candidate only when explicitely stated with `explicit = true`, in which case it will be the only candidate.

PR #682, opened 2026-06-15
ci: add stale issue handling

begumcig:
## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

---

> **Note:** 

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -13,21 +13,27 @@ on:
 jobs:
   stale:
     runs-on: ubuntu-latest
+    env:
+      DAYS_BEFORE_ISSUE_STALE: 21
+      DAYS_BEFORE_ISSUE_CLOSE: 14
+      DAYS_BEFORE_PR_STALE: 10
+      DAYS_BEFORE_PR_CLOSE: 7
     steps:
       - uses: actions/stale@v9
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
           # Issues
-          days-before-issue-stale: 30
-          days-before-issue-close: 7
+          days-before-issue-stale: ${{ env.DAYS_BEFORE_ISSUE_STALE }}
+          days-before-issue-close: ${{ env.DAYS_BEFORE_ISSUE_CLOSE }}
           stale-issue-label: 'stale'
-          stale-issue-message: 'This issue has been inactive for 30 days and is now marked as stale. It will be closed in 7 days if there is no further activity.'
+          stale-issue-message: 'This issue has been inactive for ${{ env.DAYS_BEFORE_ISSUE_STALE }} days and is now marked as stale. It will be closed in ${{ env.DAYS_BEFORE_ISSUE_CLOSE }} days if there is no further activity.'
           close-issue-message: 'Closing this issue due to inactivity. Please reopen it or create a new issue if this is still relevant.'
           exempt-issue-labels: 'pinned,security,do-not-close'
 
           # PRs
-          days-before-pr-stale: 10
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yaml
```diff
@@ -15,5 +17,17 @@ jobs:
       - uses: actions/stale@v9
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Issues
+          days-before-issue-stale: 30
+          days-before-issue-close: 7
+          stale-issue-label: 'stale'
+          stale-issue-message: 'This issue has been inactive for 30 days and is now marked as stale. It will be closed in 7 days if there is no further activity.'
```
gsprochette: I don't suppose that we can reuse the variables instead of writing 30 and 7 in plain text?

.github/workflows/stale.yaml
```diff
@@ -15,5 +17,17 @@ jobs:
       - uses: actions/stale@v9
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+          # Issues
+          days-before-issue-stale: 30
+          days-before-issue-close: 7
```
gsprochette: 7 seems quick compared to 30 for marking stale. I would prefer issues to be marked stale quicker but we have more time to un-stale them if it's by mistake. Like 3 weeks before stale and 2 weeks before close. WDYT?

.github/workflows/stale.yaml
```diff
@@ -15,5 +17,17 @@ jobs:
...
+          stale-issue-label: 'stale'
+          stale-issue-message: 'This issue has been inactive for 30 days and is now marked as stale. It will be closed in 7 days if there is no further activity.'
+          close-issue-message: 'Closing this issue due to inactivity. Please reopen it or create a new issue if this is still relevant.'
+          exempt-issue-labels: 'pinned,security,do-not-close'
+
+          # PRs
           days-before-pr-stale: 10
-          stale-pr-message: 'This PR has been inactive for 10 days and is now marked as stale.'
+          days-before-pr-close: 7
```
gsprochette: 7 days seems super quick for closing a PR, can we make it 2 weeks? In case someone is on vacation or something. Stale should already be a very strong message...

.github/workflows/stale.yaml
```diff
@@ -15,5 +17,17 @@ jobs:
...
+          stale-pr-label: 'stale'
+          stale-pr-message: 'This PR has been inactive for 10 days and is now marked as stale.'
```
gsprochette: There should be a warning about the PR being closed in x days. Also, in the case where we can reuse the values from the stale-?-label and stale?-message, let's use those.

gsprochette (changes requested):
Great work let's merge super quick, but before that I have a couple of questions about the delays and warning messages.

.github/workflows/stale.yaml
begumcig: Ah I think the default is 7 so it was already 7 before. I just chose to make it more visible. We can change it to be 2 weeks, I agree I have lost some PRs then had to chase them :D

.github/workflows/stale.yaml
begumcig: yes indeed!!!

.github/workflows/stale.yaml
gsprochette: I see, let's keep it the way it was, if we feel later on that this is causing problem we'll change it then :+1:

gsprochette (approved):
Perfect ✨ thanks for addressing the comments 💅

PR #684, opened 2026-06-15
fix: update moe tiny model for testing (transformers 5)

sdiazlor:
## Description

Current MoE implementation has been modified (https://github.com/huggingface/transformers/blob/main/src/transformers/integrations/moe.py). Now, it seems that some MoE layers expects certain tensor strides to be 16-byte as they're using torch grouped_mm, which makes it stricter.

Previous model has really small dimensions (2-4), producing errors when loading in bf16.

I saw you had already some compatibility issue (https://github.com/PrunaAI/pruna/pull/533). So, not sure which are your thoughts about it.

```py
FAILED tests/algorithms/test_algorithms.py::test_full_integration[TestReduceNOE_qwen_moe_tiny_random-cuda] - RuntimeError: strides should be multiple of 16 bytes
```

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactor (no functional change)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Testing

- [ ] I added or updated tests covering my changes
- [ ] Existing tests pass locally (`uv run pytest -m "cpu and not slow"`)

For full setup and testing instructions, see the [Contributing Guide](https://docs.pruna.ai/en/stable/docs_pruna/contributions/how_to_contribute.html).

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code, especially for agent-assisted changes
- [ ] I updated the documentation where necessary

---

Thanks for contributing to **Pruna**! We're excited to review your work.

New to contributing? Check out our [Contributing Guide](https://docs.pruna.ai/en/stable/docs_pruna/contributions/how_to_contribute.html) for everything you need to get started.

> **Note:** 
> - Draft PRs or PRs without a clear and detailed overview may be delayed.  
> - Please mark your PR as **Ready for Review** and ensure the sections above are filled out.
> - Contributions that are entirely AI-generated without meaningful human review are discouraged.

PR #683, opened 2026-06-15
fix: patch torchmetrics (transformers 5)

sdiazlor:
## Description

With Transformers 5, it is returned BaseModelOutputWithPooling instead of a tensor. This is a temporary patch so it extracts it if needed.

Related issue and PRs in torchmetrics: https://github.com/Lightning-AI/torchmetrics/issues/3407 and https://github.com/Lightning-AI/torchmetrics/pull/3408

```py
FAILED tests/evaluation/test_torch_metrics.py::test_clip_score[LAION256-cpu] - AttributeError: 'BaseModelOutputWithPooling' object has no attribute 'norm'
FAILED tests/evaluation/test_torch_metrics.py::test_clip_score[LAION256-cuda] - AttributeError: 'BaseModelOutputWithPooling' object has no attribute 'norm'
= 23 failed, 469 passed, 5 skipped, 245 deselected, 211 warnings in 2385.86s (0:39:45) =
sys:1: DeprecationWarning: builtin type swigvarlink has no __module__ attribute
```

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Refactor (no functional change)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Testing

- [ ] I added or updated tests covering my changes
- [ ] Existing tests pass locally (`uv run pytest -m "cpu and not slow"`)

For full setup and testing instructions, see the [Contributing Guide](https://docs.pruna.ai/en/stable/docs_pruna/contributions/how_to_contribute.html).

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code, especially for agent-assisted changes
- [ ] I updated the documentation where necessary

---

Thanks for contributing to **Pruna**! We're excited to review your work.

New to contributing? Check out our [Contributing Guide](https://docs.pruna.ai/en/stable/docs_pruna/contributions/how_to_contribute.html) for everything you need to get started.

> **Note:** 
> - Draft PRs or PRs without a clear and detailed overview may be delayed.  
> - Please mark your PR as **Ready for Review** and ensure the sections above are filled out.
> - Contributions that are entirely AI-generated without meaningful human review are discouraged.

PR #685, opened 2026-06-15
feat: compatibility with transformers 5

sdiazlor:
## Description

This PR includes updates to make pruna compatible with Transformers 5:

- unpin transformers version (not sure if you had a minimum previously) and pin in extras
- update hf command
- add the is_available_algorithm to make the unavailable under certain circumstances (e.g. transformers 5)
- small updates
- Other PRs: https://github.com/PrunaAI/pruna/pull/684 and  https://github.com/PrunaAI/pruna/pull/683

- Pinning to 5.10.2 fixes the errors regarding pipelines on the HF side (https://github.com/huggingface/transformers/pull/44621). But I don't know the preferred approach.

```py
FAILED tests/algorithms/test_algorithms.py::test_full_integration[TestIFW_whisper_tiny_random-cuda] - AttributeError: 'AutomaticSpeechRecognitionPipeline' object has no attribute 'modelcard'
```

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

---

> **Note:** 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,6 +115,7 @@ dependencies = [
     "torchvision>=0.22.0",
     "torchmetrics[image]",
     "requests>=2.31.0",
+    "transformers<6.0.0"
     "pytorch-lightning",
     "huggingface-hub[hf-xet]>=0.30.0",
     "datasets>=3.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -144,7 +144,7 @@ dependencies = [
     "gliner; python_version >= '3.11'",
     "piq",
     "opencv-python",
-    "kernels<=0.14.1", # until transformers compatibility with kernels>=0.15
+    "kernels>=0.12.0,<0.15.0",
     "aenum",
     "imageio-ffmpeg",
     "jaxtyping",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,10 +158,9 @@ dependencies = [
 [project.optional-dependencies]
 transformers-legacy = [
     "transformers<5.0.0",
-
 ]
-# ctranslate2 is isolated as an optional extra for c_translate / c_generate / c_whisper.
 
+# ctranslate2 is isolated as an optional extra for c_translate / c_generate / c_whisper.
 
 ctranslate = [
     "ctranslate2==4.6.0",
@@ -173,6 +172,7 @@ whisper = [
 vllm = [
     "vllm>=0.16.0",
     "ray",
+    "pruna[transformers-legacy]",
 ]
 llamacpp = [
     "llama-cpp-python>=0.2.78", # Required for running and inferencing Llama.cpp models
@@ -201,6 +201,7 @@ awq = [
     "compressed-tensors >= 0.13.0",
     "llmcompressor>=0.9",
     "torch>=2.9.0",
+    "pruna[transformers-legacy]",
 ]
 upscale = [
     "realesrgan",
@@ -251,7 +252,6 @@ evaluation = [
     "pruna[rapidata]",
     "pruna[lmharness]"
 ]
-
 kvpress = [
     "kvpress>=0.5.2",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,7 @@ dependencies = [
     "torchvision>=0.22.0",
     "torchmetrics[image]",
     "requests>=2.31.0",
-    "transformers<6.0.0"
+    "transformers<6.0.0",
     "pytorch-lightning",
     "huggingface-hub[hf-xet]>=0.30.0",
     "datasets>=3.0",
@@ -156,19 +156,12 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-<<<<<<< HEAD
-<<<<<<< HEAD
+transformers-legacy = [
+    "transformers<5.0.0",
+
+]
 # ctranslate2 is isolated as an optional extra for c_translate / c_generate / c_whisper.
-=======
-transformers-new = ["transformers>=5.10.2"]
-transformers-legacy = ["transformers<=4.48.0"]
 
-
-=======
->>>>>>> 9c34960 (ci: remove transformers as extras)
-# whisper-s2t and ctranslate2 are isolated because importing whisper_s2t
-# at test time causes side-effects that break unrelated tests.
->>>>>>> 89d27bc (ci: transformers dependency as extra)
 
 ctranslate = [
     "ctranslate2==4.6.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,6 @@ dependencies = [
     "torchvision>=0.22.0",
     "torchmetrics[image]",
     "requests>=2.31.0",
-    "transformers<5.0.0",
     "pytorch-lightning",
     "huggingface-hub[hf-xet]>=0.30.0",
     "datasets>=3.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,13 +78,15 @@ url = "https://prunaai.pythonanywhere.com/simple/"
 explicit = true
 
 [tool.uv]
-index-strategy = "unsafe-best-match"
+index-strategy = "first-index"
 
 conflicts = [
     [{ extra = "awq" }, { extra = "vbench" }],
     [{ extra = "vllm" }, { extra = "vbench" }],
     [{ extra = "gptq" }, { extra = "awq" }],
     [{ extra = "kvpress" }, { extra = "vbench" }],
+    [{ extra = "transformers-new" }, { extra = "transformers-legacy" }],
+    [{ extra = "transformers-pruna-current" }, { extra = "transformers-new" }]
 ]
 
 [tool.uv.sources]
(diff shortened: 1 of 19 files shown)


pyproject.toml
```diff
@@ -115,7 +115,6 @@ dependencies = [
     "torchvision>=0.22.0",
     "torchmetrics[image]",
     "requests>=2.31.0",
-    "transformers<5.0.0",
```
begumcig, line 118: I think we should not get rid of depending on transformers completely, perhaps we should extend it to something like "transformers<6.0.0" ?

src/pruna/smash.py
```diff
@@ -63,6 +64,8 @@ def smash(
         ensure_device_consistency(model, smash_config)
         # check if the algorithm packages are available for the given configuration
         check_algorithm_packages_availability(smash_config)
+        # check if the active algorithms are available in the current environment
+        check_algorithm_availability(smash_config)
```
begumcig: i belive the `check_algorithm_packages_availablity ` call above attempts to import all packages an algorithm is going to use by calling` import_algorithm_packages`. If the new version of transformer introduces something that's breaking (similar to what happened in text pipelines) I feel like we would break even before we can check for algorithm availability. How do you feel about doing the availability check before the package availability check OR in the package availability check?

src/pruna/algorithms/global_utils/recovery/perp_recoverer.py
```diff
@@ -319,3 +320,22 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
...
+        Return whether this PERP recoverer / distiller can run in the current environment.
+
+        These algorithms rely on transformers internals that changed in transformers 5.
+
+        Returns
+        -------
+        bool
+            True if the installed transformers version is below 5.0.0, False otherwise.
+        """
+        if transformers_version >= "5.0.0":
```
begumcig: Maybe we should do something like Version(5.0.0) rather than a raw string like here?

src/pruna/algorithms/token_merging.py
```diff
@@ -628,3 +629,20 @@ def import_algorithm_packages(self) -> dict[str, Any]:
...
+    def is_available_algorithm(self) -> bool:
+        """
+        Return whether Token Merging can run in the current environment.
+
+        Returns
+        -------
+        bool
+            True if the installed transformers version is below 5.0.0, False otherwise.
+        """
+        if transformers_version >= "5.0.0":
```
begumcig: Again the same comment about using a version rather than a string for the version

src/pruna/engine/model_checks.py
```diff
@@ -28,8 +29,11 @@
     AutomaticSpeechRecognitionPipeline,
 )
 from transformers.pipelines.image_classification import ImageClassificationPipeline
-from transformers.pipelines.text2text_generation import Text2TextGenerationPipeline
-from transformers.pipelines.text_generation import TextGenerationPipeline
+
+if transformers_version < "5.0.0":
```
begumcig: Again the same comment about using Version

src/pruna/algorithms/token_merging.py
```diff
@@ -628,3 +629,20 @@ def import_algorithm_packages(self) -> dict[str, Any]:
         from transformers.models.vit.modeling_vit import ViTLayer, ViTSelfAttention
 
         return dict(ViTLayer=ViTLayer, ViTSelfAttention=ViTSelfAttention)
+
+    def is_available_algorithm(self) -> bool:
```
begumcig: I really like this `is_available_algorithm` function! To also make pruna installing easier, how do you feel about adding a "transformer_older" like extra in the `pyproject.toml` then separate the algorithms that need the older versions of transformers?

pyproject.toml
```diff
@@ -78,7 +78,7 @@ url = "https://prunaai.pythonanywhere.com/simple/"
 explicit = true
```
begumcig, line 78: Just a heads up I think we should add the conflicts for the new version of the transformers with the extras that clash with them (awq and vllm on top of my head) under the conflicts, otherwise you may have problems with installing pruna

begumcig (changes requested):
Thanks a lot for tackling this Sara! I really like the` is_available `check, i think we should introduce some optional dependencies and conflicts to make installing pruna easier, but looks super good already!!!!

src/pruna/config/pre_smash_routines.py
```diff
@@ -115,6 +115,21 @@ def check_algorithm_packages_availability(smash_config: SmashConfig) -> None:
...
+
+    Parameters
+    ----------
+    smash_config : SmashConfig
+        The SmashConfig object containing the algorithm configuration.
+    """
+    for algorithm_name in smash_config.get_active_algorithms():
+        algorithm = AlgorithmRegistry[algorithm_name]
+        if not algorithm.is_available_algorithm():
+            raise ValueError(f"Algorithm '{algorithm_name}' is not available in the current environment.")
```
gsprochette: This is not super informative on how to fix it I feel, shouldn't the is_available_algorithm return both a boolean and a message stating why it's not available? This way we can provide the user with some help about how to make it available.

gsprochette: After further reading, I now see that you log the errors in the pruna logger, so I completely agree with your implementation and I would say we can just prompt the user to check the logs for further details on why it's not available?

src/pruna/algorithms/base/pruna_base.py
```diff
@@ -256,6 +256,17 @@ def import_algorithm_packages(self) -> Dict[str, Any]:
         """
         return dict()
 
+    def is_available_algorithm(self) -> bool:
+        """
+        Return whether this algorithm can run in the current environment.
```
gsprochette: For the contributors, I think it would be good to write somewhere in the docstring that this method should log an error about why it's not available, when it's not.

src/pruna/algorithms/token_merging.py
gsprochette: i'm not sure what you mean, if it's just a single dependency like transformers, the user can install pruna with `uv pip install pruna "transformers<5"` and it should be fine, no?
Or do you mean this for managing the tests on both transformers<5 and transformers>=5? If so we intend to make everything compatible with transformers>=5 in the long run so I think this old_transformers extra is a bit overkill. WDYT?

src/pruna/engine/model_checks.py
```diff
@@ -28,8 +29,11 @@
...
+    from transformers.pipelines.text2text_generation import Text2TextGenerationPipeline as TextGenerationPipeline
+else:
+    from transformers.pipelines.text_generation import TextGenerationPipeline
```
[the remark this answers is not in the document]
gsprochette: This looks important indeed

pyproject.toml
gsprochette: ohhh if that's the reason for adding a old_transformers extra then yes we need to talk about this. Are there many extras that aren't compatible with transformers 5? Does it result in the algorithms being unavailable in a transformers 5+ env? Because there isn't many algorithms that are marked unavailable.

gsprochette (commented):
Already super clean and efficient code, great job this will get merged in no time 🫶  I think the unavailability error could be slightly more explicit, but the only big thing is we need to discuss the idea of having a `old_transformers` or `legacy_transformers` extra and what that would bring. At the moment I'm not super sure what is the justification for this is and where it is used 🙃

pyproject.toml
```diff
@@ -115,7 +115,7 @@ dependencies = [
     "torchvision>=0.22.0",
     "torchmetrics[image]",
     "requests>=2.31.0",
-    "transformers<5.0.0",
+    "transformers<6.0.0",
```
begumcig, line 118: LET'S GOOOOOOOOO 💅💅💅💅💅💅💅💅

begumcig (approved):
Looks amazing thank you so much Sara!!! I added one comment about the uv indexing strategy but looks super good!!! Let's get this to merging

pyproject.toml
```diff
@@ -78,7 +78,7 @@ url = "https://prunaai.pythonanywhere.com/simple/"
 explicit = true
 
 [tool.uv]
-index-strategy = "unsafe-best-match"
+index-strategy = "first-index"
```
gsprochette: this should be avoided at all cost, it creates problem when trying to install pruna and torch together on the same line with the torch extra index url because old versions of some packages are found on their index and therefore uv doesn't look for them on pypi anymore. Can you revert back to `unsafe-best-match` please or did it create problems?

tests/config/test_pre_smash_routines.py
```diff
@@ -21,6 +21,9 @@
 import networkx as nx
 
 from pruna import SmashConfig
+from pruna.algorithms.distillation_perp import TextToImagePERPDistillation
+from pruna.algorithms.perp import TextToImagePERP
+from pruna.algorithms.token_merging import TokenMerging
```
gsprochette: These imports are not used anymore I think? Leaving only formatting in this file so I would be in favor of not touching this file at all and restoring it to its state in main

gsprochette (changes requested):
Looks great ! One minor thing about a file that's only formatted, and the very important index thing :)

pyproject.toml
```diff
@@ -191,14 +196,13 @@ gptq = [
...
+    "torch>=2.9.0",
+    "pruna[transformers-legacy]",
 ]
 upscale = [
     "realesrgan",
 ]
-full = [
-    "pruna[stable-fast]",
-]
+full = []
```
gsprochette: full doesn't exist let's remove it

gsprochette (approved):
Let's remove full and let's gooooooooo

PR #687, opened 2026-06-15
test: fix breaking nightly tests for ONEIG dataset

begumcig:
## Description

Nine of our nightly tests were failing, all related to OneIG dataset checks.

After investigation, we found that eight of these were not actual failures. The tests were asserting that each category contained four samples, but some of those categories have fewer than four samples in the original dataset. We verified the OneIG dataset directly to confirm that these categories genuinely contain fewer than four samples and that the issue was not caused by a loading error on our side.

The remaining failure was caused by our loaders being unable to find the `Multilingualism` category. This is a real issue. Currently, `multilingualism` is not loaded through `OneIGBench`, but through `OneIGBenchZH`, which also includes Chinese prompts for all other main OneIG categories. We do not currently support this setup.

To enable benchmarking with Chinese prompts, we would need to integrate not only `multilingualism`, but also the other Chinese categories, which would require additional work. For now, I removed `multilingualism` from the list of available categories.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

---

> **Note:** 

### DIFF
--- a/src/pruna/data/datasets/prompt.py
+++ b/src/pruna/data/datasets/prompt.py
@@ -25,7 +25,6 @@ OneIGCategory = Literal[
     "Anime_Stylization",
     "General_Object",
     "Knowledge_Reasoning",
-    "Multilingualism",
     "Portrait",
     "Text_Rendering",
     "3d rendering",
--- a/tests/data/test_datamodule.py
+++ b/tests/data/test_datamodule.py
@@ -121,13 +121,13 @@ def _benchmarks_with_category() -> list[tuple[str, str]]:
 @pytest.mark.parametrize("dataset_name, category", _benchmarks_with_category())
 def test_benchmark_category_filter(dataset_name: str, category: str) -> None:
     """Test dataset loading with each category filter; dataset has at least one sample."""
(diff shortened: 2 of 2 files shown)


gsprochette (approved):
Makes super sense! Thanks a lot for the investigation and explanation :)

Is there anywhere where we should write about re-integrating multilingualism in the future? Was there a specific use-case for this category or was it just a mixup between the two sources?

begumcig:
> Makes super sense! Thanks a lot for the investigation and explanation :)
> 
> Is there anywhere where we should write about re-integrating multilingualism in the future? Was there a specific use-case for this category or was it just a mixup between the two sources?

@gsprochette We could not use multilingualism with the current implementation because the category doesn't exist in our current source. We also currently do not have a use-case for it in the benchmarking pipeline. I think there was a PR in the pipeline that worked on it that got closed (https://github.com/PrunaAI/pruna/pull/545/changes#diff-8d3c63a2fe8b8f64fda6543afe36c1e31b6f279e2ae86739ed536b1884067b55R770) so we might go back to it in the future.

PR #701, opened 2026-06-25
fix: make CallSequenceTracker unwrap idempotent for bound methods

FAdrien:
## Description

Fixes a `TypeError` in `CallSequenceTracker` that breaks `total_macs` and `total_params` metrics when the same module’s `forward` is unwrapped more than once.

### Context

`ModelArchitectureStats` wraps all `nn.Module` forwards before a single inference run, records the call sequence, then calls `unwrap()` on each module while computing MACs/params. If a module appears multiple times in the call sequence, `unwrap()` runs on the same module more than once.

### Cause

`unwrap()` previously restored the module's `forward` by checking if the it has the `__wrapped__` attribute. The first unwrap correctly restores the bound method. As this bound method can still expose the `__wrapped__` attribute, a second unwrap then replaces it with the unbound function. Next calls to the model (e.g., to generate an image) then fail with a `TypeError` because `self` is no longer bound.

### Fix

This PR stores the original forward on a dedicated `__call_tracker_original_forward__` attribute, skips re-wrapping modules that are already tracked, and makes `unwrap()` idempotent on multiple calls.

## Related Issue

Fixes #(issue number)

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

---

> **Note:** 

### DIFF
--- a/src/pruna/engine/call_sequence_tracker.py
+++ b/src/pruna/engine/call_sequence_tracker.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from functools import wraps
-from typing import Any, Dict, List, cast
+from typing import Any, Dict, List
 
 import torch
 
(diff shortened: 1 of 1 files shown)


begumcig (approved):
looks super good to me! We can merge once the tests pass :) Thank youuuuuuuu

PR #706, opened 2026-07-06
fix: parameters order assumption in diffuser

kfachikov:
## Description
This change ensures that the inference handler is compatible with diffuser pipelines that do not have "prompt" as a first parameter, such as [Flux2KleinPipeline](https://huggingface.co/docs/diffusers/api/pipelines/flux2#diffusers.Flux2KleinPipeline.__call__).

#### Extra
- Updated the documentation for `DiffuserHandler`, specifying the expected behavior for `prepare_inputs`.
- Updated the documentation for the `collate.py` module, elaborating on the return types for each collate method.

## Related Issue
Many diffusers pipelines, such as [QwenImagePipeline](https://huggingface.co/docs/diffusers/api/pipelines/qwenimage#diffusers.QwenImagePipeline.__call__), have "prompt" as a first parameter. The existing code relied on this assumption - as we can see [here](https://github.com/PrunaAI/pruna/blob/04e1edf2a47dd4777adfc96819b3c2e401efc9ae/src/pruna/engine/handler/handler_diffuser.py#L32),

> The first element of the batch is passed as input to the model.

However, not all diffusers pipelines have the same signature. [Flux2KleinPipeline](https://huggingface.co/docs/diffusers/api/pipelines/flux2#diffusers.Flux2KleinPipeline.__call__) for example expects model on first position and prompt on second.

Running a [pruna evaluation](https://docs.pruna.ai/en/stable/reference/evaluation.html) with the Flux2 model resulted in an error

```
ValueError: Provide either `prompt` or `prompt_embeds`. Cannot leave both `prompt` and `prompt_embeds` undefined.
```
## Important

This change is backward compatible and does not break any existing working scenarios. It expands the usecases to diffusion models which do not expect `prompt` on first position (e.g., [Flux2KleinPipeline](https://huggingface.co/docs/diffusers/api/pipelines/flux2#diffusers.Flux2KleinPipeline.__call__)), but does NOT resolve the following **hypothetical** problems/limitations.

- We have a dataset and its first batch element is a dict that is meant to be passed as a single value to one argument (rather than spread).
- Any diffusion model wrapped by PrunaModel and run with `run_inference` would expect the model inputs to be the first element in the batch passed as argument.

Neither the old implementation nor the new code handle these.

## How?

The change in the metrics follows the already established patter to unwrap the _prepared inputs_ - the way it's done in [PrunaModel.run_inference](https://github.com/PrunaAI/pruna/blob/04e1edf2a47dd4777adfc96819b3c2e401efc9ae/src/pruna/engine/pruna_model.py#L111-L116) used in [evaluation_agent](https://github.com/PrunaAI/pruna/blob/04e1edf2a47dd4777adfc96819b3c2e401efc9ae/src/pruna/evaluation/evaluation_agent.py#L269), the [elapsed time metric](https://github.com/PrunaAI/pruna/blob/04e1edf2a47dd4777adfc96819b3c2e401efc9ae/src/pruna/evaluation/metrics/metric_elapsed_time.py#L151,L157), and handlethe [memory metric](https://github.com/PrunaAI/pruna/blob/04e1edf2a47dd4777adfc96819b3c2e401efc9ae/src/pruna/evaluation/metrics/metric_memory.py#L361).

## Type of Change

- [x] Documentation update

## Testing

  - I added comprehensive tests for the `handler_diffusers` module. They exercise the `prepare_inputs` behavior under
    -  pruna native datasets (i.e., results from collate - `image_generation_collate`, `prompt_collate`, and `prompt_with_auxiliaries_collate`),
    - and custom datasets (i.e., including image for image-edit, or negative prompt for conditional generation)

## Checklist

## TL;DR

I explored all the pruna native datasets (resolvable through `PrunaDataModule.fromString`) and the way they are processed (i.e., the collate functions). Here is a list with their groups by collate:

1. image_generation_collate
2. prompt_collate
3. prompt_with_auxiliaries_collate
4. audio_collate
5. image_classification_collate
6. text_generation_collate
7. question_answering_collate

Considering our scope - diffusion models, only Image Generation tasks are relevant. This limits the datasets to 1, 2, and 3. The audio collate is not appropriate for audio generation as it provides only the original audio and its transcription, without any description of what's in the audio.

Furthermore, diffusion models can also be used for the other tasks (image classification, text generation, or Q&A), even though they are not really popular and require careful manipulation. The structure provided by 5, 6, and 7 does not match the desired _custom_ structure requested by these models and thus we do not support them.

---

> **Note:** 


kfachikov:
@begumcig What do you think about reusing `run_inference` in the metric_energy and the metric_model_architecture`?

I see that you have made such a change in some other metrics a year ago (e.g., [here](https://github.com/PrunaAI/pruna/commit/ed4a7386001560465e3c828953a181ae970e1eef#diff-7d5dcf8477e2c851534d6557ac46b2a638ae0f00bdcff20787dfc8d633f5fdae)) and wondered whether there was a particular reason why these two were not changed.

One small concern with metric_energy is that the current code avoids constantly _moving the batch to the "device" and preparing the inputs_ (as it does it only once - code [here](https://github.com/PrunaAI/pruna/blob/main/src/pruna/evaluation/metrics/metric_energy.py#L102-L113). In case we use `run_inference`, this will happen on every iteration, possibly skewing the measurements.

AFAIU, the metric_model_architecture would not be affected.

kfachikov:
@begumcig 

An update on using `run_inference` in the two metrics - I still have to run experiments and evaluate whether a (or any) difference is observed. As this change is just a QoL, I will leave it out of this PR.

_edit_

Posting experiment numbers here for history.

image

tests/fixtures.py
```diff
@@ -206,6 +206,7 @@ def get_autoregressive_text_to_image_model(model_id: str) -> tuple[Any, SmashCon
     "flux_tiny_random_with_tokenizer": partial(
         get_diffusers_model_with_tokenizer, "katuni4ka/tiny-random-flux", torch_dtype=torch.float16
     ),
+    "flux2_tiny_random": partial(get_diffusers_model, "tiny-random/flux2", torch_dtype=torch.bfloat16),
```
gsprochette, line 209: does flux2 require bumping the minimum versions of transformers or diffusers in the dev extra?

gsprochette (approved):
Looks good, @begumcig knows more than me about the inference handler so maybe she'll find missing edge cases but I didn't see one :) Thanks a lot !

tests/fixtures.py
kfachikov: That's a good point. My virtual environment seems to have versions that are high enough, but they are not enforced. I updated the diffusers version to min 0.37, which is when Flux2KleinPipeline was launched ([commit in diffusers](https://github.com/huggingface/diffusers/commit/61f175660a8ac54f1470a74a810e6c38fb4795d5)).

pyproject.toml
```diff
@@ -121,7 +121,7 @@ dependencies = [
     "datasets>=3.0",
     "numpy>=1.24.4",
     "numpydoc>=1.6.0",
-    "diffusers>=0.21.4",
+    "diffusers>=0.37.0",
```
gsprochette: I don't think we need to change the main dependency, only in the dev extra since we're using it for tests. If the only reason for this is the flux2 fixture, that's only reachable in tests not in src/pruna so we can leave the diffusers dependency here pretty loose

src/pruna/engine/handler/handler_diffuser.py
```diff
@@ -48,22 +48,31 @@ def __init__(self, call_signature: inspect.Signature, model_args: Optional[Dict[
...
+        batch : Tuple[List[str] | torch.Tensor | dict[str, Any], ...]
             The batch to prepare the inputs for.
 
         Returns
         -------
         Any
             The prepared inputs.
         """
-        if "prompt" in self.call_signature.parameters or "args" in self.call_signature.parameters:
+        if "prompt" in self.call_signature.parameters:
```
begumcig, line 72: I remember we had a discussion about this, do we know for sure that the first element of the batch will always be the prompt for the data that will be directed to the diffusers handler?

begumcig (approved):
Looks super good to me! Just asked one question to make sure we are not labeling things that are not prompts as prompts, but feel free to merge if that's not an issue!!!!! Thank you so much Kaloyan! You dropped this 👑

src/pruna/engine/handler/handler_diffuser.py
kfachikov: It's a good and important question. In **theory**, yes. Whenever users use the `PrunaDataModule` and the collate pruna functions, the prompt is the first element in a batch. We also have a comment in the class description for DiffuserHandler saying "The first element of the batch is passed as input to the model.", creating a clear expectation towards the behavior. To make it even more explicit, I added an elaborate comment in the method too. However, in **practice**, we cannot prevent the user using a custom collate functions and a custom dataset. This is actually intentionally allowed by the framework, allowing users to use pruna in more scenarios. Yet, as we quickly mentioned it, we cannot handle every single usecase and guard every single user (as there would also be people who are willingly breaking things...)

PR #723, opened 2026-07-30
fix: validate explicit algorithm order

ShiroKSH:
## Description

Fix explicit algorithm-order validation so that only a dependency placed in reverse order is rejected. Independent algorithms no longer make a valid custom order fail.

## Related Issue

N/A

## Type of Change

## Testing

- Added a regression test covering both independent algorithms and a reversed dependency.
- Ran Python compilation for the changed source and test files.
- Ran a targeted behavioural smoke test against the changed validation function.

## Checklist

- [x] My code follows the style guidelines of this project.
- [x] I have performed a self-review of my code, especially for agent-assisted changes.

## AI-assisted

This is an AI-assisted contribution. The change was independently reviewed against the order-graph semantics and covered with a focused regression test.


ShiroKSH:
> This PR has been inactive for 10 days and is now marked as stale. It will be closed in 7 days if there is no further activity.

123

ShiroKSH:
> [!NOTE]
>

PR #733, opened 2026-08-27
feat: add static fp8 quantization algorithm

kfachikov:
## Description
This algorithm implements symmetric linear quantization using the `amax` value.

- The weights are quantized once and the full-precision weights are discarder.
- The activations are quantized on-the-go, using a fixed scale. The scale is computed during calibration - complete noise-to-image generations during "smash".

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

### Setup

There are multiple other algorithms in the comparison, used for reference.

- `f8_dummy` - The practical speed limit when it comes to static quantization. The method skips calibration and uses a fixed (random) number as scales. _Thus, it does not need any boolean flags indicating the "mode" we run in, nor any conditional statements._
- `fp8_dynamic` - The theoretical best when it comes to quality. It does not precompute scales, but instead each layer computes them on-the-go, as the input arrives. This is what `torchao` provides under [Float8DynamicActivationFloat8WeightConfig](https://docs.pytorch.org/ao/main/api_reference/generated/torchao.quantization.Float8DynamicActivationFloat8WeightConfig.html#torchao.quantization.Float8DynamicActivationFloat8WeightConfig).

Our goal is to implement a static quantization algorithm, s.t. it _matches_ the dummy's speed and the dynamic's quality.

### Results

The experiments are conducted using Qwen-Image-2512. Pairwise scores are computed w.r.t. a bf16 variant. All models are compressed using `smash`. The `_compiled` suffix indicates `torch.compile` applied with `target: module_list` and `fullgraph: false`.

| Quantization Algorithm | CLIP Score ⬆️ | Inference (ms) | Speedup | Disk Memory | Inference Memory | Memory Cmp. |
| --- | --- | --- | --- | --- | --- | --- |
| fp8_dummy_compiled | N/A | 8997.427506 | 1.75x | 42490 | 47118 | 0.77x |
| static_fp8_diffusers_compiled | 23.551832 | 9171.238864 | 1.72x | 42518 | 47148 | 0.77x |
| fp8_dynamic_compiled | 23.835505 | 9875.194669 | 1.60x | 42430 | 47170 | 0.77x |
| static_fp8_diffusers | 23.833471 | 15453.424490 | 1.02x | 42554 | 47580 | 0.77x |
| baseline | 23.956564 | 15760.324677 | 1.00x | 55302 | 60208 | 1.00x |

OpenImage, 100 prompts

| Quantization Algorithm | Pairwise CLIP Score ⬆️ | CMMD ⬇️ | LPIPS ⬇️ | MSSSIM ⬆️ | PSNR ⬆️ |
| --- | --- | --- | --- | --- | --- |
| fp8_dynamic_compiled | 98.247383 | 0.007629 | 0.071664 | 0.948175 | 26.475830 |
| static_fp8_diffusers_compiled | 98.211365 | 0.009179 | 0.076860 | 0.944918 | 26.176754 |

LAION256, 46 prompts

| Quantization Algorithm | Pairwise CLIP Score ⬆️ | CMMD ⬇️ | LPIPS ⬇️ | MSSSIM ⬆️ | PSNR ⬆️ |
| --- | --- | --- | --- | --- | --- |
| fp8_dynamic_compiled | 96.642136 | 0.019193 | 0.075059 | 0.950166 | 26.583673 |
| static_fp8_diffusers_compiled | 96.565979 | 0.021219 | 0.084726 | 0.939995 | 25.400696 |

#### How to store the scales?

The AlgorithBase has a property called `dataset_required` which feels really appropriate for this algorithm. However, it comes with a caveat.

1. On [smash](https://github.com/PrunaAI/pruna/blob/main/src/pruna/smash.py#L73), the smash config is checked. In particular, whenever `dataset_calibrated` is set to True, the data property is also required (call in [smash](https://github.com/PrunaAI/pruna/blob/main/src/pruna/smash.py#L73) → [check](https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/pre_smash_routines.py#L138-L139))
2. On model save, the smash configuration is saved as JSON, but the data property is intentionally omitted (call in [save_pruna_model](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/save.py#L100) → [code](https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/smash_config.py#L338-L339)).
3. On model load, the smash functions is invoked again when there is an algorithm with save_before_apply (call in [load_pruna_model](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/load.py#L85-L86) → (default) [resmash](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/load.py#L226))

Now consider that we have an algorithm with `dataset_required=True`. We smash it, save it, load it, and … crash. The resmashing checks whether there is any data and finds none. For this reason, I decided to use the data property (allowing users to pass any dataset for calibration - as a string among the pruna native datasets, as a PrunaDataModule, etc; [code](https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/smash_config.py#L431-L478)) but manually check whether the data is present when the model is being smashed for the first time (i.e., not loaded from previously saved one). This is a similar approach as the one adopted in [Quanto](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/quanto.py), even though calibration is not required there.

#### What do we store?

- the input_running_amax, allowing us to change the input quantization upon load (e4m3 to e5m2 or vice versa) without the need to run calibration again.

#### Why custom `get_model_dependent_hyperparameter_defaults`?
Why did I choose to create a custom method but not follow the example in torchao (skipped modules [here](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/torchao.py#L39-L73) and function [here](https://github.com/PrunaAI/pruna/blob/main/src/pruna/algorithms/torchao.py#L185-L216)?

There are numerous [diffusers](https://huggingface.co/docs/diffusers/index) models and each and every one might name its modules in a different manner. Trying to list all _common_ embedding and norm modules might only lead to an unexpected behavior - when some modules are quantized and others are not, using a _heuristic_ unknown to the user. The simple solution I adopt clearly states that, by default, it would skip any layers which include "embed" or "norm" in their naming.

---

> **Note:** 

### DIFF
--- a/src/pruna/algorithms/global_utils/quantization/symmetric_scale.py
+++ b/src/pruna/algorithms/global_utils/quantization/symmetric_scale.py
@@ -1,0 +1,57 @@
+# Copyright 2025 - Pruna AI GmbH. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import torch
+
+
+def amax_to_scale(amax: torch.Tensor, max_val: float) -> torch.Tensor:
+    """
+    Convert an absolute-max value to a per-tensor quantization scale.
+
+    Parameters
+    ----------
+    amax : torch.Tensor
+        The absolute max value of the tensor to quantize.
+    max_val : float
+        The maximum representable magnitude of the target float dtype.
+
+    Returns
+    -------
+    torch.Tensor
+        The scale to use for quantization.
+    """
+    return (max_val / torch.clamp(amax, min=1e-12)).clamp(max=max_val)
+
+
+def scale_and_clamp(x: torch.Tensor, scale: torch.Tensor, max_val: float) -> torch.Tensor:
+    """
+    Map the input to the range [-max_val, max_val] using the given scale.
+
+    Parameters
+    ----------
+    x : torch.Tensor
+        The input to quantize.
+    scale : torch.Tensor
+        The scale to use for quantization.
+    max_val : float
+        The maximum representable magnitude of the target float dtype.
+
+    Returns
+    -------
+    torch.Tensor
+        The scaled input, clamped to ``[-max_val, max_val]``.
+    """
+    return (x * scale).clamp(-max_val, max_val)
--- a/src/pruna/algorithms/global_utils/quantization/symmetric_scale.py
+++ b/src/pruna/algorithms/global_utils/quantization/symmetric_scale.py
@@ -33,7 +33,7 @@ def amax_to_scale(amax: torch.Tensor, max_val: float) -> torch.Tensor:
     torch.Tensor
         The scale to use for quantization.
     """
-    return (max_val / torch.clamp(amax, min=1e-12)).clamp(max=max_val)
+    return (max_val / torch.clamp(amax, min=1e-12))
 
 
 def scale_and_clamp(x: torch.Tensor, scale: torch.Tensor, max_val: float) -> torch.Tensor:
--- a/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
+++ b/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
@@ -203,7 +203,7 @@ class StaticFp8Diffusers(PrunaAlgorithmBase):
         if target_modules is None:
             target_modules = self.get_model_dependent_hyperparameter_defaults(model, smash_config)["target_modules"]
             target_modules = cast(TARGET_MODULES_TYPE, target_modules)
-        
+
         target_linear_modules = filter_targeted_modules(
             keep_targeted_fn=lambda module, path: isinstance(module, torch.nn.Linear),
             model=model,
--- a/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
+++ b/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
@@ -150,8 +150,10 @@ class StaticFp8Diffusers(PrunaAlgorithmBase):
         """
         Provide default `target_modules` using `target_backbone`, excluding sensitive modules.
 
-        Extends the base backbone targets by excluding embedding, norm, lm_head, and proj_out
-        modules which are sensitive to FP8 quantization.
+        Extends the base backbone targets by excluding layers which are sensitive to FP8 quantization,
+        including embedding, norm, lm_head, and proj_out layers. The matching pattern excludes
+        only layers which match the patterns "*embed*", "*norm*", "*lm_head", and "proj_out".
+        For a more precise control, the user should specify a custom matching pattern in the smash config.
 
         Parameters
         ----------
@@ -166,6 +168,7 @@ class StaticFp8Diffusers(PrunaAlgorithmBase):
             A dictionary with a "target_modules" key defining which modules should be targeted by default.
         """
         target_modules = target_backbone(model)
+
         proj_out_patterns = ["unet.proj_out", "transformer.proj_out", "proj_out"]
         extra_exclude = ["*embed*", "*norm*", "*lm_head"] + proj_out_patterns
         target_modules["exclude"].extend(extra_exclude)
@@ -200,6 +203,7 @@ class StaticFp8Diffusers(PrunaAlgorithmBase):
         if target_modules is None:
             target_modules = self.get_model_dependent_hyperparameter_defaults(model, smash_config)["target_modules"]
             target_modules = cast(TARGET_MODULES_TYPE, target_modules)
+        
         target_linear_modules = filter_targeted_modules(
             keep_targeted_fn=lambda module, path: isinstance(module, torch.nn.Linear),
             model=model,
@@ -235,6 +239,7 @@ class StaticFp8Diffusers(PrunaAlgorithmBase):
                     },
                 )
 
+            # Collect the quantized layers in each submodule.
             for submodule in module.modules():
                 if isinstance(submodule, StaticFp8Linear):
                     quantized_layers[id(submodule)] = submodule
--- a/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
+++ b/src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
@@ -18,11 +18,7 @@ from collections.abc import Iterable
 from typing import Any, cast
 
 import torch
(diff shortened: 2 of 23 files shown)


src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
```diff
@@ -0,0 +1,314 @@
...
+    def get_model_dependent_hyperparameter_defaults(
+        self, model: Any, smash_config: SmashConfigPrefixWrapper
+    ) -> dict[str, Any]:
+        """
+        Provide default `target_modules` using `target_backbone`, excluding sensitive modules.
+
+        Extends the base backbone targets by excluding layers which are sensitive to FP8 quantization,
+        including embedding, norm, lm_head, and proj_out layers. The matching pattern excludes
+        only layers which match the patterns "*embed*", "*norm*", "*lm_head", and "proj_out".
+        For a more precise control, the user should specify a custom matching pattern in the smash config.
```
kfachikov, line 159: Should this behavior be documented in the algorithm (and thus appear in the pruna API docs) rather than being commented only in the code?

src/pruna/algorithms/global_utils/quantization/symmetric_scale.py
```diff
@@ -0,0 +1,57 @@
...
+        The absolute max value of the tensor to quantize.
+    max_val : float
+        The maximum representable magnitude of the target float dtype.
+
+    Returns
+    -------
+    torch.Tensor
+        The scale to use for quantization.
+    """
+    return (max_val / torch.clamp(amax, min=1e-12)).clamp(max=max_val)
```
kfachikov: The second clamp (i.e., restricting the scale) has a particular reason.

- It is only triggered when the amax is smaller than 1. The tensor would not be stretched to the full fp8 range. As floating-points numbers have fixed **relative** error, the rounding errors are not changed. The only drawback is that very small numbers will fall into the closest-to-zero bins - the zero, the subnormal numbers, and the smallest normal numbers, which form 17 equi-width bins, thus incurring fixed **absolute** error. 
- By keeping the scale small, we ensure that matrix multiplication would not blow up when fast accumulation is used - a torch.compile feature which uses lower-precision for intermediate results, speeding up computations.

src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
```diff
@@ -0,0 +1,316 @@
...
+        "Q-Diffusion": "https://arxiv.org/abs/2302.04304",
+        "Post-training Quantization on Diffusion Models": "https://arxiv.org/abs/2211.15736"
+    }
+    save_fn = SAVE_FUNCTIONS.save_before_apply
+    tokenizer_required = False
+    processor_required = False
+    runs_on: list[str] = ["cpu", "cuda", "accelerate"]
+    # dataset_required=True is not compatible with the save_before_apply hook,
+    # as the dataset is not stored on save, and thus not available on load.
+    dataset_required = False
```
begumcig, line 68: I completely understand the purpose of this, but I still find it a bit misleading. Because in the case of smashing, we need a dataset, and will fail in the _apply method. Totally understand if this is out of scope for this PR, and this is mostly a note to myself, but maybe we do need another flag in the future, something for calibration for example that checks the smash case, but not the load case

src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
begumcig: Really good question, I don't think we have the explanation for this in the docs for any other algorithm that implements this function, and it's a gap. I think for now, it's okay to keep it this way though!

src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
```diff
@@ -0,0 +1,316 @@
...
+        smash_config : SmashConfigPrefixWrapper
+            The configuration for the quantization.
+
+        Returns
+        -------
+        Any
+            The quantized, calibrated model.
+        """
+        weight_float8_dtype = (
+            torch.float8_e4m3fn if smash_config["weight_float8_dtype"] == "torch.float8_e4m3fn" else torch.float8_e5m2
```
begumcig: I understand that the above categorical choices ensures that we never get anything that's not these values. I would still prefer if we also explicitly check for the `float8_e5m2` case. Because if we add a new option in the hyperparameters, but forget to change this, we would still incorrectly map anything else to `float8_e5m2`. So a little nitpicky future proofing :D

begumcig (changes requested):
Insanely good Kaloyan, thank you so much for this!! Just one comment requesting a tiny tiny change, and nothing else. Already ready to merge!

src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
kfachikov: Hmm, I do understand you. Considering that it needs documentation in code, it is immediately clear that it's a way to _bypass_ the traditional logic rather than an intended usage. Thus, I do agree that it is not sustainable over time.

Here is the invocation chain.

- [this](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/load.py#L51) is the method which _loads_ a model
- [this](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/load.py#L226) is the default method used for "re-smashing"
- [here](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/load.py#L253) it invokes the same smash the user can use
- [here](https://github.com/PrunaAI/pruna/blob/main/src/pruna/smash.py#L73) the smash method invokes [the function](https://github.com/PrunaAI/pruna/blob/main/src/pruna/config/pre_smash_routines.py#L118) which check that validates a dataset exists when `dataset_required` is truthy

Considering this, I believe a long-term solution could be to add the following.
- _as you recommended_, a flag indicating that a dataset is specifically for calibration, and
- a flag indicating that we are in "loading" phase, necessary to ensure runtime can make a difference between _first time smash_ and _post-save load_.

There are multiple caveats.

- We must ensure that algorithms relying on calibration (i.e., a calibration dataset) will save an artifact that is enough to restore the _calibrated_ state on its own.
- I am not entirely sure what other use cases might a smash config dataset have. What I know is that it is not and cannot be stored on `save`. Thus, any "reapply" algorithm (i.e., with [`save_before_apply`](https://github.com/PrunaAI/pruna/blob/main/src/pruna/engine/save.py#L558) save function) that uses such a dataset should NOT be applied to a model that's intended to be saved and loaded, as a crash would appear during the load, when the dataset check kicks-in.

I would create a ticket and try to explain (1) what is the current problem, (2) what does not allow for a straightforward solution, and (3) what should be kept in mind.

begumcig: Yes this sounds awesome!

begumcig (approved):
Amazing, thank youuuuu! Let's merge this

src/pruna/algorithms/static_fp8_diffusers/static_fp8_diffusers.py
kfachikov: https://github.com/PrunaAI/pruna/issues/736

PR #737, opened 2026-09-02
feat: add time-aware fp8 quantization for diffuser

kfachikov:
## Description
This algorithm implements symmetric linear quantization using an amax value computed per time-step.

- The weights are quantized once and the full-precision weights are discarder.
- Similarly to the StaticFp8Diffusers algorithm (added in https://github.com/PrunaAI/pruna/pull/733), it calibrates over complete noise-to-image generations. However, rather than accumulating a single statistic across the entire denoising process, it records the relevant statistic(s) for each layer per time-step. Thus, the scales are per time-step, thus accounting for the variance in activation distributions across time.

## Type of Change

- [ ] Documentation update

## Testing

## Checklist

### Setup

There are multiple other algorithms in the comparison, used for reference.

- `f8_dummy` - The practical speed limit when it comes to static quantization. The method skips calibration and uses a fixed (random) number as scales. _Thus, it does not need any boolean flags indicating the "mode" we run in, nor any conditional statements._
- `fp8_dynamic` - The theoretical best when it comes to quality. It does not precompute scales, but instead each layer computes them on-the-go, as the input arrives. This is what `torchao` provides under [Float8DynamicActivationFloat8WeightConfig](https://docs.pytorch.org/ao/main/api_reference/generated/torchao.quantization.Float8DynamicActivationFloat8WeightConfig.html#torchao.quantization.Float8DynamicActivationFloat8WeightConfig).
- `static_fp8_diffusers` - Static quantization. The current best in (1) reaching the practical limit on speedup and (2) approximating the dynamic method in quality.

Our goal is to implement a time-aware quantization algorithm, s.t. it keeps the speed achieved by the static algorithm, but improves its quality.

### Results

The experiments are conducted using Qwen-Image-2512. Pairwise scores are computed w.r.t. a bf16 variant. All models are compressed using `smash`. The `_compiled` suffix indicates `torch.compile` applied with `target: module_list` and `fullgraph: false`.

<p>On pairwise metrics, the time-aware algorithm performs slightly better than the dynamic quantization (which is honestly surprising). On unary quality metrics (CLIP score), it performs in-between dynamic and static quantization. On speed, it almost matches the practical limit, giving approximately 7.5% speed increase.</p>
<p>LAION256 image dataset with 46 prompts in the test dataset (used by EvaluationAgent)<br>
100 generations for latency/throughput, 10 warmup; <em>The entries are sorted based on speed.</em></p>

Quantization Algorithm | CLIP Score ⬆️ | Inference (ms) | Speedup | Disk Memory | Inference Memory | Memory Cmp.
| --- | --- | --- | --- | --- | --- | --- |
| fp8_dummy_compiled | N/A | 8997.427506 | 1.75x | 42490.000000 | 47118.000000 | 0.77x |
| time_aware_fp8_diffusers_compiled | 23.705751 | 9137.848845 | 1.72x | 42510 | 48122 | 0.77x |
| static_fp8_diffusers_compiled | 23.551832 | 9171.238864 | 1.72x | 42518.000000 | 47148.000000 | 0.77x |
| fp8_dynamic_compiled | 23.835505 | 9875.194669 | 1.60x | 42430.000000 | 42430.000000 | 0.77x |
| time_aware_fp8_diffusers | 23.700428 | 15397.741959 | 1.02x | 42510 | 47144 | 0.77x |
| static_fp8_diffusers | 23.833471 | 15453.424490 | 1.02x | 42554.000000 | 47580.000000 | 0.77x |
| baseline | 23.956564 | 15760.324677 | 1.00x | 55302.000000 | 60208.000000 | 1.00x |

<p>OpenImage, 100 prompts</p>

Quantization Algorithm | Pairwise CLIP Score ⬆️ | CMMD ⬇️ | LPIPS ⬇️ | MSSSIM ⬆️ | PSNR ⬆️
-- | -- | -- | -- | -- | --
fp8_dynamic_compiled | 98.247383 | 0.007629 | 0.071664 | 0.948175 | 26.475830
static_fp8_diffusers_compiled | 98.211365 | 0.009179 | 0.076860 | 0.944918 | 26.176754
time_aware_fp8_diffusers_compiled | 98.270973 | 0.008225 | 0.079726 | 0.942991 | 25.820734

<p>LAION256, 46 prompts</p>

Quantization Algorithm | Pairwise CLIP Score ⬆️ | CMMD ⬇️ | LPIPS ⬇️ | MSSSIM ⬆️ | PSNR ⬆️
-- | -- | -- | -- | -- | --
fp8_dynamic_compiled | 96.642136 | 0.019193 | 0.075059 | 0.950166 | 26.583673
static_fp8_diffusers_compiled | 96.565979 | 0.021219 | 0.084726 | 0.939995 | 25.400696
time_aware_fp8_diffusers_compiled | 97.250900 | 0.019550 | 0.071551 | 0.953376 | 27.008932

<h1>TL;DR</h1>
<h3>Monitoring the Timestep</h3>
<p>The implementation expects the model denoiser used to have a <code>timestep</code> parameter in its signature. This is captured in the <code>model_check_fn</code>. However, the parameter existence doesn’t ensure the model call passes it down - thus we need/do a check in the pre-forward hook too.</p>
<h3>Buckets</h3>
<p>The original implementation was quite simple (also NOT <code>torch.compile</code> compatible), tracking the changes in the time (a value between 0 and 1) and incrementing a counter symbolizing the current step (integer value) - I refer to this as <em>tracking the monotonously decreasing time</em>. Despite intuitively simple, this coupled calibration and inference - they both need to run the exact same time schedule - steps, and time changes.</p>
<p>Even when the step count is the same, this would break when the schedule is not the same… For example, running inference with 10 steps exponentially spaced on a model calibrated on 10 steps uniformly spaced. Step 2 is calibrated on the image (noise) distribution at t=0.9 (previous timestep), while during inference - for e.g. at t=0.7.</p>
<p>In order to calibration and inference independent, we use bins that <em>group</em> nearby time ranges. In particular, we record the exact times and create bin-edges between every two neighboring values.</p>
<h3>Storing the scales in the TimestepTracker</h3>
<p>My first implementation required each layer to store its per-timestep scales. This required a <code>torch.bucketize</code> call on each forward pass, in order to find the appropriate scale for the current timestep. However, empirical evaluation showed me that this implementation is ~1 sec (w.r.t. 8.5 seconds execution time) slower compared to the static algorithm. Removing <code>torch.bucketize</code> and setting the scale to a dummy value removed this overhead. Despite being a small (and intuitively fast) method, it is invoked 708 times for a complete generation for Qwen-Image-2512. Considering it cannot be merged into the <code>_scaled_mm</code> kernel, it introduces not only overhead, but unnecessary control swap.</p>
<h3>Save/Load Artifacts</h3>
<p>Similarly to how we handle artifacts in the <code>static_fp8_diffusers</code> method, we store only the observed values and re-drive any scale-related computations on load. In this case, we store the timesteps and the running amax recorded for each. On load (via  <code>freeze_input_scales</code>), we compute the bin edges.</p>

<h3>Byte Alignment</h3>

Since cuda 12.9.0, the scales should be 16B aligned. The original algorithm implementation was developed on older version, which did not require such an alignment. This allowed storing the scales for all timesteps for a given layer in a single 1D tensor.

---

> **Note:** 


begumcig (approved):
No notes! Let's merge this! Perfect, thank you so much <3

